PR #371, opened 2020-11-05
Content: ItemList remove hash redirect, add todo

nibblebot:


shrunyan:
Solved in https://github.com/zesty-io/manager-ui/pull/386

PR #386, opened 2020-11-10
Content: refactor ItemList to functional

### LINKED ISSUE #362: Content: ItemList single item should redirect to ItemEdit
currently redirects via window.location.hash which doesn't work,
should be a history.push to the correct itemZUID

nibblebot:
fixes #362


src/apps/content-editor/src/app/views/ItemList/ItemList.js
```diff
@@ -56,708 +63,629 @@ export default connect((state, props) => {
...
-      height: 1000,
-      width: 1000,
-      sortedBy: null,
-      status:
-        (this.props.filters[this.props.modelZUID] &&
-          this.props.filters[this.props.modelZUID].status) ||
-        "all"
+})(function ItemList(props) {
+  const history = useHistory();
+  const _isMounted = useRef(false);
```
shrunyan, line 68: What is the purpose of using a ref when later it is just boolean evaluation?

src/apps/content-editor/src/app/views/ItemList/ItemList.js
```diff
@@ -56,708 +63,629 @@ export default connect((state, props) => {
...
+  // table, list dimensions
+  const table = useRef();
+  const list = useRef();
+  const [height, setHeight] = useState(1000);
+  const [width, setWidth] = useState(1000);
+
+  // saving, loading states
+  const [saving, setSaving] = useState(false);
+  const [loading, setLoading] = useState(false);
+  const [loadingSecondary, setLoadingSecondary] = useState(false);
```
shrunyan, line 79: Lets give line 78 & 79 more semantic names. Ideally they would communicate what they are affecting when in a loading state.

src/apps/content-editor/src/app/views/ItemList/ItemList.js
```diff
@@ -56,708 +63,629 @@ export default connect((state, props) => {
...
-        modelZUID: this.props.modelZUID,
-        filters: {
-          filterTerm: this.state.filterTerm,
-          sortedBy: this.state.sortedBy,
-          status: this.state.status,
-          colType: this.state.colType,
-          reverseSort: this.state.reverseSort
-        }
-      });
+  // HACK: runFilters after waiting for new props
```
shrunyan, line 136: The majority of places where you have included a `// HACK` comment need more information on why it is a hack. 
e.g. > runFilters after waiting for new props

Is this a hack vs running the filters synchronously? The "hack" on the `useEffect` is confusing because what is described is the default behavior of `useEffect`. i.e. waiting for new props.

src/apps/content-editor/src/app/views/ItemList/ItemList.js
```diff
@@ -56,708 +63,629 @@ export default connect((state, props) => {
...
+    setLoadingSecondary(true);
+
+    try {
+      const [, itemsRes] = await Promise.all([
+        props.dispatch(fetchFields(modelZUID)),
+        props.dispatch(fetchItems(modelZUID, { limit: PAGE_SIZE, page: 1 }))
+      ]);
+      if (_isMounted.current) {
+        // render 1st page of results
+        // HACK: runFilters after waiting for new props
```
shrunyan, line 193: Is this trying to communicate that filters will be run at the next render cycle vs that code executing immediately?

shrunyan (approved):
Overall a welcomed improvement to the code structure and readability of this component. I've left a few comments. Also would you give this another once over now that you have had to look at it for awhile. See if anything pops out. Otherwise this looks ready to go.

src/apps/content-editor/src/app/views/ItemList/ItemList.js
nibblebot: ref will retain it's value across renders unlike a regular variable

shrunyan: Wouldn't `useState` have the same behavior? 

My concern with a ref is weight and possibility of a memory leak. Meaning under the hood I assume react is allocating more resources when generating a ref vs state. Could there be a potential for memory leaks if `ref` references are being maintained. i.e. Could this create detached DOM nodes?  Maybe a non issue as the ref is not being attached to a DOM node. p.s. I don't have any authoritative answers to these theoretical questions, just posing thoughts for consideration. I'm not viewing this as a blocker

src/apps/content-editor/src/app/views/ItemList/ItemList.js
```diff
@@ -56,708 +63,630 @@ export default connect((state, props) => {
...
+  // table, list dimensions
+  const table = useRef();
+  const list = useRef();
+  const [height, setHeight] = useState(1000);
+  const [width, setWidth] = useState(1000);
+
+  // saving, loading states
+  const [saving, setSaving] = useState(false);
+  const [loadingFirstPage, setLoadingFirstPage] = useState(false);
+  const [loadingSecondaryPages, setLoadingSecondaryPages] = useState(false);
```
shrunyan, line 79: I am still not clear on these. What is a page? An app screen or an API page? Would something like `initialLoad` and `backgroundLoad` explain what these deal with?

src/apps/content-editor/src/app/views/ItemList/ItemList.js
```diff
@@ -190,24 +188,26 @@ export default connect((state, props) => {
       ]);
       if (_isMounted.current) {
         // render 1st page of results
-        // HACK: runFilters after waiting for new props
+        // HACK: runFilters next render cycle with fresh props/state
```
shrunyan, line 191: Still not sure I'm following why this is a "hack" as it is the default behavior on setting state. **Unless I'm missing something**.

If the intent is to document the flow of this code as it can be complex for those learning React I would suggest these all get moved up to where the state is initialized and describe the flow without noting it is a hack. Also include why we are describing the flow that is standard behavior. This should also help with the repetition of these comments.

src/apps/content-editor/src/app/views/ItemList/ItemList.js
nibblebot: changes to `useState` triggers re-render whereas changes to a `ref` does not. `useRef` is the standard way to track isMounted in functional components (vs. class components which can use instance variables)
prior art: https://dev.to/tusharkashyap63/use-refs-to-check-if-a-component-is-still-mounted-2gk7 http://naomiajacobs.com/useRef-vs-useState/
since we are tracking a boolean here there is no worry about detached dom nodes.

src/apps/content-editor/src/app/views/ItemList/ItemList.js
nibblebot: initial/background Load works for me

PR #420, opened 2020-12-02
usePermissions custom hook

shrunyan:
Many times within a product we will need to determine if a user should be able to see specific content or perform specific action. This hooks presents an API for determining an answer to this question; Can a user do "use" a specific permission?

```
import { usePermissions } from 'shell/hook/use-permissions'
....
const canPublish = usePermissions("PUBLISH") 
```

- Does not account for granular roles. Marked as TODO
- Has product access bolted on. This is non elegant and in the future could use a refactor. For now it gets around the primary question of can this user access the code editor
- In order to use this new hook a few components had to be refactored to functional components. This caused significant line number changes. Unfortunately making the diff not very easy to read.

### DIFF
--- a/src/shell/hooks/use-permissions.js
+++ b/src/shell/hooks/use-permissions.js
@@ -11,15 +11,15 @@ export function usePermission(action, zuid) {
   const state = store.getState();
   const role = state.userRole;
 
-  // "With great power comes great responsibility" - Uncle Ben
+  // "With great power comes great responsibility" - Benjamin Franklin Parker
   if (state.user.staff) {
-    return state.user.staff;
+    return true;
   }
 
   // Check super
   // super users are allowed all actions across all resources
   if (role.systemRole.super) {
-    return role.systemRole.super;
+    return true;
   }
 
   // Check granular
--- a/src/shell/hooks/use-permissions.js
+++ b/src/shell/hooks/use-permissions.js
@@ -1,0 +1,58 @@
+"use strict";
+
+import { useStore } from "react-redux";
+
+/**
+ * TODO given a specific ZUID, determine whether that user is allowed to do a certain action
+ * e.g. Can user publish content item
+ */
+export function usePermission(action, zuid) {
+  const store = useStore();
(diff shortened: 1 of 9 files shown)


src/shell/hooks/use-permissions.js
```diff
@@ -0,0 +1,58 @@
...
+
+    // check for specific product access
+    // NOTE this has been bolted on to the action check
+    // this should get refactored at some point to separate these concepts
+    case "CODE":
+      // List of role names with access to code editor
+      return ["Owner", "Admin", "Developer"].includes(role.systemRole.name);
+
+    default:
+      false;
```
nibblebot, line 56: return false

src/shell/hooks/use-permissions.js
```diff
@@ -0,0 +1,58 @@
...
+ * e.g. Can user publish content item
+ */
+export function usePermission(action, zuid) {
+  const store = useStore();
+  const state = store.getState();
+  const role = state.userRole;
+
+  // "With great power comes great responsibility" - Uncle Ben
+  if (state.user.staff) {
+    return state.user.staff;
```
nibblebot, line 16: return true

src/shell/hooks/use-permissions.js
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  // Check super
+  // super users are allowed all actions across all resources
+  if (role.systemRole.super) {
+    return role.systemRole.super;
```
nibblebot, line 22: return true

nibblebot (commented):
We can go ahead with the hooks approach but computing derived state from the store can also be achieved with reselect and selectors
https://redux.js.org/recipes/computing-derived-data/

src/shell/hooks/use-permissions.js
shrunyan: :facepalm: ouch

shrunyan:
hook has been updated to memoize redux state selection with `useSelector` and permission calculation with react's `useMemo`

PR #421, opened 2020-12-02
Chore/build

shrunyan:
**The bundle size has doubled. We will need to understand why and whether we can reduce this before merging.**

Updates build process and dependencies to latest

- webpack@5
  - config updated to new syntax
  - fallbacks added for `stream`, `buffer` and `util`
    - webpack no longer auto polyfills these
    - use `DefinePlugin` to expose `Buffer` global in bundle
  - **can not** update `webpack-cli@4` because `webpack-dev-server` breaks with latest
- babel latest
- react@17

- added: `file-loader` to support latest `monaco` loading of `ttf` files

- removed: unused deps
- removed: invalid css modules


shrunyan:
Closed in favor of https://github.com/zesty-io/manager-ui/pull/757

Issue #426, opened 2020-12-09
Leads CSV export: does not escape commas in strings


kakoga:
**Describe the bug**
When exporting Leads data to CSV, data is converted from JSON to a CSV, however, it will parse commas in strings

**To Reproduce**
1. Navigate to `{manager_ui_url}/leads` on **zesty.pw prod testing instance**
2. Click on the `EXPORT CSV` button
3. Open downloaded CSV and notice that `formData` is not valid JSON and has been split

**Expected behavior**
When exported Leads data is parsed it will not create new cols when encountering a comma in a string.

**Screenshots**
<img width="1061" alt="Screen Shot 2020-12-08 at 4 50 30 PM" src="https://user-images.githubusercontent.com/16694502/101559800-ba552b00-3976-11eb-93f6-5f9603f9bd55.png">

**Desktop (please complete the following information):**
 - OS: Mac 
 - Browser Chrome
 - Version 87.0

Additional context
Notes from @MukeshGKastala
Exporting leads into a CSV is a purely frontend feature(no API call is made). The javascript library we use, [json-to-csv-export](https://www.npmjs.com/package/json-to-csv-export), seems to be having issues with nested JSON structures and since we no longer export just the **Form Body**(https://github.com/zesty-io/manager-ui/issues/425) but rather the whole lead we're running into issues.

kakoga:
related to issue #425

PR #427, opened 2020-12-09
fix(leads): export forms instead of leads

### LINKED ISSUE #425: Leads CSV Export different between Legacy & Manager UI
**Describe the bug**
Lead exports from Legacy and Manager UI result in 2 different sets of data. 

**To Reproduce**
1. Go to legacy and export a Leads table
2. Go to Manager UI and export the same Leads table.
3. Compare both exports and notice that the Manager UI exports more data than Legacy and should have only the Form Data col parsed as a CSV... not all the extra cols.

**Expected behavior**
When I export data I expect to see only the Form Data - not all the Zuid, etc cols too.

**Screenshots**
Screen Shot 2020-12-08 at 3 13 30 PM

 - OS: Mac Catalina
 - Browser Chrome
 - Version 87.0

**Additional context**
Notes from @MukeshGKastala 
Legacy seems to be exporting just the **Form Body**, while Manager UI is working on the whole lead.

MukeshGKastala:
Port the CSV export logic of Legacy to Manager UI

Closes #425

### DIFF
--- a/src/apps/leads/src/app/components/LeadExporter/DownloadCSVButton/DownloadCSVButton.js
+++ b/src/apps/leads/src/app/components/LeadExporter/DownloadCSVButton/DownloadCSVButton.js
@@ -33,13 +33,14 @@ export default connect(state => {
       );
       leads = FilterService.filterByDate(leads, this.props.filter);
       leads = FilterService.filterByFuzzyText(leads, this.props.filter);
+      const forms = leads.map(lead => lead.formData);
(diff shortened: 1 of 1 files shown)


MukeshGKastala:
@nibblebot @shrunyan 
Since forms are 1-1 key-value pairs(no nested JSON), can https://github.com/zesty-io/manager-ui/issues/426 also be closed?

shrunyan:
Yes. Closes https://github.com/zesty-io/manager-ui/issues/426

Issue #418, opened 2020-11-30
Content: Deleted items still displaying in UI


giseleblair:
**Describe the bug**
After deleting a content item, it still displays in the content UI

**To Reproduce**
1. Go to a content item and delete
2. Click on its content model in the UI to see that the deleted item is still showing.

**Expected behavior**
When an item is deleted, it shouldn't display in the UI. It's confusing that it still shows and I get an error message when I click into it.

**Loom**
https://www.loom.com/share/125d939a7feb4beaa03f95f61a09d5dc

d88naimi:
Branch:  feature/#418-content-deleted
@shrunyan @giseleblair I believe after creating new item is made and the `Meta` link is never selected the selected item is never save with `meta` tag hence crashing the app indefinitely.  I resolved the issue by adding optional chaining for the undefined meta tag https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining However im not sure if this is band aid and if something with the backend needs to be updated when passing the Meta. Stuart let me know how to proceed if you want me to push up my fix.

d88naimi:
Side note* the bug discussed here may be separate of the delete bug Gisele has mentioned but in the video her app was crashed and believe they may be related.

Couple ideas to be discussed is to automatically fill the meta  tag link text.
Possibly redirect user to Meta view.

d88naimi:
@giseleblair Update

shrunyan:
This is now deployed to production. https://github.com/zesty-io/manager-ui/commit/e8b22d97381daeba9bab3d42461a79cf8dcc67c3

PR #429, opened 2020-12-10
added optional chaining for undefined values

d88naimi:
https://github.com/zesty-io/manager-ui/issues/418

Hot Fix for app crash #418 

I left notes in the original ticket, lets sync up and let me know how to proceed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -83,6 +83,7 @@
     "@babel/core": "7.8.7",
     "@babel/plugin-proposal-class-properties": "7.8.3",
     "@babel/plugin-proposal-json-strings": "7.8.3",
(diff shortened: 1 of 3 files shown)


shrunyan:
Accepting the syntax addition of [optional chaining operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) as it is a [stage 4 proposal](https://github.com/tc39/proposal-optional-chaining#status). I have added the [babel optional chaining plugin](https://babeljs.io/docs/en/babel-plugin-proposal-optional-chaining) to handle this change.

PR #430, opened 2020-12-10
fixed height calc for schema nav

### LINKED ISSUE #422: Schema - unable to view all models
**Describe the bug**
In the Schema section, i'm unable to scroll & view all models in the left-hand sidebar. More accurately, I can only scroll so far... even if I zoom out to try to view more - just in case my screen size is an issue - I still can only scroll so far.

**To Reproduce**
1. Go to Schema - do this in an instance that has lots of models
2. Scroll down & see that you're not seeing all models.

**Expected behavior**
I expect to be able to see all models when I scroll through the list. 

**Screenshots**
Screen Shot 2020-12-04 at 1 25 38 PM

 - OS: Mac 13" screen
 - Browser Chrome
 - Version 87

d88naimi:
Adjusted the height calc to fit more list items in the container that were originally being cut off.
fixed #422 

Screen Shot 2020-12-10 at 10 35 15 AM
CC: @kakoga

### DIFF
--- a/src/apps/schema/src/app/components/Nav/SchemaNav.less
+++ b/src/apps/schema/src/app/components/Nav/SchemaNav.less
@@ -5,12 +5,11 @@
     background-color: #4b5468;
     display: flex;
     flex-direction: column;
-    padding: 4px 8px 4px 8px;
+    padding: 8px;
 
     @media (min-width: 2000px) {
       align-items: center;
       flex-direction: row;
-      padding: 8px;
     }
 
     .Search {
@@ -40,7 +39,7 @@
   }
 
   .ModelList {
-    height: calc(100vh - 54px - 86px);
+    height: calc(100vh - 54px - 102px);
 
     margin: 0;
     overflow-y: scroll;
--- a/src/apps/schema/src/app/components/Nav/SchemaNav.less
+++ b/src/apps/schema/src/app/components/Nav/SchemaNav.less
@@ -8,7 +8,6 @@
     padding: 8px;
 
     @media (min-width: 2000px) {
-      align-items: center;
       flex-direction: row;
     }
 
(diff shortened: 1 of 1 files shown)


shrunyan:
@d88naimi does this account for the differing navigation height when app is constrained and unconstrained? As the button would display in the same row of the input at that point.

d88naimi:
@shrunyan Yes this edit works on both constrained and unconstrained views. 

Screen Shot 2020-12-11 at 9 35 32 AM

shrunyan:
The height change is fine but the padding I'm not sure why that was added. It also pushes the scrollbar away from the edge of the app area, we want those to be flush.

d88naimi:
@shrunyan edit the padding on actions to prevent the content from going out of the screen and now the scroll bar is flush again.

Screen Shot 2020-12-15 at 1 25 43 PM

shrunyan:
@d88naimi what is the purpose of adding the padding? What is this fixing?

d88naimi:
@shrunyan  if no padding is added then then search and create button is flushed against the container. 

Image with no padding, this would not then be consistent with Content and Code view 
Screen Shot 2020-12-15 at 4 00 42 PM

shrunyan:
Something seems off, I am not seeing that behavior in production.

Linux (Ubuntu)
Chrome Browser
image

d88naimi:
@shrunyan under `Test Dataset 4` do you have some more models being cut off ? 

Screen Shot 2020-12-16 at 6 50 05 AM

shrunyan:
@d88naimi I am referring to the padding around the search and creation button, as noted in your previous screenshot. Compare that to the one I showed.

d88naimi:
@shrunyan finally calculated the correct ModelList height thank you.

PR #432, opened 2020-12-11
webpack: remove moment locales

nibblebot:
removing moment locales reduces vendor bundle size from 5.96MB min to 5.73MB min


shrunyan:
@nibblebot do we need these for the publish schedule feature, so you can select different timezones?

PR #434, opened 2020-12-11
Content: headless nav allow nested items

### LINKED ISSUE #416: Model parent not displaying as expected
**Describe the bug**
In Schema, when you select a model parent for a content model, it does not display as a nested item in the Content sidebar.

**To Reproduce**
1. Go to Schema and click on a content model
2. Click on Model Settings and select a Parent content model
3. Go to the Content section
4. See that the child/parent content models are not nested as expected

**Expected behavior**
I would expect that when a content model has a Parent model that it would appear nested under that parent in the Content sidebar

**Additional context**
The relationship is saved. When I go back to Schema, I can see that the selected parent model is still there for the child. It is just not reflected in the Content sidebar.

nibblebot:
fixes #416


shrunyan:
@nibblebot this is showing duplicates of models.

image

shrunyan:
@nibblebot I have added additional changes. Would appreciate you taking a peek. This one almost broke my brain.. recursion!
**TL;DR;**
We need to generate 3 navigation; primary, headless and hidden.
The API returns a linked list of nodes. We convert this into a tree structure. Once we have the tree we should be able to bisect that tree into the 3 distinct navs we need.
By walking the tree we can ask questions of each leaf to determine which nav it falls into. Building up the 3 navs.

PR #450, opened 2021-01-05
Revert "Revert: Legacy unpublish"

shrunyan:
Reverts zesty-io/manager-ui#449

Merge when WebEngine uses the new deleted publishings structure


shrunyan:
Closing as we now proxy to legacy through latest api

PR #483, opened 2021-02-03
Media: remove legacy media

### LINKED ISSUE #4: Refactor Media to React
The media app was built using the [riotjs](https://v3.riotjs.now.sh) JavaScript library. This is a legacy dependency and needs to be refactored out. There is also a `window.zesty` global which uses the `riot.observable` method to allow for listening to events within the media application. 

## Concepts

- **Bin**: Represents the cloud resource of a GCP bucket
- **Group**: A record which acts as a way to relate assets to be displayed together and thought of as a logical collection 
- **Asset**: An individual file uploaded to the media app. Can be of any file type. But will display and have differing experiences based upon the assets file type.

## Architecture & Performance

The original media app was not designed for large amounts of images and groups. More than 10,000 images and 2,000 groups. Performance issues arise in the current experience when approaching these numbers. This is because of 2 issues identified so far;

**Performance Issues**

1. All files are loaded from a bin. Doing so causes the response latency to increase as the amount of files returned does. 
2. Groups are render in the app sidebar menu. Once it reaches a certain amount, e.g. 2,000 or more, re-renders begin to become expensive and slow down app responsiveness.

**Performance Requirements**:

- Responsive (non-laggy) UI interactions with multiple bins. Group amounts of 50,000 and asset amounts of 500,000. _Theoretically this should scale to orders of magnitudes further with delayed loading._

### State Management

All media state should be added to the app shell global redux store. Asset state is consumed by other sub apps which reference media.

### Delayed Loading

Instead of loading all of the images from a bin initially we should only load the images relevant to the currently viewed group. Allowing state to build up over time.

Note: _Bins can function as groups as well as being a bin. The will need to follow this pattern as well when they are viewed. Only loading the assets specific to that bin-group._

#### Search

If we change to delayed loading we will need a new `media-manager-service` API. One which will take a string and search for a matching asset. The current experience simply filters on the in memory assets as they are all loaded up front.

### Deep Links
When originally built the app did not provide deep linking. This is something we want to added. The goal being to allow users to share links to exact assets they want to collaborate on. In addition the rest of the manager-ui can use these to allow cross sub-app functionality.

For example;

- bin link: `/media/1-000-00000`
- group link: `/media/1-000-00000/2-000-00000`
- asset link: `/media/1-000-00000/2-000-00000/3-000-00000`

## User Experience

Various experiences we want to add or improve with the media refactor

### Asset Settings
Currently when an asset settings are viewed this is displayed as a "shelf". I.E. it slides open inline to display a larger (original size device screen allowing) image. It also shows meta data associated to the asset. Potentially we should consider an alternate solution. Maybe a modal? This would allow for a full screen take over providing more room to display original sizing of larger images on smaller devices. Whatever solution is used it should feel seamless when deep linking directly to an asset.

### Asset Variations

With image assets we allow for dynamic resizing of them via the `media-modify-service`. The asset settings screen should display all available image variant URLs along with their parameters (dimension, crop style, etc). This will allow for access to these pre-created assets as well as providing a general sense of how the original image is being consumed.

### Asset Generation

We should provide the ability to request a specific asset variation from the asset settings. I am imagine a couple of inputs; dimension and crop/fit, which would then make a request to the `media-modify-service` to generate that desired variant. Which would then display and provide the URL for the variant result.

nibblebot:
closes #4 
- remove riot dependency
- remove DnD dependency
- remove Clipboard dependency
- remove tags.js
- remove extraneous fontawesome
- remove growl (replace growl with dispatch)
- remove extraneous npm tasks
- rename media-2/ to media/


shrunyan:
@nibblebot I'd like to break this PR apart more. Can you separate out the tasks of;

- update favicon -> Media to use new media app
- update Styles -> Media to use new media app

This way we can get the experiences corrected first and then evaluate the removal of the riotjs code.

shrunyan:
@nibblebot are we using DnD in schema field ordering? Will removal effect that?

PR #516, opened 2021-02-18
Bug/missing pubsub

nibblebot:


src/shell/components/global-actions/components/ActivePreview/ActivePreview.js
```diff
@@ -89,8 +90,7 @@ export default connect(state => {
       "height=850, width=1260, location=0, menubar=0, status=0, titlebar=0, toolbar=0"
     );
 
-    zesty.on("PREVIEW_ROUTE", route);
```
nibblebot, line 92: `PREVIEW_ROUTE` is never triggered so this is appears to be dead code

shrunyan:
@nibblebot what is the resulting bundle size difference between riot and pubsub-js?

shrunyan:
If we just used a shakeable import. e.g. {oservable} from riot would that be supported and how would the bundle size compare?

shrunyan:
Lets try using the `riotjs/observable@4.0.4`. Previously we avoid upgrades becazuse it would break the riot app. Now that we only have the single dependency from the library this should be a non-issue. As long as the API of the riot observable hasn't significantly changed

Issue #522, opened 2021-02-23
Web IDE - missing publish icon


kakoga:
**Describe the bug**
when an item is saved and not published in the IDE, there it an icon next to the item's name indicating this - this indicator no longer shows in the UI and I can't tell which items have changes that aren't published.

I didn't notice this last week - it seems like it's something new this week

**To Reproduce**
1. In the IDE, click on any file type, and save it. 
2. You will not see the arrow/cloud publish icon next to the file name

**Expected behavior**
I expect this to work the same as Legacy. When I save a file in legacy I can see the publish icon.

**Screenshots**
<img width="1398" alt="Screen Shot 2021-02-23 at 9 07 37 AM" src="https://user-images.githubusercontent.com/16694502/108880530-3bf04880-75b7-11eb-92fe-e5089b1a0973.png">

**Desktop (please complete the following information):**
 - OS: Mac
  - Browser Chrome
 - Version 88

kakoga:
A customer just reported this as well. it caused an issue in which they were not able to debug their own code - I'd like to bump the priority on this one if possible, @nibblebot

PR #534, opened 2021-03-02
refactor: provide react component to nav node actions

shrunyan:
- [ ] depends on https://github.com/zesty-io/design-system/pull/98


shrunyan:
Fixes https://github.com/zesty-io/manager-ui/issues/522

nibblebot (approved):

PR #547, opened 2021-03-09
fixed: undefined reference

shrunyan:


src/shell/store/contentVersions.js
```diff
@@ -11,7 +11,7 @@ export function contentVersions(state = {}, action) {
         // DeDupe versions
         [action.itemZUID]: [action.data, ...existingVersions].reduce(
           (acc, item) => {
-            if (!acc.find(el => el.meta?.version === item.meta.version)) {
+            if (!acc.find(el => el?.meta?.version === item?.meta?.version)) {
```
nibblebot, line 14: if `item` is `null`, I think we should skip item, not `acc.push(item)`

shrunyan: changed to check for presence of action data before inserting into state

PR #564, opened 2021-03-10
fixed: handle 401 and provide a re-login experience

shrunyan:
instead of throwing errors

### DIFF
--- a/src/utility/request.js
+++ b/src/utility/request.js
@@ -1,5 +1,7 @@
 import Cookies from "js-cookie";
+import { store } from "shell/store";
 import { notify } from "shell/store/notifications";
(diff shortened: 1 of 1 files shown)


src/utility/request.js
```diff
@@ -51,12 +53,28 @@ export function request(url, opts = {}) {
...
+          store.dispatch(
+            notify({
+              kind: "warn",
+              message: "Your session has ended. Opening the login modal."
+            })
+          );
+
+          setTimeout(() => {
+            store.dispatch(verify());
+          }, 5000);
```
nibblebot, line 72: why use a notification at all and why wait 5 seconds? why not just immediately hit the user with a login modal?

PR #567, opened 2021-03-10
fixed: ensure versions data structure before mapping

shrunyan:


src/apps/content-editor/src/app/views/ItemEdit/components/Header/ItemVersioning/VersionSelector/VersionSelector.js
```diff
@@ -71,16 +71,17 @@ export default connect((state, props) => {
...
-          <Option
-            key={`${item.meta?.ZUID}-${item.meta?.version}`}
-            className={styles.VersionOption}
-            value={item.meta?.version}
-            html={`Version ${item.meta?.version} <small>${moment(
-              item.web?.createdAt
-            ).format("MMM Do YYYY, [at] h:mm a")}</small>`}
-          />
-        ))}
+        {Array.isArray(props.versions) &&
```
nibblebot, line 74: not sure why we need this check and we are doing a `const versions = state.contentVersions[props.itemZUID] || [];` on line 12

PR #568, opened 2021-03-10
removing non used state

d88naimi:
Removed non used actions

### DIFF
--- a/src/shell/store/user.js
+++ b/src/shell/store/user.js
@@ -8,11 +8,7 @@ export function user(
     firstName: "",
     email: "",
     permissions: [],
-    selected_lang: "",
-    latest_edits: [],
-    latest_publishes: [],
-    total_user_actions: [],
-    total_everyone_actions: []
+    selected_lang: ""
   },
   action
 ) {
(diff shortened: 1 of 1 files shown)


src/shell/store/user.js
```diff
@@ -44,17 +40,7 @@ export function user(
     //DRY Creating latest_edits reducer for dashboard
     case "FETCH_USER_LOGS_SUCCESS":
```
shrunyan, line 41: I think this whole case can be removed as it was added with the dashboard work. https://github.com/zesty-io/manager-ui/commit/14b29a0754cf2b1575f58f0dbd957bbe544ec169#diff-da848ab72737014691325cb22eb36812b100596c01de290443fa2b66f52177d4R45

shrunyan (changes requested):

src/shell/store/user.js
d88naimi: Okay removed, I though it was necessary for logs store but I now see log has it's own action call

shrunyan (approved):

PR #572, opened 2021-03-12
Feature/adding content titles

d88naimi:
showing 5 latest unique edited items


shrunyan (changes requested):
Lets use the lodash uniqBy function with the content item ZUID being the unique value. https://lodash.com/docs/4.17.15#uniqBy

PR #588, opened 2021-03-17
fix: ItemList fetch items by user lang

### LINKED ISSUE #471: Toggling multi-lang in list view
**Describe the bug**
When selecting another language in a content model's list view, the items in that list will not display.
**To Reproduce**
1. Go to the list view of a content model in Content.
2. Click on dropdown to select another language.
3. See that list items do not load.

**Expected behavior**
I would expect for all cloned items to load when toggling through languages. 

**Screenshots**
https://www.loom.com/share/66c504bcd0cf41bfabfe1fe69f9e5c97

 - OS: MacOS Big Sur
 - Browser: Chrome and Firefox. NOTE: this behavior does not happen on Safari

nibblebot:
fixes #471

### DIFF
--- a/src/apps/content-editor/src/app/views/ItemList/ItemList.js
+++ b/src/apps/content-editor/src/app/views/ItemList/ItemList.js
@@ -46,8 +46,11 @@ export default connect((state, props) => {
     ? state.languages.find(lang => lang.code === state.user.selected_lang) || {}
     : {};
 
+  const userSelectedLang = state.user.selected_lang || "en-US";
+
   return {
     selectedLang,
+    userSelectedLang,
     model,
     modelZUID,
     user: state.user,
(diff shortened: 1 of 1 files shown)


src/apps/content-editor/src/app/views/ItemList/ItemList.js
```diff
@@ -665,7 +676,7 @@ export default connect((state, props) => {
                   type: "UNMARK_ITEMS_DIRTY",
                   items: props.dirtyItems
                 });
-                return load(props.modelZUID);
+                return Promise.resolve();
```
nibblebot, line 679: if we are navigating away from ItemList, and Discard the dirty data, there is no reason to call `load()` so I am just returning a resolved promise to satisfy the onDiscard API

src/apps/content-editor/src/app/views/ItemList/ItemList.js
```diff
@@ -46,8 +46,11 @@ export default connect((state, props) => {
     ? state.languages.find(lang => lang.code === state.user.selected_lang) || {}
     : {};
 
+  const userSelectedLang = state.user.selected_lang || "en-US";
```
nibblebot, line 49: on instances without multilang, `selected_lang` is undefined so we'll default to `en-US`. Do you see any problems with this approach? @shrunyan

shrunyan: That looks correct to me

shrunyan (approved):

PR #589, opened 2021-03-18
Feature/#169 purge btn gi

d88naimi:
Purge All button added in global instance

Screen Shot 2021-03-18 at 8 03 01 AM

### DIFF
--- a/src/shell/components/global-instance/GlobalInstance.js
+++ b/src/shell/components/global-instance/GlobalInstance.js
@@ -18,6 +18,7 @@ import { Select, Option } from "@zesty-io/core/Select";
 import { Url } from "@zesty-io/core/Url";
 
 import { useDomain } from "shell/hooks/use-domain";
+import { notify } from "shell/store/notifications";
 
 import styles from "./GlobalInstance.less";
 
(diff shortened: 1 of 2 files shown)


d88naimi:
@shrunyan i'm blocked on testing this locally.  Accounts.dev invites is in a broken state, there is no way of testing user role or signing up a new user.

shrunyan:
Have it say "Purge Instance" and add some margin so it feels like a discreet action from the instance links.

src/shell/components/global-instance/GlobalInstance.js
```diff
@@ -100,6 +109,30 @@ export default connect(state => {
...
+                setPurge(true);
+                return request(
+                  `${CONFIG.CLOUD_FUNCTIONS_DOMAIN}/fastlyPurge?zuid=${props.instance.ZUID}`
+                ).then(res => {
+                  setPurge(false);
+                });
+              }}
+            >
+              {purge ? (
+                <FontAwesomeIcon icon={faSpinner} />
```
shrunyan, line 128: this needs the `spin` property to animate

src/shell/components/global-instance/GlobalInstance.js
```diff
@@ -100,6 +109,30 @@ export default connect(state => {
...
+        {canPurge && (
+          <div>
+            <Button
+              kind="warn"
+              disabled={purge}
+              onClick={() => {
+                setPurge(true);
+                return request(
+                  `${CONFIG.CLOUD_FUNCTIONS_DOMAIN}/fastlyPurge?zuid=${props.instance.ZUID}`
+                ).then(res => {
```
shrunyan, line 122: use the `finally` function, so it stops on failure. add a `catch` and dispatch a notice if it failed.

shrunyan (changes requested):

src/shell/components/global-instance/GlobalInstance.js
```diff
@@ -119,17 +120,30 @@ export default connect(state => {
...
+                    });
+                  })
+                  .finally(() => {
+                    setPurge(false);
+                  });
               }}
             >
               {purge ? (
-                <FontAwesomeIcon icon={faSpinner} />
+                <FontAwesomeIcon className={styles.Spinner} icon={faSpinner} />
```
shrunyan, line 142: We do not need to style this ourselves fontawesome provides a `spin` property which handles this.

src/shell/components/global-instance/GlobalInstance.js
```diff
@@ -119,17 +120,30 @@ export default connect(state => {
                 setPurge(true);
                 return request(
                   `${CONFIG.CLOUD_FUNCTIONS_DOMAIN}/fastlyPurge?zuid=${props.instance.ZUID}`
-                ).then(res => {
-                  setPurge(false);
-                });
+                )
+                  .then(res => {
```
shrunyan, line 124: This `then` condition is superflous as it is covered in the `finally` case

shrunyan (changes requested):

Issue #544, opened 2021-03-04
Web IDE notification placement hinder workflow


kakoga:
This one falls between bug & enhancement - I'm filing it here so it doesn't get lost.

**Describe the bug**
When working in the Web IDE, the notifications cover up the files behind them and hinder the users workflow. There is not an obvious way to dismiss it therefore it impedes workflow.

**To Reproduce**
1. Go to the Web IDE
2. Click on a file & save it.
3. See the notification towards the lower left-hand corner with the bell icon
4. See that the notification is blocking access to the files behind the notification. This is especially problematic when the user is scrolled to the bottom of the sidebar.

**Expected behavior**
I would expect that there is an x in the corner of the notification that could allow me to quickly dismiss it. Or another clear way to dismiss the notification so it doesn't block my workflow.
OR
I would expect the notification to show up in the different place - like it does in the Legacy in a place that doesn't block a users workflow. 

**Screenshots**

**Manager UI**
<img width="1361" alt="Manager-UI-notifications" src="https://user-images.githubusercontent.com/16694502/110043633-99de0800-7cfc-11eb-832c-5b330040eae1.png">

**Legacy**
<img width="1347" alt="Legacy-notifications" src="https://user-images.githubusercontent.com/16694502/110043650-a4989d00-7cfc-11eb-8e27-71b56456285a.png">

**Desktop (please complete the following information):**
 - OS: Mac
 - Browser Chrome
 - Version 88

**Additional context**
<br><br><i>(related to <a href='https://zestysupport.zendesk.com/agent/tickets/1257'>Zendesk ticket #1257</a>)<br>gz#1257</i>

Issue #349, opened 2020-10-30
Notification Styles Feature


d88naimi:
**Is your feature request related to a problem? Please describe.**
add icons that will correspond to the notification message

**Describe the solution you'd like**
Design considering the importance of your message

**Describe alternatives you've considered**
Leave as is

**Additional context**
https://uxdesign.cc/a-comprehensive-guide-to-notification-design-2fff67f08b7a#:~:text=The%20placement%20of%20notifications%20is,of%20notifications%20with%20various%20viewports.

https://uxplanet.org/how-to-design-notifications-for-better-ux-6fb0711be54d

https://ux.stackexchange.com/questions/39265/placement-of-flash-growl-notifications-or-messages-in-web-application

Current
![Screen Shot 2020-10-30 at 11 33 36 AM](https://user-images.githubusercontent.com/22800749/97744590-be9f4600-1aa4-11eb-9bb1-6d89eee47327.png)

Possible styles: 
![Screen Shot 2020-10-30 at 11 35 42 AM](https://user-images.githubusercontent.com/22800749/97744191-0d001500-1aa4-11eb-92a3-eb9418a83212.png)

shrunyan:
related https://github.com/zesty-io/manager-ui/issues/544

PR #591, opened 2021-03-18
refactor: notifications into a drawer of actions

shrunyan:
image
image

Resolves https://github.com/zesty-io/manager-ui/issues/349
Resolves https://github.com/zesty-io/manager-ui/issues/544

# Notifications

These refactor changes up the location, styling and experience of action notifications in manager-ui.

We have chosen the bottom right of the view port display the current notice toast. This location is the least invasive to other actions which can be taken across the application. As the vast majority of interactions are stacked towards the top left and right of the view port.

The notification bell is now shown in the top right of the view port, next to the users avatar. This location was chosen as it is a common location within in other web based applications, i.e. learned behaviors. The drawer which slides out to reveal all notifications when clicked is kept near the action which triggers it. Which also feels well placed, the drawer, as it is on the right side of the screen which is the same location of the current action toast.


shrunyan:
Particularly looking for experience and design feedback/ideas.

ardeay:
I like the flyout. 

I'd like to see the single noti match the style, do this by

- Make it a square rectangle, edged up to bottom right and left
- have single flyout rectangle styling match the flyout draw

ardeay:
Note on behavior updates: https://www.loom.com/share/87487f060c0a4646ba4c3b4e7d33cf13

Issue #533, opened 2021-03-02
Content items in sidebar not reordering


giseleblair:
**Describe the bug**
Items in Content app sidebar not reordering after saving in the Reorder Modal.

**To Reproduce**
1. Go to an instance's Content app.
2. Click on the Reorder button
3. Drag and drop to reorder items and hit Save.
4. See that saved changes are not reflected in the UI

**Expected behavior**
Expect the UI to display items in the order the user has saved using the reorder modal.

**Screenshots**
https://www.loom.com/share/ff5fde2019a14ec7acf02078087dff87

**Desktop (please complete the following information):**
 - OS: macOS Big Sure
 - Browser: safari + chrome

**Additional context**
This also has similar behavior to #518 where dragging/dropping does not immediately maintain that state. But additionally, the re-ordered items on "Save" are not reflected in the UI. 
<br><br><i>(related to <a href='https://zestysupport.zendesk.com/agent/tickets/1152'>Zendesk ticket #1152</a>)<br>gz#1152</i>

kakoga:
A customer has reported this issue as they were reordering items the sidebar did not reflect their actions.

PR #600, opened 2021-03-23
fix: content sort root nav items

nibblebot:
fixes https://github.com/zesty-io/manager-ui/issues/533

### DIFF
--- a/src/apps/content-editor/src/store/navContent.js
+++ b/src/apps/content-editor/src/store/navContent.js
@@ -276,6 +276,8 @@ function split(tree) {
 
   walk(tree, (tree, leaf) => {
     if (leaf.hidden) {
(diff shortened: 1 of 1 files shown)


src/apps/content-editor/src/store/navContent.js
```diff
@@ -274,8 +276,6 @@ function split(tree) {
 
   walk(tree, (tree, leaf) => {
     if (leaf.hidden) {
-      // remove leaf from tree
-      tree.splice(tree.indexOf(leaf), 1);
```
shrunyan, line 278: These seems to fundamentally change how walking the tree works. I would be concerned about this introducing issues for other areas this data is used.

shrunyan (approved):

PR #603, opened 2021-03-23
Feature/#586 schema field dd icons

d88naimi:
Fixed text icon alignment, removed unused styles. 

Screen Shot 2021-03-23 at 3 44 39 PM


src/apps/schema/src/app/views/SchemaEdit/FieldAdd/FieldAdd.less
```diff
@@ -15,25 +15,4 @@
       font-size: 32px;
     }
   }
-  main {
```
shrunyan, line 18: This deletion seems like it have a large effect on styling with nothing to replace it.

shrunyan (approved):

Issue #609, opened 2021-03-24
Hiding items in content nav causes 2 items to be hidden


nibblebot:
**Describe the bug**
Hiding an item in the content nav will cause hidden item and the adjacent one to be hidden.

**To Reproduce**
1. Go to Content Nav
2. Click on Hide (Eye icon)
3. See missing adjacent item

**Expected behavior**
Only the hidden item should be removed from nav

PR #616, opened 2021-03-26
fix: splice hidden items from nav tree

nibblebot:
fixes https://github.com/zesty-io/manager-ui/issues/609

### DIFF
--- a/src/apps/content-editor/src/store/navContent.js
+++ b/src/apps/content-editor/src/store/navContent.js
@@ -276,8 +276,6 @@ function split(tree) {
 
   walk(tree, (tree, leaf) => {
     if (leaf.hidden) {
-      // remove leaf from tree
-      tree.splice(tree.indexOf(leaf), 1);
       hidden.push(leaf);
     } else if (!leaf.parentZUID) {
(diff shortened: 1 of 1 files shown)


src/apps/content-editor/src/store/navContent.js
```diff
@@ -309,5 +308,14 @@ function walk(tree, operate) {
...
     operate(tree, leaf);
+    if (leaf.hidden) {
+      toSplice.push(index);
+    }
+  });
+  // Splice hidden leaves off the tree in reverse
+  // so that indexes are are not shifted during splicing
+  // e.g. toSplice = [0, 2] removes 3rd item, then 1st item
+  toSplice.reverse().forEach(index => {
+    tree.splice(index, 1);
```
nibblebot, line 319: notice, I am moving the splice operation _outside_ of the forEach loop since splicing the array while looping it will result in skipped items

shrunyan: *mindblown*

shrunyan (approved):
Nice work on this solution

PR #647, opened 2021-04-09
fix: Multi-Lang related Fields filter by content lang

### LINKED ISSUE #610: Multi-lang: related fields show all lang opts
**Describe the bug**
when selecting items from a one-to-one and one-to-many field in an instance that multi-langs, the item should only show the related items in 1 language. however it shows items in _all_ languages. 

**To Reproduce**
1. Go to an instance with multi-lang and a relational fields. (eg blog w/ tags - I'll use this example throughout)
2. Open an item in the blog set. Select a related tag.
3. Go to the tags set. and in the table view select a language other than EN. 
4. Go back to blog set and select an article and click the relational dropdown to see all the tags - see that both EN and tags in other languages are available.

**Expected behavior**
related items will be displayed only in same language as the 'parent' item. Eg: a blog article with EN selected, will only show related tags in EN. 

**Screenshots**

Screen Shot 2021-03-24 at 2 19 16 PM

- OS: Mac
- Browser Chrome

**Additional context**
this is resolved by clearing indexeddb. this could possibly be related to other idexeddb issues that were reported.

<br><br><i>(created from <a href='https://zestysupport.zendesk.com/agent/tickets/1767'>Zendesk ticket #1767</a>)<br>gz#1767</i><br><br><i>(related to <a href='https://zestysupport.zendesk.com/agent/tickets/1315'>Zendesk ticket #1315</a>)<br>gz#1315</i>

### LINKED ISSUE #637: Multi Lang: Content loads related fields in 'en-US' instead of Content selected language
**Describe the bug**
starting with a fresh indexeddb, I load 'de' version of content and only see 'en' for one-to-one and one-to-many fields

**Expected behavior**
loading 'de' version of content should load 'de' version of related fields

nibblebot:
Multi-Lang content with related fields will now use the Content Lang to filter the related fields. 
If non-English content has English related fields, we will continue to show the selected English fields and user can remove them. Redux Middleware for related fields continues to loads items in English.

Hoist fetch languages to LoadInstance so that languages are guaranteed available for us to filter fields by language.

fixes #637 
fixes #610

### DIFF
--- a/src/apps/content-editor/src/app/components/Editor/Field/Field.js
+++ b/src/apps/content-editor/src/app/components/Editor/Field/Field.js
@@ -468,15 +468,12 @@ export default connect(state => {
       }
 
       const onOneToOneOpen = useCallback(() => {
-        if (!props.allLanguages.length) {
-          return Promise.resolve();
-        }
         return dispatch(
           fetchItems(relatedModelZUID, {
             lang: props.allLanguages.find(lang => lang.ID === langID).code
           })
         );
-      }, [props.allLanguages.length]);
+      }, [props.allLanguages.length, relatedModelZUID, langID]);
 
       let oneToOneOptions = useMemo(() => {
         return resolveRelatedOptions(
@@ -568,9 +565,6 @@ export default connect(state => {
 
       // Delay loading options until user opens dropdown
       const onOneToManyOpen = useCallback(() => {
-        if (!props.allLanguages.length) {
-          return Promise.resolve();
-        }
         return Promise.all([
           dispatch(fetchFields(relatedModelZUID)),
           dispatch(
--- a/src/apps/content-editor/src/app/components/Editor/Field/Field.js
+++ b/src/apps/content-editor/src/app/components/Editor/Field/Field.js
@@ -62,7 +62,14 @@ function sortHTML(a, b) {
   return 0;
 }
 
-function resolveRelatedOptions(fields, items, fieldZUID, modelZUID) {
+function resolveRelatedOptions(
+  fields,
+  items,
+  fieldZUID,
+  modelZUID,
+  langID,
+  value
+) {
   // guard against absent data in state
   const field = fields && fields[fieldZUID];
   if (!field || !items) {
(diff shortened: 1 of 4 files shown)


src/shell/components/load-instance/index.js
```diff
@@ -60,7 +63,8 @@ export default connect(state => {
               props.products &&
               props.instance.ID &&
               props.instance.domains &&
-              props.user.ID
+              props.user.ID &&
+              props.languages.length
```
shrunyan, line 67: With non multi-lang instances do we get a single default entry of "en-US"? Making this case work for both types of instances?

src/apps/content-editor/src/app/views/ItemEdit/components/Header/LanguageSelector/LanguageSelector.js
```diff
@@ -59,7 +50,7 @@ export const LanguageSelector = connect((state, props) => {
           className={cx(styles.LanguageSelector, props.className)}
           value={props.selectedLang}
           onSelect={handleSelect}
-          loading={loading}
+          loading={false}
```
shrunyan, line 53: Should this prop just be removed? I assume we no longer need it since language loading has been hoisted to the app boot?

src/apps/content-editor/src/app/components/Editor/Field/Field.js
```diff
@@ -529,22 +550,30 @@ export default connect(state => {
           props.allFields,
           props.allItems,
           relatedFieldZUID,
-          relatedModelZUID
+          relatedModelZUID,
+          langID,
+          value
         );
       }, [
-        Object.keys(props.allFields).length,
```
shrunyan, line 535: I am concerned this change is a performance regression. The reason we only ran the memo on length change of `allFields` and `allItems` is because a change to any property on those objects will trigger this. Which in turn can trigger renders across other components in the view. With highly related content items this becomes a significant rendering performance problem.

src/apps/content-editor/src/app/components/Editor/Field/Field.js
```diff
@@ -453,23 +467,30 @@ export default connect(state => {
...
       }
 
-      const onOneToOneOpen = useCallback(
-        () => Promise.resolve(dispatch(fetchItems(relatedModelZUID))),
-        []
-      );
+      const onOneToOneOpen = useCallback(() => {
+        return dispatch(
+          fetchItems(relatedModelZUID, {
+            lang: props.allLanguages.find(lang => lang.ID === langID).code
```
shrunyan, line 473: This function happens twice. Maybe worth hoisting out of the switch statement to a `selectedLang` variable?

shrunyan (commented):
Couple of questions/thoughts.

src/shell/components/load-instance/index.js
nibblebot: yes

src/apps/content-editor/src/app/components/Editor/Field/Field.js
nibblebot: restored

src/apps/content-editor/src/app/components/Editor/Field/Field.js
nibblebot: factored out the function, not the value, since we don't want the value calculated for all Fields

shrunyan (approved):

Issue #640, opened 2021-04-07
SEO app: no matches message displays after selecting path


giseleblair:
**Describe the bug**
When using the search field to search for an existing path, the results will display but when the user clicks on the item an error message displays. 

**To Reproduce**
https://www.loom.com/share/df97f5813f0549749164aa79277ea8c6

**Expected behavior**
There should not be an error message when the user clicks on the listed item.
<br><br><i>(related to <a href='https://zestysupport.zendesk.com/agent/tickets/1831'>Zendesk ticket #1831</a>)<br>gz#1831</i>

shrunyan:
After selection from the options list the list should close.

PR #648, opened 2021-04-10
Refactor ContentSearch

shrunyan:
Fixes https://github.com/zesty-io/manager-ui/issues/640

Large refactor to `ContentSearch` component. Smooths out existing functionality of;
- loading display
- no results message
- Arrow up/down through options
- Code structure
- [x] show item language
- [x] show item content model icon
- [x] show item last edited date
- [x] show last edited by user name
  - [ ] Can the app load user lookup include previous instance members? i.e. We want to be able to show historical information even after a user has left an instance.
- [ ] when redirecting internal to instance show instance live domain, if present
- [ ] Search for a path that does not include the initial preceding forward slash will not return results. Even if that path exists.

### DIFF
--- a/src/apps/seo/src/app/app.less
+++ b/src/apps/seo/src/app/app.less
@@ -35,6 +35,6 @@
     height: calc(100vh - 54px);
     overflow: scroll;
     padding: 0;
-    background: @zesty-field-blue;
+    background: @appBkgColor;
   }
 }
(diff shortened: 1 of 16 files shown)


shrunyan:
@nibblebot while reviewing this change would you think through the edge case / oddities you dealt with? Want to make sure I haven't lost any of those.

shrunyan:
@d88naimi would you please add styling to this layout?

PR #652, opened 2021-04-14
Fix/640 redirect no results

d88naimi:
fix filter your redirects by url in SEO

### DIFF
--- a/src/apps/seo/src/views/RedirectsManager/RedirectActions/RedirectFilter/RedirectFilter.js
+++ b/src/apps/seo/src/views/RedirectsManager/RedirectActions/RedirectFilter/RedirectFilter.js
@@ -1,41 +1,65 @@
-import React, { Component } from "react";
+import React, { useState, useRef } from "react";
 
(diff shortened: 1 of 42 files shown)


src/apps/seo/src/views/RedirectsManager/RedirectActions/RedirectFilter/RedirectFilter.js
```diff
@@ -1,41 +1,26 @@
-import React, { Component } from "react";
+import React, { useState, useRef } from "react";
```
shrunyan, line 1: useRef looks like it's not being used. should this be removed?

shrunyan (approved):

src/apps/seo/src/views/RedirectsManager/RedirectActions/RedirectFilter/RedirectFilter.js
d88naimi: removed

PR #662, opened 2021-04-20
swapped icon for loader to lock and hide delete btn for loader

### LINKED ISSUE #62: Web IDE: allows user to delete loader without warning
The editor in the manager allows the user to delete `loader` without warning.  This then breaks every HTML page on the site and it is not obvious what has happened or that the user needs to add a `loader` back with a call to `{{current_view}}` before anything will work again.  This is a terrible product experience.

Suggest some combination of the following to improve this:

* Don't let the user delete `loader` at all, and parse it for `{{current_view}}` to validate when saving
* Let the user delete `loader` and have the system fall back to one that contains just `{{current-view}}` if it doesn't find a file called `loader`
* Let the user delete `loader` but only after they have OK'ed an explicit message that tells them that this will break all HTML pages on their site (which may be OK for example if the instance they are working with has no HTML views because it is a pure backend)

d88naimi:
fixes #62 

Swapped loader icon to lock 
hiding delete button in loader view. 

code-loader-lock

### DIFF
--- a/src/apps/code-editor/src/app/components/FileActions/components/Delete/Delete.js
+++ b/src/apps/code-editor/src/app/components/FileActions/components/Delete/Delete.js
@@ -22,14 +22,18 @@ export const Delete = React.memo(function Delete(props) {
 
   return (
     <div className={styles.DeleteBtn}>
-      <Button
-        kind="warn"
-        onClick={() => setOpen(true)}
-        className={styles.Button}
-      >
-        <FontAwesomeIcon icon={faExclamationCircle} />
-        Delete
-      </Button>
+      {props.fileName !== "loader" ? (
+        <Button
+          kind="warn"
+          onClick={() => setOpen(true)}
+          className={styles.Button}
+        >
+          <FontAwesomeIcon icon={faExclamationCircle} />
+          Delete
+        </Button>
+      ) : (
+        " "
+      )}
       <Modal
         className={styles.DeleteFileModal}
         type="local"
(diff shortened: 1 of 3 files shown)


src/apps/code-editor/src/app/components/FileActions/components/Delete/Delete.js
```diff
@@ -22,14 +22,18 @@ export const Delete = React.memo(function Delete(props) {
...
     <div className={styles.DeleteBtn}>
-      <Button
-        kind="warn"
-        onClick={() => setOpen(true)}
-        className={styles.Button}
-      >
-        <FontAwesomeIcon icon={faExclamationCircle} />
-        Delete
-      </Button>
+      {props.fileName !== "loader" ? (
```
shrunyan, line 25: `fileName` is a user controlled value and as such can be brittle when relying on it. Lets use the `type`, which should also be `loader` for this specific view, as that is a system controlled value.

shrunyan (changes requested):
Use the `type` across all appropriate locations

src/apps/code-editor/src/app/components/FileActions/components/Delete/Delete.js
d88naimi: @shrunyan props is showing  fileType: "views"  for loader view, thats why I went with fileName.

shrunyan (approved):

PR #664, opened 2021-04-20
increased z-index so that drop downs do not overlap content header

d88naimi:
Increased z-index for header content but then that was conflicting with modals so I then increase design system modal aligner z-index. I tested other pages and did not see any other conflicts. I did notice tooltips had an index of 10 however tooltips will render only on hover so I don't think there should be conflicts. @shrunyan let me know if there are any other use cases that I may have missed thanks. 

PR in design systems https://github.com/zesty-io/design-system/pull/109

https://www.loom.com/share/00514a1b49074cf09ba549fb816a88b3

### DIFF
--- a/src/apps/content-editor/src/app/views/ItemEdit/components/Header/Header.less
+++ b/src/apps/content-editor/src/app/views/ItemEdit/components/Header/Header.less
@@ -3,7 +3,7 @@
 
 .Header {
   padding: 16px 32px;
-  z-index: 10;
+  z-index: 120;
   position: sticky;
   top: 0;
   background-color: #f2f7fc;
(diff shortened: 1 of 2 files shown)


src/apps/content-editor/src/app/views/ItemEdit/components/Header/Header.less
```diff
@@ -3,7 +3,7 @@
 
 .Header {
   padding: 16px 32px;
-  z-index: 10;
+  z-index: 120;
```
shrunyan, line 6: What is the logic behind the choice of the specific value of 120?

d88naimi: @shrunyan dropdown is 110 so I went with 120 and made modalAligner 130 following zesty's concept of increments by 10 base.

shrunyan (approved):

crzestyio (approved):

PR #667, opened 2021-04-21
fixed: only update instances list on successful response

shrunyan:


src/shell/store/instances.js
```diff
@@ -16,15 +17,41 @@ export function fetchInstances() {
...
-        throw res;
+        dispatch(
+          notify({
+            kind: "warn",
+            messages: "You are forbidden from loading an instances list"
+          })
+        );
+      }
+
+      if (res.status === 500) {
```
nibblebot, line 47: 5xx responses don't make it to the .then() case, should go to .catch(), right?

crzestyio (approved):

PR #671, opened 2021-04-21
Fix/613-idb-exception

shrunyan:
Use a custom idb-keyval store so we can keep a reference and recreate the store if deleted during runtime


src/utility/idb.js
```diff
@@ -0,0 +1,23 @@
...
+
+const DB = "zesty";
+const STORE = "manager-ui";
+
+// use a custom store so we can recreate the reference
+// if it gets deleted during runtime
+let customStore = createStore(DB, STORE);
+
+// no need to wrap so just  re-export
+export { get, getMany } from "idb-keyval";
```
nibblebot, line 12: don't we need to specify `customStore` for these functions as well?

shrunyan: Yep. Good catch.

src/utility/idb.js
```diff
@@ -0,0 +1,56 @@
...
+} from "idb-keyval";
+
+const DB = "zesty";
+const STORE = "manager-ui";
+
+// use a custom store so we can recreate the reference
+// if it gets deleted during runtime
+let customStore = createStore(DB, STORE);
+
+export default {
```
shrunyan, line 21: I feel like there is a terser way to do this instead of re-implementing every function. Maybe proxies but that looks like it'll end up being the same if not more code. Some sort of dynamic function call reference.

PR #673, opened 2021-04-21
fixed: check for the presence of data before generating a custom message

shrunyan:
- check for data before generating message 
- removed default case as that is handled with an or statement in the jsx

### DIFF
--- a/src/apps/content-editor/src/app/views/Dashboard/components/UserLatest/UserLatest.js
+++ b/src/apps/content-editor/src/app/views/Dashboard/components/UserLatest/UserLatest.js
@@ -2,6 +2,7 @@ import React, { useState, useEffect } from "react";
 import { useDispatch } from "react-redux";
 import moment from "moment";
 import uniqBy from "lodash/uniqBy";
+import cx from "classnames";
 import zuid from "zuid";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 import { faChevronRight, faClock } from "@fortawesome/free-solid-svg-icons";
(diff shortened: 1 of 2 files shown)


src/apps/content-editor/src/app/views/Dashboard/components/UserLatest/UserLatest.js
```diff
@@ -48,20 +48,23 @@ export function UserLatest(props) {
...
             return dispatch(searchItems(log.affectedZUID)).then(item => {
-              if (log.action === 2) {
-                log.recentTitle = `Modified Content Item ${item.data[0]?.web?.metaTitle}`;
-              } else if (log.action === 4) {
-                log.recentTitle = `Published Content Item ${item.data[0]?.web?.metaTitle}`;
-              } else {
-                log.recentTitle = log.meta.message;
+              if (item?.data[0]?.web?.metaTitle) {
+                if (log.action === 2) {
+                  log.recentTitle = `Modified Content Item ${item.data[0]?.web?.metaTitle}`;
```
nibblebot, line 53: can get rid of null operator here and line 55 since we have the new check on line 51

src/apps/content-editor/src/app/views/Dashboard/components/UserLatest/UserLatest.js
```diff
@@ -48,20 +48,23 @@ export function UserLatest(props) {
...
+
               return log;
             });
 
           // Display model labels
           case zuid.prefix.SITE_CONTENT_SET:
             return dispatch(fetchModel(log.affectedZUID)).then(model => {
-              log.recentTitle = `Modified Schema ${model.payload?.label}`;
+              if (model?.payload?.label) {
+                log.recentTitle = `Modified Schema ${model.payload?.label}`;
```
nibblebot, line 66: no need for null check

PR #677, opened 2021-04-23
Hotfix/robot.js https

d88naimi:
refactored to use domain in iframe with custom https


src/apps/settings/src/app/views/Robots/Robots.js
```diff
@@ -20,6 +20,7 @@ import styles from "./Robots.less";
 
 export default connect(state => {
   return {
+    domain: state.instance.domain,
```
shrunyan, line 23: This should be removed right? Because we are using the `useDomain` hook vs this value.

src/apps/settings/src/app/views/Robots/Robots.js
```diff
@@ -179,7 +180,7 @@ export default connect(state => {
           </h2>
           <iframe
             className={styles.Iframe}
-            src={`${robotURL}?q=${Math.random()
+            src={`https://${robotURL.slice(7)}?q=${Math.random()
```
shrunyan, line 183: I would move this up into the functional component body so we can include a comment which explains what is happening and why it needs to be done.

src/apps/settings/src/app/views/Robots/Robots.js
```diff
@@ -47,6 +47,11 @@ export default connect(state => {
 
   const robotURL = `${domain}/robots.txt`;
 
+  //https url required for iframe
```
shrunyan, line 50: Lets describe why it is required for the iframe, for future reference.

shrunyan (approved):

PR #691, opened 2021-05-06
fix: Internal/External link options use redux/memo

nibblebot:


shrunyan:
@nibblebot please provide a summary of what these changes correct or add within the experience. Screenshots would also be great.

PR #701, opened 2021-05-12
adding optional chaining for undefined nav

d88naimi:
Sentry bug 
TypeError
e.props.nav.find(...) is undefined

### DIFF
--- a/src/apps/content-editor/src/app/components/ReorderNav/DragComponents/DraggableItem.js
+++ b/src/apps/content-editor/src/app/components/ReorderNav/DragComponents/DraggableItem.js
@@ -3,18 +3,16 @@ import React, { Component } from "react";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 import { faArrowsAlt, faCaretRight } from "@fortawesome/free-solid-svg-icons";
 
-import { AppLink } from "@zesty-io/core/AppLink";
-
 import styles from "./styles.less";
 export default class DraggableItem extends Component {
   render() {
-    const { ZUID, label, children, handleNestChange } = this.props;
+    const { ZUID, label, children, handleNestChange, ...rest } = this.props;
     return (
-      <li {...this.props}>
+      <li {...rest}>
         <FontAwesomeIcon icon={faArrowsAlt} />
 
         {children.length ? (
-          <AppLink
+          <span
             className={styles.node}
             onClick={evt => {
               evt.preventDefault();
@@ -23,7 +21,7 @@ export default class DraggableItem extends Component {
           >
             {label}
             <FontAwesomeIcon icon={faCaretRight} />
-          </AppLink>
+          </span>
         ) : (
           <span>{label}</span>
         )}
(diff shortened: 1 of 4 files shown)


src/apps/content-editor/src/app/components/ReorderNav/ReorderNav.js
```diff
@@ -72,7 +72,7 @@ class ReorderNav extends Component {
       });
     }
     this.setState({
-      [name]: [...this.props.nav.find(item => item.ZUID === name).children],
+      [name]: [...this?.props?.nav.find(item => item.ZUID === name).children],
```
shrunyan, line 75: I do not think this will solve the  core issue from Sentry, `e.props.nav.find(...) is undefined`. As the message indicates the result from the `find` function is what is undefined. Which ultimately makes the `.children` reference in valid.

shrunyan (changes requested):
Additional in DragList.js:23 there is a reference to an `item.ZUID` which can crash on a nesting level with a single item. The `map` operation for this callback should be preceded by a `filter` which ensures the presence of an item.

d88naimi:
Resolved broken draggable item 
Resolved React Warnings referenced https://reactjs.org/warnings/unknown-prop.html



`Warning: React does not recognize the `ZUID` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `zuid` instead. If you accidentally passed it from a parent component, remove it from the DOM element.`

`Warning: React does not recognize the `handleNestChange` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `zuid` instead. If you accidentally passed it from a parent component, remove it from the DOM element.`

src/apps/content-editor/src/app/components/ReorderNav/ReorderNav.js
d88naimi: I fixed the Draglist issue and hopefully that may resolve the `children` bug.

src/apps/content-editor/src/app/components/ReorderNav/DragComponents/DraggableItem.js
```diff
@@ -3,18 +3,16 @@ import React, { Component } from "react";
...
   render() {
-    const { ZUID, label, children, handleNestChange } = this.props;
+    const { ZUID, label, children, handleNestChange, ...rest } = this.props;
     return (
-      <li {...this.props}>
+      <li {...rest}>
         <FontAwesomeIcon icon={faArrowsAlt} />
 
         {children.length ? (
-          <AppLink
```
shrunyan, line 17: What is the reason we are choosing to not use an `AppLink` component here?

shrunyan (changes requested):
This PR no longer seems to add optional chaining. Please update the original PR title and description to explain what the solution is.

src/apps/content-editor/src/app/components/ReorderNav/DragComponents/DraggableItem.js
d88naimi: The draggable component cannot recognize the  AppLink `a` tag thus causing on drag to crash the app.  The span allows the the element to be nested in the list parent and reorder/drag works properly.

d88naimi (commented):
@stuart I believe after resolving the Applink issue reorder nav should not cause an app crash and .`children` may be read correctly.

shrunyan (approved):

PR #702, opened 2021-05-12
feat: Dynamic width for global tabs based on mix/max/avail space

nibblebot:
__UI changes__
- dynamic width tabs with a min and max width similar to Chrome tabs.
- show tab close buttons all the time 
- remove tab transitions on all properties
- remove color transition on Tab > AppLink

__Behavior changes__
- new tabs open up to right of current tab
- closing tabs will switch to next tab. If no next tab, navigate to prev tab. if no prev tab, do nothing
- Deleting a Media Group now calls closeTab() to cleanup after itself

__Refactor__
- move unrelated components from GlobalTabs to new GlobalTopBar component
- move tab state from component to redux slice so we can call closeTab() from anywhere in the app
- move history from sentry to utility so we can import history into redux slice for use in action thunk
- switch from BrowserRouter to Router so we can pass our custom history

__Extra__
- Why Did You Render debugging package (commented out in code)


shrunyan:
I think we should keep the drop shadow which overlaps the final tab close button. As a visual it indicates there are more and that the tabs have disappeared behind the account fly out. Otherwise the tab aligns perfectly to it and feels as though there are no more.  Once tabs overflow a user will need to scroll in order to get a tab into view to be able to close it.

shrunyan (approved):

PR #706, opened 2021-05-19
refactor: render App after IndexedDB load

nibblebot:
Load IndexedDB caches before App is rendered.

If we only render App once all IndexedDB loads, we don't need to store state like `loadedTabs` for GlobalTabs component to conditionally do stuff.

One possible change is to move the `render()` into a `finally{}` clause so user can use the App even with IndexedDB failure.


shrunyan:
This creates a hard requirement on a browser supporting IndexedDB. Is this something we want? What does modern browser support of IndexedDB look like? What browsers would have issues? Mobile?

shrunyan:
Lets move the render into the finally clause

PR #712, opened 2021-05-24
Feature/#454 nav enhancement

### LINKED ISSUE #454: Stack name and icon in manager UI nav on smaller screens
I have felt frustrated with the new UI because on my laptop I only see the icons in the nav and not the name of the section eg: Content, Schema, Media, etc. It's been difficult really explain the nav as I am updating docs as there are just icons there. I didn't know that names had been added until I happened to zoom out today. I believe I'll need to re-do the images in the docs to better reflect the nav at a larger screen size.

On smaller screens, such as my 13" laptop, I would like the nav to have the icon stacked on top of the name of the section.

I've tried to memorize what icon belongs to what section but there are still a few that it takes time to remember. I understand that there are tooltips for these icons, however, I have to hover on the icon, wait for the tooltip to appear - I don't think it's a great user experience. 

**Additional context**

manager-ui-nav

d88naimi:
Nav open and close enhancement 
fixes #454 
branch `feature/#454-nav-enhancement`

Screen Shot 2021-05-24 at 1 59 08 PM

Screen Shot 2021-05-24 at 1 59 00 PM


shrunyan:
I think the colors on the sub-menu should all remain muted. Besides the collapse action.

d88naimi:
Reverted subnav back to zesty gray colors 
Screen Shot 2021-05-25 at 10 53 09 AM

shrunyan:
The icon font alignment seems off. Like it's baselined.

src/shell/views/Shell/Shell.less
```diff
@@ -24,18 +24,38 @@
...
-  scrollbar-width: thin;
+  //CLOSED MENU
+  &.NavClosed {
+    grid-template-columns: 60px 1fr;
 
-  @media (min-width: 2000px) {
-    & {
-      grid-template-columns: 170px 1fr;
-      font-size: 14px;
+    a[href="/content"],
```
shrunyan, line 33: please use a class to target these

src/shell/views/Shell/Shell.less
```diff
@@ -24,18 +24,38 @@
...
+    }
+    [data-icon="eye"] + span {
+      display: none;
+    }
+    [data-icon="hashtag"] {
+      display: none;
     }
   }
 
+  scrollbar-color: @zesty-grey @zesty-light-grey;
```
shrunyan, line 56: these are defined at the global app and cascade. do these values differ?

shrunyan (changes requested):

src/shell/components/global-actions/index.js
```diff
@@ -35,32 +36,36 @@ export default React.memo(function GlobalActions(props) {
...
         <Url
           href={`https://zesty.io`}
-          title="View source code commit"
+          title="Zesty.io"
           target="_blank"
           className={styles.GlobalAction}
         >
           <img
             src="https://brand.zesty.io/zesty-io-logo.svg"
             alt={`Zesty.io`}
```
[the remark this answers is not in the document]
d88naimi: Not that I am aware of, removed jsx block.

src/shell/views/Shell/Shell.less
d88naimi: @shrunyan these live inside NavClose class I am not 100% understanding the action on this one?

src/shell/views/Shell/Shell.less
d88naimi: @shrunyan Im not 100% following the action on this one as well. Is this referring to scrollbar values?

shrunyan: Yes scrollbar values. What's the reason for defining these? These styles should already exist at the shell and cascade to this area.

src/shell/views/Shell/Shell.less
shrunyan: These selectors use href attributes to target the elements. So we have to maintain this list going forward. Can we use a class instead?

shrunyan (approved):

src/shell/views/Shell/Shell.less
d88naimi: Yes, these existed already here in shell level, I relocated them back to there proper place after looking at previous master https://github.com/zesty-io/manager-ui/blob/f611e263b994f145c3b12fde517af6e7ae0353e8/src/shell/views/Shell/Shell.less#L29-L30

src/shell/views/Shell/Shell.less
d88naimi: Since we are at shell top level I specifically used these selectors so I wouldn't have a cascading affect on other `a` tags

PR #713, opened 2021-05-24
added missing value prop and added clear input after onclick submit

### LINKED ISSUE #697: SEO redirect creation form isn't reset after redirects are created.
**Describe the bug**
After a redirect is created the form does not reset. It becomes a confusing experience  because it looks like the redirect wasn't created, when in fact it was. 

**To Reproduce**
1. Go to SEO
2. Enter a relative path in the From field. And select a target from the To dropdown.
3. Click Create Redirect.
4. See that even though the redirect is created the form does not reset. 

**Expected behavior**
When redirects are successfully created the form should reset (previous values should be removed).

**Screenshots**

Screen Shot 2021-05-11 at 3 02 05 PM

 - OS: Mac
 - Browser Chrome
 - Version 89??

**Additional context**

d88naimi:
Input fields now  reset after submission

fixes #697

### DIFF
--- a/src/apps/seo/src/store/redirects.js
+++ b/src/apps/seo/src/store/redirects.js
@@ -78,7 +78,7 @@ export function createRedirect(redirect) {
           dispatch(
             notify({
               kind: "success",
-              message: `Created new redirect`
+              message: `Created redirect`
             })
           );
 
@@ -94,7 +94,7 @@ export function createRedirect(redirect) {
           dispatch(
             notify({
               kind: "error",
-              message: "Failed to create SEO redirect"
+              message: "Failed creating redirect"
             })
           );
         }
@@ -103,7 +103,7 @@ export function createRedirect(redirect) {
         dispatch(
           notify({
             kind: "error",
-            message: "Failed to create SEO redirect"
+            message: "Failed creating redirect"
           })
         );
       });
--- a/src/apps/seo/src/store/redirects.js
+++ b/src/apps/seo/src/store/redirects.js
@@ -68,13 +68,20 @@ export function createRedirect(redirect) {
     dispatch({
       type: "REDIRECT_CREATE"
     });
-    request(`${CONFIG.API_INSTANCE}/web/redirects`, {
+    return request(`${CONFIG.API_INSTANCE}/web/redirects`, {
       method: "POST",
       json: "true",
       body: redirect
     })
       .then(json => {
         if (!json.error) {
+          dispatch(
+            notify({
+              kind: "success",
+              message: `Created new redirect`
+            })
+          );
+
           dispatch({
             type: "REDIRECT_CREATE_SUCCESS",
             redirect: {
--- a/src/apps/seo/src/views/RedirectsManager/RedirectsTable/RedirectCreator/RedirectCreator.js
+++ b/src/apps/seo/src/views/RedirectsManager/RedirectsTable/RedirectCreator/RedirectCreator.js
@@ -26,6 +26,8 @@ export function RedirectCreator(props) {
         code: code === 1 ? 301 : 302 // API expects a 301/302 value
       })
     );
+    setFrom("");
+    setTo("");
   };
 
   return (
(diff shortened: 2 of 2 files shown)


src/apps/seo/src/views/RedirectsManager/RedirectsTable/RedirectCreator/RedirectCreator.js
```diff
@@ -26,6 +26,8 @@ export function RedirectCreator(props) {
         code: code === 1 ? 301 : 302 // API expects a 301/302 value
       })
     );
```
shrunyan, line 28: Assuming the `createRedirect` is returning the fetch promise this should be a thenable value. Lets put the reset into a `then` clause. As well as lets ensure a system notification is being shown on when a bad request or error occurs. This way on a bad request we don't wipe the fields and a user can make a correction to try again.

src/apps/seo/src/views/RedirectsManager/RedirectsTable/RedirectCreator/RedirectCreator.js
```diff
@@ -36,6 +38,7 @@ export function RedirectCreator(props) {
           name="redirectFrom"
           type="text"
           placeholder="URL path to redirect from"
+          value={from}
```
shrunyan, line 41: If this wasn't present before how was the value showing in the field? What does adding this to the field provide us?

shrunyan (changes requested):

src/apps/seo/src/views/RedirectsManager/RedirectsTable/RedirectCreator/RedirectCreator.js
d88naimi: add then(), added notify success to the store for successful redirects.

src/apps/seo/src/views/RedirectsManager/RedirectsTable/RedirectCreator/RedirectCreator.js
d88naimi: I believe the input field was listening on line 45 `evt.target.value` but the input field does not clear after the submit if I dont have line 41 inserted `value={from}`. Please let me know if I am going about this incorrectly thanks.

shrunyan: Is there a cursor jump bug now that this is a controlled component?

src/apps/seo/src/store/redirects.js
```diff
@@ -68,13 +68,20 @@ export function createRedirect(redirect) {
...
       method: "POST",
       json: "true",
       body: redirect
     })
       .then(json => {
         if (!json.error) {
+          dispatch(
+            notify({
+              kind: "success",
+              message: `Created new redirect`
```
shrunyan, line 81: We should drop "new" as it's superfluous. "Created" implies it's new.

shrunyan: For the errors later we should match the creation structure. e.g.
Change: "Failed to create SEO redirect"
To: "Failed creating redirect"

shrunyan (changes requested):

src/apps/seo/src/views/RedirectsManager/RedirectsTable/RedirectCreator/RedirectCreator.js
d88naimi: No input is cleared, here is a quick video for both working and non working links
https://www.loom.com/share/6e3cd2fdf67242bfbde8ac34d36e7249

d88naimi: @shrunyan there is not a cursor jump on the `From` section but there is in `To` section which is happening currently on master prior to my fixes.

shrunyan: Ok. Would you split that bug of into a new ticket?

PR #714, opened 2021-05-25
adding optional chaining to data

d88naimi:
Sentry bug: MANAGER-UI-5V

TypeError
Cannot read property 'data' of undefined
/src/apps/code-editor/src/app/components/FileDrawer/FileDrawer.js at line 54:16

### DIFF
--- a/src/apps/code-editor/src/app/components/FileDrawer/FileDrawer.js
+++ b/src/apps/code-editor/src/app/components/FileDrawer/FileDrawer.js
@@ -51,7 +51,7 @@ export const FileDrawer = React.memo(function FileDrawer(props) {
 
         // Logs should always exist
         setLogs(
-          logs.data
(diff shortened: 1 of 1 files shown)


src/apps/code-editor/src/app/components/FileDrawer/FileDrawer.js
```diff
@@ -51,7 +51,7 @@ export const FileDrawer = React.memo(function FileDrawer(props) {
 
         // Logs should always exist
         setLogs(
-          logs.data
+          logs?.data
```
shrunyan, line 54: If be concerned with the comment on line 52. It seems to think logs would have always existed and rely on that. If they aren't there then `logs?` optional chaining would set undefined which may break other expectations.

crzestyio (approved):

PR #729, opened 2021-06-03
added NotFound componenet for broken error handling

d88naimi:
fixes sentry Error MANAGER-UI-MJ
No Results Found for ZUID: 7-255694-4gmw9w2

Screen Shot 2021-06-03 at 11 35 04 AM


shrunyan (commented):
When switching to another item the `notFound` seems like it would not reset and it would continue to show a not found component

crzestyio (approved):

PR #745, opened 2021-06-11
updating error notice with more description

### LINKED ISSUE #696: SEO - add more verbose error notifications
**Describe the bug**
When a redirect fails to create I don't know why cause the error isn't verbose.  

**To Reproduce**
1. Go to SEO
2. Create a redirect FROM http://www.example.com/  -> TO /
3. Click Create Redirect
4. See error in bottom right-hand corner

**Expected behavior**
I expect errors to be more verbose eg: 
Redirect couldn't be created. From xxx [insert value] is invalid. Must be a relative path.

**Screenshots**
Screen Shot 2021-05-11 at 2 45 54 PM

 - OS: Mac
 - Browser Chrome
 - Version 89??

**Additional context**
other SEO issue https://github.com/zesty-io/manager-ui/issues/695
If a user tries to enter an invalid From path they should get a verbose error too.

d88naimi:
fixes #696 

Screen Shot 2021-06-11 at 1 44 37 PM

### DIFF
--- a/src/apps/seo/src/store/redirects.js
+++ b/src/apps/seo/src/store/redirects.js
@@ -64,7 +64,6 @@ export function fetchRedirects() {
 }
 
 export function createRedirect(redirect) {
-  console.log(redirect.path);
   return dispatch => {
     dispatch({
       type: "REDIRECT_CREATE"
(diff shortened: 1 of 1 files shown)


src/apps/seo/src/store/redirects.js
```diff
@@ -64,6 +64,7 @@ export function fetchRedirects() {
 }
 
 export function createRedirect(redirect) {
+  console.log(redirect.path);
```
nibblebot, line 67: remove console.log

nibblebot (approved):

crzestyio (approved):

PR #747, opened 2021-06-14
fix most cypress tests; use cypress.json

nibblebot:


d88naimi (approved):

d88naimi:
There are still some broken test however this is still a move forward from the current test.

PR #748, opened 2021-06-14
ismounted hook

nibblebot:
for use in functional components. Whenever you do an async action, after it comes back you need to check if the component is still mounted before trying to setState


d88naimi (approved):

d88naimi:
Can I get a quick description of why we are utilizing the ismounted hook for future reference?

PR #767, opened 2021-06-22
adding publishing_version=2 to content publish request NO MERGE!!

d88naimi:


shrunyan:
What is the purpose of this PR? Is this query param supported?

d88naimi:
Incident Report from 6/21/21 : Front end changes made a PR that points Publish request API to a back up API that was never used since backend managed to resolve kubes cluster. I will close this out.

PR #768, opened 2021-06-22
 swapped out request from site manager to instance api

d88naimi:


d88naimi:
POST Unpublished Item API 
https://instances-api.zesty.org/#963c883e-4262-4f48-8b6a-24a1d23d6b15

src/shell/store/content.js
```diff
@@ -605,9 +605,9 @@ export function unpublish(modelZUID, itemZUID, publishZUID, options = {}) {
     }
 
     return request(
-      `${CONFIG.LEGACY_SITES_SERVICE}/${instance.ZUID}/content/items/${itemZUID}/publish-schedule/${publishZUID}`,
+      `${CONFIG.API_INSTANCE}/content/models/${modelZUID}/items/${itemZUID}`,
```
nibblebot, line 608: this doesn't look like an unpublish endpoint, is this correct? `/content/models/$modelZUID/items/$itemZUID`

d88naimi: @nibblebot I followed the Unpublished API from here: 
https://instances-api.zesty.org/#963c883e-4262-4f48-8b6a-24a1d23d6b15

Screen Shot 2021-06-22 at 9 55 26 AM

nibblebot: ok thanks, i'm seeing the documentation for the endpoint says it takes an `unpublish` body but we are using `take_offline_at` below

PR #787, opened 2021-06-30
making edit code link goto correct file path

### LINKED ISSUE #764: Schema linking Edit Code to correct file
Schema > Model Info > Edit Code link goes to IDE 

Would lIke Edit Code link to goto correct IDE file. 

Example: Edit a code file and see dashboard your latest edits `View` link

schema-linking

d88naimi:
fixes #764 
Lifting fetchFIles into load-instance since it being utilized in three places now.

Now allowing Schema => model info => edit code link to redirect to relative code path


src/shell/hooks/use-filePath.js
```diff
@@ -0,0 +1,49 @@
+import zuid from "zuid";
```
nibblebot, line 1: the filename for this file, `use-filePath` is a mix of kebab-case and camelCase. Let's pick one, I prefer camelCase.

src/shell/hooks/use-filePath.js
```diff
@@ -0,0 +1,49 @@
...
+import { useSelector } from "react-redux";
+import { createSelector } from "reselect";
+
+export function useFilePath(ZUID = "") {
+  const getFiles = state => state.files;
+  const selectFiles = createSelector([getFiles], files => files);
```
nibblebot, line 7: since the output selector is the identity function (i.e. x => x), `selectFiles` is redundant, and we can just use useSelector(getFiles)

src/shell/components/load-instance/index.js
```diff
@@ -79,7 +86,8 @@ export default connect(state => {
               props.instance.ID &&
               props.instance.domains &&
               props.user.ID &&
-              props.languages.length
+              props.languages.length &&
+              props.files
```
nibblebot, line 90: `props.files` will always be true since the store defaults to `[]` which is truthy
should we enforce `props.files.length` or do users sometimes not have any files?

src/shell/components/load-instance/index.js
```diff
@@ -56,6 +59,10 @@ export default connect(state => {
       props.dispatch(fetchLangauges("enabled"));
       // Used in Publish Plan and Content sections
       props.dispatch(fetchItemPublishings());
+      //Code Editor
```
nibblebot, line 62: also used in `useFilePath` hook

nibblebot (changes requested):

src/shell/components/load-instance/index.js
d88naimi: added `props.files.length`

src/shell/components/load-instance/index.js
d88naimi: added to comments

src/shell/components/load-instance/index.js
```diff
@@ -56,6 +59,10 @@ export default connect(state => {
       props.dispatch(fetchLangauges("enabled"));
       // Used in Publish Plan and Content sections
       props.dispatch(fetchItemPublishings());
+      // Used in Code Editor, Publish, PublishAll, useFilePath hook
```
nibblebot, line 62: no need to comment about Publish/Publish All, they have there own invocations of `fetchFiles`

src/shell/components/load-instance/index.js
```diff
@@ -79,7 +86,8 @@ export default connect(state => {
               props.instance.ID &&
               props.instance.domains &&
               props.user.ID &&
-              props.languages.length
+              props.languages.length &&
+              props.files.length
```
nibblebot, line 90: @shrunyan do all instances have at least one file? can we make that a loading blocker?

shrunyan: Theoretically yes, the loader, but it is possible to delete all files.

src/shell/hooks/useFilePath.js
```diff
@@ -0,0 +1,47 @@
...
+    case zuid.prefix.SITE_CONTENT_SET:
+      let findfile = files.find(file => file.contentModelZUID === ZUID);
+      if (findfile) {
+        return `/code/file/views/${findfile.ZUID}`;
+      } else {
+        return "/code";
+      }
+
+    case zuid.prefix.SITE_CONTENT_ITEM:
+      let findFile = files.find(file => file.modelZUID === ZUID);
```
shrunyan, line 21: This logic is invalid. It would compare a model zuid (6-) against a provided item zuid (7-)

shrunyan (changes requested):

src/shell/hooks/useFilePath.js
d88naimi: Removed SITE_CONTENT_ITEM since all use cases for edit code link use SITE_RESOURCE to check relative file path

shrunyan (approved):

PR #797, opened 2021-07-02
Chore: Babel automatic jsx codemod

nibblebot:
Implements new preferred style for babel jsx import. Upgrade instructions here: https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html
comes with a codemod script that updates all the old import statements and prefers destructed imports on React.


shrunyan:
@nibblebot this seems to contain prettier code mods on each file as well. Would you separate these changes by setting up a PR which is just the latest version of prettier ran via cli across the repo?

shrunyan:
Conflicts after prettier 2.0 mod across codebase

PR #801, opened 2021-07-06
adding error handling

d88naimi:
Sentry : MANAGER-UI-PB

https://sentry.io/organizations/zestyio/issues/2489977620/?project=5441698&referrer=AssignedActivityEmail


src/shell/components/ContentSearch/index.js
```diff
@@ -47,6 +47,7 @@ export default React.forwardRef((props, providedRef) => {
       if (term) {
         dispatch(searchItems(term))
           .then((res) => {
+            if (res.status !== 200) return;
```
shrunyan, line 50: there is already an else condition if not a 200 status code

shrunyan (changes requested):

PR #804, opened 2021-07-07
adding usefilepath look up on a 7-

d88naimi:


src/shell/hooks/useFilePath.js
```diff
@@ -3,9 +3,13 @@ import { useSelector } from "react-redux";
...
 
 export function useFilePath(ZUID = "") {
-  const getFiles = state => state.files;
-  const selectFiles = createSelector([getFiles], files => files);
+  const getFiles = (state) => state.files;
+  const selectFiles = createSelector([getFiles], (files) => files);
   const files = useSelector(selectFiles);
+
+  const getContent = (state) => state.content;
+  const content = useSelector(getContent);
```
shrunyan, line 11: You seem to be missing the selector creation.

src/shell/hooks/useFilePath.js
```diff
@@ -14,15 +18,31 @@ export function useFilePath(ZUID = "") {
...
+      let findfile = files.find((file) => file.contentModelZUID === ZUID);
       if (findfile) {
         return `/code/file/views/${findfile.ZUID}`;
       } else {
         return "/code";
       }
 
+    // Searching on Content Item Zuid 7-
+    case zuid.prefix.SITE_CONTENT_ITEM:
+      let findItemZUID = Object.keys(content).find((file) => file === ZUID);
```
shrunyan, line 30: If content items is an object keyed by ZUID then we should be able to do a direct item look up. e.g. 
```
const item = content[ZUID]
```

You will need to confirm the item is present in full as we pre-populate them when fetching publish records. I.e. you'll need to confirm the `item.meta.contentModelZUID` is present before finding it's corresponding file match.

shrunyan (changes requested):

src/shell/hooks/useFilePath.js
```diff
@@ -14,15 +18,31 @@ export function useFilePath(ZUID = "") {
...
+      let findMatch = Object.values(content).filter(
+        (file) => file.meta.ZUID === findItemZUID
+      );
+      let findFile = files.find(
+        (file) => file.contentModelZUID === findMatch[0].meta.contentModelZUID
+      );
+
+      if (ZUID === findItemZUID && findFile) {
```
shrunyan, line 38: It is only necessary to confirm a file was found. If so we know it's ZUID will be present on that object.

shrunyan (changes requested):

src/shell/hooks/useFilePath.js
```diff
@@ -3,26 +3,46 @@ import { useSelector } from "react-redux";
...
 
 export function useFilePath(ZUID = "") {
-  const getFiles = state => state.files;
-  const selectFiles = createSelector([getFiles], files => files);
+  const getFiles = (state) => state.files;
+  const selectFiles = createSelector([getFiles], (files) => files);
   const files = useSelector(selectFiles);
+
+  const getContent = (state) => state.content;
+  const selectContent = createSelector([getContent], (content) => content);
```
nibblebot, line 11: I would just inline these selectors: 
```
const files = useSelector(state => state.files)
const content = useSelector(state => state.content)
```
It doesn't make sense to createSelectors in a hook. Usually if you create custom memoized selectors you would hoist it outside the component/hook so it's not created each time.

src/shell/hooks/useFilePath.js
```diff
@@ -3,26 +3,46 @@ import { useSelector } from "react-redux";
...
         return `/code/file/views/${findfile.ZUID}`;
       } else {
         return "/code";
       }
 
+    // Searching on Content Item ZUID
+    case zuid.prefix.SITE_CONTENT_ITEM:
+      const item = content[ZUID];
+
+      let findFile = files.find(
```
nibblebot, line 34: would move this inside the if.
would also use `const` here
```javascript
if (item.meta.contentModelZUID) {
  const file = files.find(...)
  if (file) {
    return `/code/file/views/${findFile.ZUID}`;
  }
}
return "/code";
```

nibblebot (changes requested):

shrunyan (approved):

PR #805, opened 2021-07-08
adding conditional check for setColumns onSort

d88naimi:
fixing typeError e.props.onSort is not a function

https://sentry.io/organizations/zestyio/issues/2489996110/?project=5441698&referrer=AssignedActivityEmail


nibblebot (changes requested):
you are calling `onSort` twice, remove the first one

nibblebot (changes requested):
You are still calling the `onSort` function twice, the line you added should not execute the function but check if it is truthy

shrunyan (approved):

nibblebot (approved):

PR #842, opened 2021-07-23
adding a keyboard shortcut hook, adding a code drawer open/close shrtcut

d88naimi:
Adding a keyboard shortcut hook
Adding code drawer open/close shortcut

Screen Shot 2021-07-23 at 1 34 15 PM
Screen Shot 2021-07-23 at 1 34 23 PM


src/shell/hooks/useKeyboardShortcut.js
```diff
@@ -0,0 +1,36 @@
+import { useEffect } from "react";
+import { useSelector } from "react-redux";
+
+/*
+Usage Example
+  useKeyboardShortcut("letter", () => callback());
+
+*/
+
+export function useKeyboardShortcut(key, callback) {
```
shrunyan, line 10: Lets rename this to `useMetaKey`. 

The context of whether its a shortcut or not is irrelevant to the hook it's self.

src/apps/code-editor/src/app/components/FileDrawer/FileDrawer.js
```diff
@@ -104,6 +110,15 @@ export const FileDrawer = memo(function FileDrawer(props) {
         {open ? null : (
           <span className={styles.bodyText}>More file information</span>
         )}
+        {open ? (
+          <span className={cx(styles.bodyText, styles.DrawerShortcut)}>
+            ({platform.isMac ? "CMD" : "CTRL"} + D) Close Drawer
```
shrunyan, line 115: We could move the logic of `platform.isMac ? "CMD" : "CTRL"` into the hook and have it return that value so users of the hook do not need to handle this aspect. It would be a little awkward from a function signature and return value perspective but the utility is very convenient.

src/apps/code-editor/src/app/components/FileDrawer/FileDrawer.js
```diff
@@ -82,9 +85,12 @@ export const FileDrawer = memo(function FileDrawer(props) {
 
   function handleSetOpen() {
     // TODO persist to user settings
-    setOpen(!open);
+    setOpen((open) => !open);
   }
 
+  // Handle keyboard shortcut Drawer
+  useKeyboardShortcut("d", () => handleSetOpen());
```
shrunyan, line 92: The callback does not need to be wrapped in a function. It should be passed directly.

src/apps/code-editor/src/app/components/FileDrawer/FileDrawer.js
```diff
@@ -82,9 +85,12 @@ export const FileDrawer = memo(function FileDrawer(props) {
 
   function handleSetOpen() {
     // TODO persist to user settings
-    setOpen(!open);
+    setOpen((open) => !open);
   }
 
+  // Handle keyboard shortcut Drawer
```
shrunyan, line 91: This comment does not add any additional context which the code does not already clearly express.

src/apps/code-editor/src/app/components/FileDrawer/FileDrawer.js
```diff
@@ -21,6 +23,7 @@ import { fetchFields } from "shell/store/fields";
 import styles from "./FileDrawer.less";
 export const FileDrawer = memo(function FileDrawer(props) {
   const dispatch = useDispatch();
+  const platform = useSelector((state) => state.platform);
```
shrunyan, line 26: This will be unnecessary if the platform key logic is moved into the hook

src/shell/hooks/useKeyboardShortcut.js
```diff
@@ -0,0 +1,36 @@
...
+      if (
+        (platform.isMac && evt.metaKey === true && evt.key === key) ||
+        (!platform.isMac && evt.ctrlKey === true && evt.key === key)
+      ) {
+        evt.preventDefault();
+        callback();
+      }
+    }
+    window.addEventListener("keydown", onKeyDown);
+    return () => window.removeEventListener("keydown", onKeyDown);
```
shrunyan, line 34: We need to confirm this clean up logic gets ran when components are unmounted. Otherwise we will have a memory leak.

src/shell/hooks/useKeyboardShortcut.js
```diff
@@ -0,0 +1,36 @@
...
+import { useSelector } from "react-redux";
+
+/*
+Usage Example
+  useKeyboardShortcut("letter", () => callback());
+
+*/
+
+export function useKeyboardShortcut(key, callback) {
+  //Get OS
```
shrunyan, line 11: Comment does not add any additional context

shrunyan (changes requested):

src/apps/code-editor/src/app/components/FileDrawer/FileDrawer.js
d88naimi: I thought this was best practice to prevent bugs for asynchronous code, if I remove the callback then my code doesn't work.

src/apps/content-editor/src/app/views/ItemEdit/ItemEdit.js
```diff
@@ -79,25 +80,9 @@ export default function ItemEdit() {
...
-          save();
-        }
-      }
-    };
-    window.addEventListener("keydown", handleSaveKeyboardShortcut);
-    return () => {
-      window.removeEventListener("keydown", handleSaveKeyboardShortcut);
-    };
-  }, [item]);
+  useMetaKey("s", () => {
```
shrunyan, line 83: superfluous wrapping arrow function

src/apps/media/src/app/components/MediaDetailsModal.js
```diff
@@ -1,6 +1,8 @@
-import { memo, useRef, useState, useEffect } from "react";
+import { memo, useRef, useState, useCallback } from "react";
```
shrunyan, line 1: Is `useCallback` being used?

src/shell/hooks/useMetaKey.js
```diff
@@ -0,0 +1,36 @@
...
+
+/*
+Usage Example
+  useMetaKey("letter", callback);
+*/
+
+export function useMetaKey(key, callback) {
+  const platform = useSelector((state) => state.platform);
+
+  if (!key)
```
shrunyan, line 12: fix these `if` statements to match our code styling

shrunyan (changes requested):

shrunyan:
Please thoroughly test these 2 secarios
- [ ] Does the `useEffect` logic inside the hook get called, cleaning up the event listener, when the consuming component unmounts?
- [ ] Does consuming the hook twice in the same view, or even the same component, attach to handlers and both callbacks are triggered with the same key press?

shrunyan (approved):

PR #861, opened 2021-08-07
reverted defaultValue back to value for auto populate to work and also prevent cursor jump

### LINKED ISSUE #703: Schema: field reference names do not autopopulate
**Describe the bug**
field reference names no longer auto populate and I have to type them in manually. 

**To Reproduce**
1. Go to Schema
2. Click on any existing model and add a field.
3. When you type in the Field Label notice that the Field Name does not autopopulate.

**Expected behavior**
I expect the Field Name to populate with whatever I type into the Field Label field. 

**Screenshots**
Screen Shot 2021-05-17 at 11 48 15 AM

 - OS: Mac
 - Browser Chrome
 - Version 90

**Additional context**
I _think_ that this only started to happen recently. I don't know if this change was intended. I think that making user enter the reference name leave space for mistakes that we should eliminate.

d88naimi:
fixes #703 

Screen Shot 2021-08-06 at 6 08 53 PM

Screen Shot 2021-08-06 at 6 08 33 PM

Design systems PR https://github.com/zesty-io/design-system/pull/144

Resolve autopopulate and cursor jump with adding a useRef. 

CC: @ardeay


cypress/integration/schema/model_spec.js
```diff
@@ -23,6 +23,22 @@ describe("Schema", () => {
...
     cy.get('button[kind="save"]').contains("Add Field").click();
 
+    cy.get('input[name="label"]').type(
+      "{leftarrow}{leftarrow}{leftarrow}{leftarrow}{leftarrow} inserting  text within {rightarrow} test",
+      {
+        force: true,
+      }
+    );
+
+    cy.get('input[name="name"]').type(
```
shrunyan, line 33: For the test we want to check whether the `name` field has been set to the computer friendly version of the value entered into the `label` field.

shrunyan (changes requested):

d88naimi:
Updated new PR here with out cypress conflicts https://github.com/zesty-io/manager-ui/pull/970

PR #870, opened 2021-08-18
Feature/#643 seo wildcard

### LINKED ISSUE #643: SEO: Wildcard redirects
WebEngine now supports wildcard redirects, as well as 3 types of redirects `path`, `external`, and `page` that are not available to customers through the current redirects UI

We should update the app to give the user three options to make a redirect and include links to documentation.

API Documentation on this feature is located: https://instances-api.zesty.org/#1fdabbe3-e977-4ea5-8a4b-81ce29598263

Individual APi docs for the different types of redirects
* `path`: https://instances-api.zesty.org/#38bf3fb8-44b1-4b22-a743-376c42fda624
* `path` with Wildcard: https://instances-api.zesty.org/#4e35194e-ce30-43a0-bfe2-91bc76e5a627
* `page` (ZUID): https://instances-api.zesty.org/#9dbf6a75-cb3e-4292-850b-1e78979bd035
* `external`: https://instances-api.zesty.org/#0e1a4940-3d7f-4e59-bd3d-9b06be44d9a3

Zesty.org docs https://zesty.org/services/web-engine/30x-redirects

## Desired Update

The current UI only creates a `path`, the UI should be updated to create the 3 types of redirects.

The current UI lends itself to to work with a couple changes:

**Change 1:** The dropdown search for pages should stay, but on select it should populate with a ZUID
**Change 2:**  The "to" input should not strictly enforce the value to a path, but also a ZUID or full URL with protocol
**Change 3:** Type assumption should be made based in on the incoming string. This would change the post request type value. An assumption on the incoming target string can be made like this:
* any string that include HTTP is of type `external`
* any string with a ZUID (7-xyz-xyz) is of type `page`
* If the above does not match, default to type `path`

Change 4: Titles should change from "From" to "Incoming Path" and "To" to "Redirect Target"

d88naimi:
fixes #643

Screen Shot 2021-08-18 at 3 53 38 PM

### DIFF
--- a/src/apps/seo/src/views/RedirectsManager/RedirectsTable/RedirectsTableRow/RedirectsTableRow.js
+++ b/src/apps/seo/src/views/RedirectsManager/RedirectsTable/RedirectsTableRow/RedirectsTableRow.js
@@ -1,18 +1,38 @@
-import cx from "classnames";
+import { useEffect, useState } from "react";
 import { Link } from "react-router-dom";
+import { useDispatch } from "react-redux";
+import { fetchContentItem } from "shell/store/content";
+import cx from "classnames";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 import {
   faArrowRight,
   faExternalLinkAlt,
   faFile,
   faFileAlt,
+  faLink,
   faTrashAlt,
 } from "@fortawesome/free-solid-svg-icons";
 
 import { Button } from "@zesty-io/core/Button";
+import { Url } from "@zesty-io/core/Url";
 
 import styles from "./RedirectsTableRow.less";
 export default function RedirectsTableRow(props) {
+  const dispatch = useDispatch();
+  const [pathName, setPathName] = useState("");
+  const [pathLink, setPathLink] = useState("");
+
+  useEffect(() => {
+    if (props.targetType === "page") {
+      dispatch(fetchContentItem(props.target)).then((res) => {
+        if (res && res.data) {
+          setPathLink(res.data[0].meta.contentModelZUID);
+          setPathName(res.data[0].web.path);
+        }
+      });
+    }
+  }, [props.target]);
+
   return (
     <div className={styles.RedirectsTableRow}>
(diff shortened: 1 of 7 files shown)


src/shell/store/content.js
```diff
@@ -260,6 +260,39 @@ export function fetchItem(modelZUID, itemZUID) {
   };
 }
 
+export function fetchContentItem(itemZUID) {
```
shrunyan, line 263: Lets use the pre existing `searchItems` endpoint.

src/apps/seo/src/views/RedirectsManager/RedirectsTable/RedirectsTableRow/RedirectsTableRow.js
```diff
@@ -1,21 +1,41 @@
...
 export default function RedirectsTableRow(props) {
+  const dispatch = useDispatch();
+  const [pathName, setPathName] = useState("");
+  const [pathLink, setPathLink] = useState("");
+
+  useEffect(() => {
+    if (props.targetType === "page") {
+      dispatch(fetchContentItem(props.target)).then((res) => {
+        if (res && res.data) {
+          setPathLink(res.data[0].meta.contentModelZUID);
```
shrunyan, line 29: With naming we should prefer to keep the name and value conceptual the same. e.g. `setModelZuid` and `setPath`

shrunyan (changes requested):
I think we should settle on referring to the types as;
- **External** allows for redirecting to a different domain. e.g. https://example.org
- **Internal** allows for redirecting to a page on this instance. e.g. /about
- **Wildcard** allows for redirecting full path parts to pages on this instance. e.g. /blog/*/*

These is hiding the fact that the "wildcard" type does allow for a non-wildcard path but if you are redirect to a path which exists on the instance then you should be using the "internal" type anyways as there has to be a correlated item. This also binds the redirect to the item zuid and so if that item has it's url structure changed it will automatically reflect on the redirect.

PR #887, opened 2021-08-29
updating design systems styles and colors buttons

d88naimi:
The goal of this PR was to get the copyButton to fit well and be consistent with our Zeplin stylesheet, this resulted in a discussion with @shrunyan  on following styling format for <Button> EXAMPLE: <Button kind="outlined" type="tertiary" size="small">

I made overhaul to design systems Button.less & color.less file to follow Zeplin's stylessheet 
https://app.zeplin.io/project/5bb69f5bdfef6551f65f76e8/screen/5bbe83aa23bfd05fb6a9bff7

PR: design-systems: https://github.com/zesty-io/design-system/pull/149/files

Only a couple buttons were slightly changed every other buttons styles remain the same. What had to be changed to be consistent with Zeplin Buttons styles, see screenshots below.

Content home icon 
Before 
Screen Shot 2021-08-29 at 2 49 07 PM
Before &: hover
Screen Shot 2021-08-29 at 2 49 41 PM

After
Screen Shot 2021-08-29 at 2 50 09 PM

After:hover
Screen Shot 2021-08-29 at 2 50 31 PM

SEO & Meta >  Content Insights

Before 
Screen Shot 2021-08-29 at 2 51 48 PM

Before & hover
Screen Shot 2021-08-29 at 2 52 12 PM

After
Screen Shot 2021-08-29 at 2 56 13 PM

After : Hover
Screen Shot 2021-08-29 at 2 56 40 PM

Before Content Disabled
Screen Shot 2021-08-29 at 2 59 55 PM

After Content Disabled(Publish is a `kind="secondary` which has a different style disabled color compared to save)
Screen Shot 2021-08-29 at 3 01 09 PM

*Side notes that need feedback @shrunyan : 
I Removed class `.subdued` never saw the style be used.

I would like to swap kind="save" to kind="primary" 
I would like to swap kind="warn" to kind="destructive"
Remove cancel or swap to tertiary?

### DIFF
--- a/cypress/integration/code/actions_spec.js
+++ b/cypress/integration/code/actions_spec.js
@@ -9,7 +9,7 @@ describe("Code Editor", () => {
     cy.get(".Select").eq(1).click();
     cy.get(".Select li").eq(1).click();
     cy.get('input[type="text"]').type("mySnippet");
-    cy.get('button[kind="save"]').eq(1).click();
+    cy.get('button[type="save"]').eq(1).click();
   });
 
   it("Navigate to file", () => {
@@ -31,7 +31,7 @@ describe("Code Editor", () => {
       .trigger("mousemove", { which: 1, force: true })
       .trigger("mouseup", { force: true });
 
-    cy.get('button[kind="save"]').contains("Save Order").click({ force: true });
+    cy.get('button[type="save"]').contains("Save Order").click({ force: true });
 
     cy.contains("File sort order has been saved", { timeout: 5000 }).should(
       "exist"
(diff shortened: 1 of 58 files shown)


d88naimi:
@shrunyan buttons are now updated globally 

Since the removal of opacity I used a color picker on legacy/production to get a disabled background color. The disabled button may marginally looks different but all colors remain the same.

I also swapped globally in manager-ui everywhere a `kind`= warn, save, cancel to `type`. I have verified that this did not make any breaking changes globally.  

copybutton is now updated to use the newer button definitions described in design-systems pr
<Copybutton kind="outlined" size="compact"/>


Screen Shot 2021-09-07 at 12 31 16 PM
design-system pr: https://github.com/zesty-io/design-system/pull/149

Hover
Screen Shot 2021-09-07 at 12 32 06 PM

shrunyan (approved):

PR #890, opened 2021-09-08
swapped out media copy button for design-system button

d88naimi:
Swapped out media copy button for design-system button, cleaned up border radius and styles for Copy button.

Before
Screen Shot 2021-09-08 at 11 58 32 AM

After
Screen Shot 2021-09-08 at 11 58 44 AM

### DIFF
--- a/src/apps/media/src/app/components/MediaDetailsModal.js
+++ b/src/apps/media/src/app/components/MediaDetailsModal.js
@@ -130,7 +130,7 @@ export const MediaDetailsModal = memo(function MediaDetailsModal(props) {
         </div>
       </ModalContent>
       <ModalFooter className={shared.ModalFooter}>
-        <Button onClick={props.onClose}>
+        <Button type="cancel" onClick={props.onClose}>
           <FontAwesomeIcon icon={faBan} />
           <span>Cancel</span>
         </Button>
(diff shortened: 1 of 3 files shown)


shrunyan:
We should implement this like the other usage of `CopyButton`

d88naimi:
V2 not using new `CopyInput` component 
PR: https://github.com/zesty-io/design-system/pull/152

shrunyan:
> We should implement this like the other usage of `CopyButton`

What I meant by this was we should use the `CopyButton` as it is designed, without any modification. e.g. wrapping it around an input. The value does not need to be in an input as it can not be edited.

d88naimi:
Removed input entirely and using  CopyButton component

default
Screen Shot 2021-09-13 at 11 49 52 AM


Hover
Screen Shot 2021-09-13 at 11 50 04 AM

src/apps/media/src/app/components/MediaDetailsModal.js
```diff
@@ -150,7 +130,7 @@ export const MediaDetailsModal = memo(function MediaDetailsModal(props) {
         </div>
       </ModalContent>
       <ModalFooter className={shared.ModalFooter}>
-        <Button type="cancel" onClick={props.onClose}>
```
shrunyan, line 153: Shouldn't the type be `cancel`? Is the design-system styling incorrect?

shrunyan (changes requested):
- The media ZUID should also use a `CopyButton`.
- Test a long file path which would cause the button to overflow

src/apps/media/src/app/components/MediaDetailsModal.js
d88naimi: fixed and made a pr change

Screen Shot 2021-09-14 at 11 17 08 AM

shrunyan (approved):

PR #896, opened 2021-09-14
making zuid use ds copyButton and fix styling

d88naimi:
made media zuid use ds copyButton

Screen Shot 2021-09-14 at 2 03 06 PM

### DIFF
--- a/src/apps/media/src/app/components/MediaDetailsModal.js
+++ b/src/apps/media/src/app/components/MediaDetailsModal.js
@@ -106,23 +106,22 @@ export const MediaDetailsModal = memo(function MediaDetailsModal(props) {
             onChange={(val) => setFilename(val)}
           />
 
-          <div className={styles.DescriptionList}>
-            <div>
-              <span>ZUID: </span>
+          <dl className={styles.DescriptionList}>
+            <dt>ZUID:</dt>
+            <dd>
               <CopyButton
                 kind="outlined"
                 size="compact"
                 value={props.file.id}
               />
-            </div>
-
+            </dd>
             {props.file.updated_at && (
-              <div>
-                <span>Created at: </span>
-                <em>{props.file.updated_at}</em>
-              </div>
+              <>
+                <dt> Created at: </dt>
+                <dd>{props.file.updated_at}</dd>
+              </>
             )}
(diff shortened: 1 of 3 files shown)


src/apps/media/src/app/components/MediaDetailsModal.js
```diff
@@ -105,28 +105,24 @@ export const MediaDetailsModal = memo(function MediaDetailsModal(props) {
...
-            label={
-              <label>
-                <Infotip title="Edit Alt Attribute " />
-                &nbsp;Alt Attribute
-              </label>
-            }
-            placeholder={"Alt Attribute"}
-          /> */}
 
-          <dl className={styles.DescriptionList}>
```
shrunyan, line 120: Why are we switching away from `dl` with `dt` and `dd`?

d88naimi: Since we read left to right I felt the button nested under ZUID title didn't align nicely.  
 
Screen shot of what it did look like
Screen Shot 2021-09-14 at 2 11 03 PM

shrunyan: Why does the styling necessitate a markup change?

d88naimi: @shrunyan  I couldn't keep the Description List element tag side by side ex: `zuid: button`, do you want me to keep the  Description List element as is like the image above?

shrunyan: I would like us to use semantically meaningful markup when possible. I feel `dl` is an appropriate tag for this scenario. If the visual design of the term and detail being inline is what we want we need to find a way to accomplish this with css or additional markup that lets us achieve the effect without losing semantics.

d88naimi: @shrunyan brought back the description list semantics, making the `dl` inline goes against the the purpose of the semantics so leaving as is. I guess it keeps consistency with the remainder of the medial model.

Screen Shot 2021-09-15 at 1 27 55 PM

shrunyan (approved):

PR #901, opened 2021-10-09
fixed broken locked media image group

### LINKED ISSUE #882: Content models with a media picker field locked to a media group shows a broken modal
A content model with a media picker field locked to a media group will display a broken media model when the user is editing the content. This only happens with items that have a media picker field locked down to a media group. 
The user is still able to select an image with the broken modal but the style needs to be improved. Please see video below for workflow:

https://www.loom.com/share/2f2406074f794aa990ca3b93bfbe95d8

d88naimi:
fixes #882 
fixed broken media image locked group width

Before 
Screen Shot 2021-10-13 at 9 33 45 AM

After
Screen Shot 2021-10-08 at 5 14 43 PM

### DIFF
--- a/src/apps/media/src/app/MediaApp.js
+++ b/src/apps/media/src/app/MediaApp.js
@@ -210,7 +210,7 @@ export default connect((state) => {
           <div
             className={cx(
               styles.WorkspaceContainer,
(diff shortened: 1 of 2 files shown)


shrunyan:
What is broken? The images not loading? Is this group **not** supposed to be locked?

PR #917, opened 2021-11-10
Hot Fix for global account Avatar toggle menu

d88naimi:
Issue:
When selecting the global account avatar you can open the menu but then unable to click to close. 
https://www.loom.com/share/313a7a4089564829a071a9ecd06e6710


d88naimi:
@shrunyan any particular reason why  `evt.stopPropagation();` is commented out? When I uncomment stopPropagation 
https://github.com/zesty-io/manager-ui/blob/master/src/shell/components/global-account/GlobalAccount.js#L44

d88naimi:
@shrunyan Any particular  reason why stopPropagation is commented out ? When I uncomment stopPropagation that seems to fix the issue I reported in this PR.
https://github.com/zesty-io/manager-ui/blob/master/src/shell/components/global-account/GlobalAccount.js#L44

shrunyan:
Seems like an erroneous commit from some development that was happening. https://github.com/zesty-io/manager-ui/commit/21a229f4b1ccc7b8e577c953642f5f8973403b3e#diff-21cb17127e58326f6ab4f56620093a560c82120b709a5054dd0fd7598b4dd0a0R42

cypress/integration/content/navigation_spec.js
```diff
@@ -21,4 +21,9 @@ describe("Navigation through content editor", () => {
     cy.get('[data-value="link"]').click();
     cy.get("#CreateLinkButton").should("exist");
   });
+  it("Open and Close Global Account", () => {
+    cy.get("[data-cy=globalAccountAvatar]").click();
+    cy.get("menu").should("exist");
+    cy.get("[data-cy=globalAccountAvatar]").click();
```
shrunyan, line 27: If it's open and close should there a second assertion of "should not exist"?

src/shell/components/global-account/GlobalAccount.js
```diff
@@ -16,32 +17,30 @@ export default connect((state) => {
   };
 })(function GlobalAccount(props) {
   const [open, setOpen] = useState(false);
-  const ref = useRef();
+  const ref = useRef(null);
+
+  const handleClickOutside = (event) => {
```
shrunyan, line 22: This function definition needs to stay inside the `useEffect.` As the reference will change on re-renders causing the `removeEventListener` to use a different reference, resulting in a memory leak.

shrunyan (changes requested):

shrunyan (approved):

PR #932, opened 2021-12-29
Feature/duomode

d88naimi:


shrunyan:
image
> Screenshot of default state when first viewing an instance after DuoMode is made available.

When DuoMode is first released it will not have had a user selection. The default state for the affected UI needs to be off. i.e. From the screenshot above I would not expect the content Nav or Actions to be collapsed.

shrunyan (approved):

PR #935, opened 2022-01-05
Feature/duomode headless iframe

d88naimi:
In duo mode if the view is a headless content item then show Instant API

Screen Shot 2022-01-05 at 12 28 08 PM

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -91,6 +91,7 @@
         "node-polyfill-webpack-plugin": "^1.1.3",
         "prettier": "^2.3.2",
         "pretty-quick": "^3.1.1",
+        "react-json-view": "^1.21.3",
         "rimraf": "^3.0.2",
         "serve": "^12.0.0",
         "start-server-and-test": "^1.14.0",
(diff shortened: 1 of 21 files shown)


shrunyan (commented):
Why are you swapping the iframe url instead of the route passed to active preview?

d88naimi:
@shrunyan  updated however now I have a hanging loading sign if you have a tip on how to resolve that?

Screen Shot 2022-01-05 at 5 58 12 PM

shrunyan:
You'll need to look at the ActivePreview app to see why that appears that way.

shrunyan:
Also look into adding `react-json-view` in ActivePreview to format the json output. You can see prior art of that usage here: https://github.com/zesty-io/start.zesty.io/blob/main/src/components/GettingStarted/components/Preview/Preview.js#L23

d88naimi:
@shrunyan review por favor

Screen Shot 2022-01-10 at 12 17 54 PM

https://www.loom.com/share/b738b0caa0644fc0937fff0cd942c0e3

shrunyan (commented):
The approach generally looks correct but I think this should occur inside of the ActivePreview app, not within the editor. Meaning the interface, the `route` prop, should remain the same but in the ActivePreview app it determines whether it should be showing the iframe or json preview

d88naimi:
@stuart I am loading this ReactJSON through PreviewMode.js im not following why this needs to be moved in ActivePreview which is the Preview button https://github.com/zesty-io/manager-ui/blob/master/src/shell/components/global-actions/components/ActivePreview/ActivePreview.js#L122-L127

d88naimi:
Screen Shot 2022-01-12 at 1 21 33 PM

PR #937, opened 2022-01-10
fixed scroll on reorder modal

### LINKED ISSUE #934: Content Reordering scroll
**Describe the bug**
Content Reordering modal with lots of items is not scrollable

**To Reproduce**
Visit a instance that has lots of items to reorder Ex: 8-566cdeb-8gqvkj

**Expected behavior**
overflow-y scroll should work if there is a lot of content items to view

**Screenshots**

d88naimi:
Fixes #934 
Scroll bar now works if there is a lot of content in the Reorder nav 

Screen Shot 2022-01-10 at 1 52 47 PM

### DIFF
--- a/src/apps/content-editor/src/app/components/ReorderNav/ReorderNav.less
+++ b/src/apps/content-editor/src/app/components/ReorderNav/ReorderNav.less
@@ -3,7 +3,7 @@
 
 .ReorderNav {
   overflow-y: auto;
-  height: 70vh;
+  max-height: 70vh;
   .ModalContent {
     h3 {
       margin-bottom: 16px;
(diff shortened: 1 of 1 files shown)


src/apps/content-editor/src/app/components/ReorderNav/ReorderNav.less
```diff
@@ -2,6 +2,8 @@
 @import "~@zesty-io/core/typography.less";
 
 .ReorderNav {
+  overflow-y: auto;
+  height: 70vh;
```
shrunyan, line 6: this may be better served as a `max-height` so short lists do not use a large modal but long lists are bounded

shrunyan (approved):

PR #938, opened 2022-01-11
hiding duo mode for item create view

d88naimi:
Before
Screen Shot 2022-01-11 at 3 33 21 PM

After
Screen Shot 2022-01-11 at 3 33 08 PM

### DIFF
--- a/src/apps/content-editor/src/app/views/ItemCreate/ItemCreate.js
+++ b/src/apps/content-editor/src/app/views/ItemCreate/ItemCreate.js
@@ -5,8 +5,6 @@ import { useHistory, useParams } from "react-router-dom";
 import isEmpty from "lodash/isEmpty";
 import { createSelector } from "@reduxjs/toolkit";
 
-import { actions } from "shell/store/ui";
-
 import { WithLoader } from "@zesty-io/core/WithLoader";
 import { Divider } from "@zesty-io/core/Divider";
 import { Header } from "./Header";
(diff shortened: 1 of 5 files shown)


src/apps/content-editor/src/app/views/ItemCreate/ItemCreate.js
```diff
@@ -63,6 +65,10 @@ export default function ItemCreate() {
     }
   }, [modelZUID, item]);
 
+  useEffect(() => {
+    dispatch(actions.setDuoMode(false));
```
shrunyan, line 69: won't this turn it off though? meaning anytime you create a new item you will be booted out of duomode and forced to toggle back on?

d88naimi: @stuart This was the best solution I had in mind beside trying to display none a lot of elements while in Item create view, unless you have a better route?

shrunyan: You will need to lift the DuoMode preview out of `Editor.js` and into https://github.com/zesty-io/manager-ui/blob/f6c5d14fd234848b9bf17ce8ebd13790bb18aad2/src/apps/content-editor/src/app/views/ItemEdit/Content/Content.js

d88naimi: @shrunyan  I have reevaluated the styles and got duo mode to hide correctly for item create with out touching state.

d88naimi:
@shrunyan ready for review thank you.

Screen Shot 2022-01-12 at 1 19 13 PM

d88naimi:
Verified with these new edits it does not cost extra on the network when using display: none

PR #940, opened 2022-01-17
Force close hover state when closing content nav tree and Actions panel

d88naimi:
When closing the content nav or actions panel the hover state is still true keeping the panels open until user moves mouse outside of the containers then the panels will finally close as intended. This fixes this potential confusion for the user. 

This PR: Force closes hover state when closing toggle buttons. 

Fix Demo: https://www.loom.com/share/51e3a6ac02b34144a4cc06a06958ba43


d88naimi:
New PR here with out merge conflicts https://github.com/zesty-io/manager-ui/pull/942

PR #946, opened 2022-01-19
Feature/#482 locked media upload btn

### LINKED ISSUE #482: Media module does not show Upload button if field is locked to a group
**Describe the bug**
If a content field has been previously locked to a media group, its media module does not show the upload button. 

**To Reproduce**
1. Go to an instance with a content field that is locked to a group. See internal ticket for more details: https://zestysupport.zendesk.com/agent/tickets/1167
2. Click on media field in the content item to open the media module
3. See the module does not display "Upload" button

**Expected behavior**
Expect to see the upload button just like when an unlocked media module is displaying. 

**Screenshots**
IMG_0121

 - OS:MacOS Big Sur
 - Browser: Safari & Chrome

**Additional context**
This was reported by a client who was confused as to how to upload media. The drag & drop option is still available but it is not intuitive. An upload button would have helped relieve the confusion as well as some instruction about the drag & drop function.

d88naimi:
Fixes #482 

Added Upload button to locked image group

Screen Shot 2022-01-19 at 11 01 26 AM


d88naimi:
V2
Screen Shot 2022-01-24 at 3 26 54 PM

shrunyan:
@d88naimi please see prior comment
> In application code we should avoid blindly passing through props to components. Please switch this to explicit props

d88naimi:
@stuart V3 updated I think I understood what you men't by explicit props now

src/apps/media/src/app/MediaApp.js
```diff
@@ -221,6 +221,7 @@ export default connect((state) => {
               showDeleteGroupModal={showDeleteGroupModal}
               modal={props.modal}
               setCurrentGroupID={setCurrentGroupID}
+              groupID={props.groupID}
```
agalin920, line 224: You are passing a new Prop chain for groupID but I believe currentGroup already has the ID you need

line 55 of this file states:
```
const [currentGroupID, setCurrentGroupID] = useState(
    params.groupID || props.groupID
  );
```
along with a useEffect that updates the state whenever props.groupId changes

With this info you should be able remove the whole new prop chain you are adding since it looks like its already in place

src/apps/media/src/app/components/MediaHeader.js
```diff
@@ -10,21 +10,34 @@ import { MediaCreateGroupModal } from "./MediaCreateGroupModal";
...
       <h1 className={cx(styles.subheadline, styles.Title)}>
         <small className={styles.NumFiles}>({props.numFiles})</small>
         {props.searchTerm
           ? `Search Results "${props.searchTerm}"`
           : props.currentGroup.name}
       </h1>
 
+      {props.groupID && (
+        <UploadImage
+          currentGroup={props.currentGroup}
```
agalin920, line 36: I think naming the prop for this component currentX is misleading. Since its a higher level abstraction that can upload anything without it needing context of whether or not it is 'current' I think a much better name for it is simply 'group' and 'bin'.

PR #947, opened 2022-01-19
Sentry passing props dispatch to ItemParent

d88naimi:
Sentry issue https://sentry.io/organizations/zestyio/issues/2328692196/?project=5441698

### DIFF
--- a/src/apps/content-editor/src/app/views/ItemEdit/Meta/ItemSettings/ItemSettings.js
+++ b/src/apps/content-editor/src/app/views/ItemEdit/Meta/ItemSettings/ItemSettings.js
@@ -53,6 +53,7 @@ export const ItemSettings = memo(
                 path={web.path}
                 onChange={onChange}
                 currentItemLangID={props.item.meta.langID}
+                dispatch={props.dispatch}
               />
               <ItemRoute
                 ZUID={meta.ZUID}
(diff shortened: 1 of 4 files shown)


src/apps/content-editor/src/app/views/ItemEdit/Meta/ItemSettings/ItemSettings.js
```diff
@@ -53,6 +53,7 @@ export const ItemSettings = memo(
                 path={web.path}
                 onChange={onChange}
                 currentItemLangID={props.item.meta.langID}
+                dispatch={props.dispatch}
```
shrunyan, line 56: This seems unrelated to the error message on the sentry bug referenced; 

> Cannot read properties of undefined (reading 'data')

That appears to be that the response is empty but we are still attempting to read the `data` property

shrunyan (changes requested):

src/apps/content-editor/src/app/views/ItemEdit/Meta/ItemSettings/ItemSettings.js
d88naimi: Added Optional Chaining to `data`, but leaving the props.dispatch as ItemParent is is missing dispatch.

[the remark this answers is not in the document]
d88naimi: Removed, but im not sure how props.dispatch is being used here https://github.com/zesty-io/manager-ui/blob/master/src/apps/content-editor/src/app/views/ItemEdit/Meta/ItemSettings/settings/ItemParent.js#L95

src/shell/store/middleware/api.js
```diff
@@ -13,7 +13,9 @@ export const fetchResource =
       if (inflight.indexOf(action.uri) === -1) {
         inflight.push(action.uri);
         return request(action.uri)
-          .then(action.handler)
+          .then((res) => {
```
shrunyan, line 16: Let's switch back to the direct function reference

shrunyan (changes requested):

PR #954, opened 2022-01-24
updating title attr and info tips in media modal

### LINKED ISSUE #949: Improve media url editing
Currently it is unclear that the '`filename` property of a media asset controls the url.

We should combine the url and filename editing experience into a single field so it clearly communicates changing the filename changes the url. This will need to maintain a button to copy the url.

**Additional context**

d88naimi:
fixes #949 
PR: design systems : https://github.com/zesty-io/design-system/pull/159

Screen Shot 2022-01-24 at 11 56 18 AM

Screen Shot 2022-01-24 at 11 56 28 AM

I think the idea of combining  filename and title would cause confusion. I went with more descriptive titles and added better information in the info tip.

### DIFF
--- a/src/apps/media/src/app/components/MediaDetailsModal.js
+++ b/src/apps/media/src/app/components/MediaDetailsModal.js
@@ -78,32 +78,31 @@ export const MediaDetailsModal = memo(function MediaDetailsModal(props) {
             kind="outlined"
             value={props.file.url}
           />
-
-          <FieldTypeText
-            className={styles.Field}
-            name="title"
-            value={title}
-            label={
-              <label>
-                <Infotip title="Edit title. | Use for alt text with Parsley's .getImageTitle()" />
-                &nbsp;Title
-              </label>
-            }
-            placeholder={"Image Title"}
-            onChange={(val) => setTitle(val)}
-          />
           <FieldTypeText
             className={styles.Field}
             name="filename"
             value={filename}
             label={
               <label>
-                <Infotip title="Edit Filename " />
-                &nbsp;Filename
+                <Infotip title="Edit Filename to update your image URL " />
+                &nbsp;Edit Filename URL
               </label>
             }
             placeholder={"Image Filename"}
             onChange={(val) => setFilename(val)}
+          />
+          <FieldTypeText
+            className={styles.Field}
+            name="title"
+            value={title}
+            label={
+              <label>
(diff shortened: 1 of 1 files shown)


agalin920:
@shrunyan Thought of the edge case where a user can type in a slash. After testing I saw that the system replaces slashes with a dash. Should we do the same transformation when displaying end url on copy button as user types?

shrunyan:
@agalin920 yes. The experience should reflect the end outcome of updating the filename. i.e. The url when copied should load the image after the change is saved.

src/apps/media/src/app/components/MediaDetailsModal.js
```diff
@@ -73,37 +75,42 @@ export const MediaDetailsModal = memo(function MediaDetailsModal(props) {
...
+            value={filename}
             label={
               <label>
-                <Infotip title="Edit title. | Use for alt text with Parsley's .getImageTitle()" />
-                &nbsp;Title
+                <Infotip
+                  className={styles.InfotipFileName}
+                  title="Edit Filename to update your image URL path "
+                />
+                &nbsp;Edit Filename URL path
```
shrunyan, line 88: "Edit Filename URL path" feels to overdone. I would suggest we leave it as just "URL Filename". This accomplishes the following;
- removes the verb of the action being done "Edit". Not present on other fields.
- removes the description "path" as we do not support multi slug path parts which this seems to indicate.

shrunyan (changes requested):

src/apps/media/src/app/components/MediaDetailsModal.js
```diff
@@ -89,7 +88,8 @@ export const MediaDetailsModal = memo(function MediaDetailsModal(props) {
               </label>
             }
             placeholder={"Image Filename"}
-            onChange={(val) => setFilename(val)}
+            // Replaces all non-alphanumeric characters (excluding '.') with '-' to reflect the filename transformation done on the BE
+            onChange={(val) => setFilename(val.replaceAll(/[^a-z\d-.]/gi, "-"))}
```
shrunyan, line 92: there is a format util function, we use it when adding field labels, which ensures computer friendly strings. we need to be using that here.

shrunyan (changes requested):

shrunyan (approved):

PR #967, opened 2022-02-08
Feature/965 codelock

### LINKED ISSUE #965: CodeLock
When someone wants to work on a code file in the code-app there is no way for them to know if another user is editing the same file. Creating a situation where user A can overwrite user B changes unknowingly. 

We should introduce a "CodeLock". Replicating the exisiting "ContentLock" in the content-editor which provides this same solution for item editing.

What this means is when a code file is loaded we need to check the cache service and see whether someone else has locked the file(see: LockedItem). If the item is locked we need to render a `LockedFile` component. If the item is either unlocked or from the `LockedFile` component a user opts to "unlock" the file we would need to then lock the file to the current user in question.

**Additional context**
https://github.com/zesty-io/manager-ui/pull/962 Is a body of work which touches on the ContentLock. Particularly [`LockedItem`](https://github.com/zesty-io/manager-ui/blob/b3f0cecfdf647d6ea62e84369d2e0b42b56276c3/src/apps/content-editor/src/app/components/LockedItem/LockedItem.js) is the component to replicate into the code-app.

Testing a lock requires using 2 accounts with 2 separate sessions. By having both accounts view the file in question a lock modal should be rendered.

d88naimi:
Added CodeLock to code view to prevent two users from overriding each other work.

Door Lock  drawio (1)

fixes #965


d88naimi:
@shrunyan updated video , fixed modal flicker https://www.screencast.com/t/uqev92nDrtup
Screen Shot 2022-02-11 at 10 52 25 AM

d88naimi:
Working functionality video guide : https://www.screencast.com/t/PmXmuEGH9


Screenshot two separate  user on the same code file 

Screen Shot 2022-02-16 at 1 32 18 PM

shrunyan (approved):

PR #970, opened 2022-02-09
resolve auto populate

### LINKED ISSUE #703: Schema: field reference names do not autopopulate

### LINKED ISSUE #969: New fields do not autopopulate code reference as field name is filled out
When a user creates a new field, the code reference should auto populate as they fill out the field's label. Currently, it allows the user to create a field with a blank looking code reference which will then autogenerate that code reference using its field label. You can see that flow here:
https://www.loom.com/share/f831758d759444428ca3b4b8ad70aa4b

The solution would be that when a user creates a new field and starts typing out the field's label, that the code reference field is auto populated using the field label's value.

d88naimi:
Fixing cursor jump / auto-populate in schema.

fixes #703 
fixes #969 
https://www.screencast.com/t/j4vEvYnsz

CC: @shrunyan  @giseleblair , @fuskovic, @agalin920

### DIFF
--- a/src/apps/schema/src/app/views/SchemaEdit/FieldSettings/FieldSettings.js
+++ b/src/apps/schema/src/app/views/SchemaEdit/FieldSettings/FieldSettings.js
@@ -1,4 +1,4 @@
-import { Fragment } from "react";
+import { Fragment, useRef, useEffect } from "react";
 import cx from "classnames";
 
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
@@ -17,8 +17,6 @@ import {
   faMoneyBillAlt,
   faIdCard,
   faExclamationTriangle,
-  faImage,
-  faCalendarTimes,
   faCalendarAlt,
 } from "@fortawesome/free-solid-svg-icons";
 import { FieldTypeText } from "@zesty-io/core/FieldTypeText";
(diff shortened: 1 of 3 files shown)


src/apps/schema/src/app/views/SchemaEdit/FieldSettings/FieldSettings.js
```diff
@@ -77,17 +77,21 @@ export default function FieldSettings(props) {
...
+                props.updateValue(formatName(val), key);
               }
             }}
           />
           <FieldTypeText
             className={styles.Setting}
             name="name"
             label="Field Name (Parsley Code Reference). Can not contain spaces, uppercase or special characters."
-            defaultValue={props.field.name}
+            {...(props.new
```
shrunyan, line 88: Where does the `new` property come from?

agalin920: @shrunyan FieldAdd component has FieldSettings child with new prop. FieldEdit component has FieldSettings child without it.

shrunyan: Okay. Do we view there is a fundamental issue with how this component is structured and this is a quick workaround or is there something else wrong here and why we need this awkward decision between value or defaultValue?

agalin920: @shrunyan Yea so the cursor jump is a pretty common case within controlled inputs [Check this out](https://stackoverflow.com/questions/28922275/in-reactjs-why-does-setstate-behave-differently-when-called-synchronously/28922465#28922465). The only way around it is either to use defaultValue to avoid initial cursor jump(but this limits the inputs controllability) OR do some much more complex solution regarding cursor management or some async input handler.

In this case we only need to use value as opposed to defaultValue when creating a new field because that is where we are auto-populating the other field and we can stick to using defaultValue on edit and avoid the cursor jump. Hence the conditional use of defaultValue or value depending on whether its a new or an edit. 

IMO we shouldn't worry about this cursor jump because at this point its pretty much expected behavior. So we could just leave it as value for both new and edit.

shrunyan:
@agalin920  
What happens when editing an existing field someone inputs an invalid value into the `name` field? e.g. "my space field"
What happens when creating a new field someone inputs a value into the field and then click back into the middle of that string, to correct something?

shrunyan (approved):

PR #972, opened 2022-02-16
Content nav keyframes

d88naimi:


ardeay:
My opinion: The nature of the layering of the nav behind, then infront, then sliding out, creates the issue. If the navigation collapsed with the content editing interface and then pushed it out, it could resolve the problem. 
HEre is an example to visual what im saying

This show an open state of the nav
image

This shows a closes state of the nav
image

Video example: 
https://www.loom.com/share/85ea326c346d4ad4a7b149e06eb1ae32

PR #973, opened 2022-02-16
Notice page for user's that deleted all models and items

### LINKED ISSUE #961: Edge case: User deletes all models/items preventing the Content Manager to load
**Describe the bug**
Edge case: User deletes all models/items, including the Homepage and Globals/Clippings, causing the `/nav` to return an empty array.   
With the `/nav` empty the manager will not load. Returns a `404 Not found` in the manager and hangs on the Starting Content Editor screen. 

**To Reproduce**
1. Open instance or access through API 
2. Delete all content models and items 
3. Enter the manager to view the loading screen in the Content App
4. An error is displayed to the user indicating `404 Not found`

**Expected behavior**
Updated workflow
- either not allowing the user to delete the Homepage and/or Globals/Clippings, preventing the `/nav` to return an empty array
- Or edge case handler if deleted, to avoid loading screen issue

**Screenshots**
image

d88naimi:
fixes #961 

When a user deletes all models and items the content shows a hanging loader.
This PR shows a notice letting the user know to create a new schema and provides a link.
Using DS display class 

Screen Shot 2022-03-01 at 1 55 10 PM

Side Notes* 
if the user deletes their image/bin as well there is a broken/hanging loading media bin page. I tried creating a new model with an image field but have no way of accessing the upload image button.
 
Screen Shot 2022-02-16 at 3 36 00 PM
CC: @MukeshGKastala @shrunyan

### DIFF
--- a/src/apps/content-editor/src/app/ContentEditor.js
+++ b/src/apps/content-editor/src/app/ContentEditor.js
@@ -1,11 +1,14 @@
-import { useEffect } from "react";
+import { useEffect, Fragment } from "react";
 import { useDispatch, useSelector } from "react-redux";
 import { Switch, Route } from "react-router-dom";
 import cx from "classnames";
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+import { faDatabase } from "@fortawesome/free-solid-svg-icons";
 
 import { fetchModels } from "shell/store/models";
 import { fetchNav } from "../store/navContent";
 
+import { AppLink } from "@zesty-io/core/AppLink";
 import { WithLoader } from "@zesty-io/core/WithLoader";
 import { ContentNav } from "./components/Nav";
 import { ContentNavToggle } from "./components/Nav/components/ContentNavToggle";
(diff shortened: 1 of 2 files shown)


src/apps/content-editor/src/app/ContentEditor.js
```diff
@@ -37,57 +40,80 @@ export default function ContentEditor() {
...
-              <Route
-                path="/content/:modelZUID/:itemZUID"
-                component={ItemEdit}
-              />
-              <Route exact path="/content/:modelZUID" component={ItemList} />
-              <Route path="*" component={NotFound} />
-            </Switch>
-          </div>
+    <Fragment>
+      {navContent.headless.length === 0 ? (
```
shrunyan, line 44: We need to look `raw` data, which is the non segmented data from the api, to determine whether there are no models on the instance. `headless` only accounts for datasets and so an instance which contain single or multi page but no headless models would fail to render.

shrunyan (changes requested):

d88naimi:
Notes* Dummy Instance for broken testing http://8-9ef0f2a593-47wszf.manager.dev.zesty.io:8080/content

shrunyan (approved):

PR #974, opened 2022-02-18
prevent crash if term.meta is missing

d88naimi:
Sentry Bug
onSelect(src/apps/seo/src/views/RedirectsManager/RedirectsTable/RedirectCreator/RedirectCreator)

TypeError
Cannot read properties of undefined (reading 'meta')

https://sentry.io/organizations/zestyio/issues/3001042130/?project=5441698&query=is%3Aunresolved&statsPeriod=14d


shrunyan (approved):

dominictarantino:
Hi @d88naimi would you be so kind as to give me a little more context into this fix? What is the behavior of the user, if any, when this happens?

PR #981, opened 2022-03-01
resolved version control feature in code editor

### LINKED ISSUE #520: WebIDE: doesn't load previous versions of a code file
**Describe the bug**
When using the diff module, clicking on the version number button does not load the selected version but takes you back to the Content App.

**To Reproduce**
1. Go to WebIDE and click on a code file
2. Click on the clock icon to view the diff versions module
3. Click on a previous version (on the right hand dropdown)
4. Click on the green button with the version number
5. See that you are redirected to Content

**Expected behavior**
I would expect to have the version selected load in the WebIDE

**Screenshots**
https://www.loom.com/share/e205101e66a54b518eed649cf54e2129

 - OS: Mac Big Sur
 - Browser: Safari
 - Version:  14.0.3

d88naimi:
Fixed version control in code editor 
Fixes #520 

https://www.screencast.com/t/czh3pGNU

CC: @giseleblair @shrunyan, @agalin920,

### DIFF
--- a/src/apps/code-editor/src/app/components/Editor/components/MemoizedEditor/MemoizedEditor.js
+++ b/src/apps/code-editor/src/app/components/Editor/components/MemoizedEditor/MemoizedEditor.js
@@ -29,6 +29,13 @@ export const MemoizedEditor = memo(
       ref.current.editor.setModel(model);
     }, [props.fileName]);
 
+    useEffect(() => {
+      if (!ref.current?.editor) return;
+      if (props.code !== ref.current.editor.getValue()) {
+        ref.current.editor.getModel().setValue(props.code);
+      }
+    }, [props.code]);
+
     return (
       <MonacoEditor
         ref={ref}


src/apps/code-editor/src/app/components/Editor/components/MemoizedEditor/MemoizedEditor.js
```diff
@@ -29,6 +29,13 @@ export const MemoizedEditor = memo(
       ref.current.editor.setModel(model);
     }, [props.fileName]);
 
+    useEffect(() => {
+      if (!ref.current?.editor) return;
+      if (props.code !== ref.current.editor.getValue()) {
+        ref.current.editor.getModel().setValue(props.code);
+      }
+    }, [props.code]);
```
shrunyan, line 37: As "code" changes are made in the Monaco editor those are broadcast to redux. Which would trigger a render as the value is changing. Eventually flowing to this point in the code. Would this not `setValue` on every key stroke? What performance implications does this have?

agalin920: @shrunyan Yea it would not set the Editor value on every keystroke. The check on line 34 ensures that the content of the editor only gets overwritten by new content in the redux store if and only if it is different to the current editor content. Therefore any other part of the app can modify the content value in the redux store and have it reflected onto the editor (without us making the Editor a fully controlled component)

shrunyan (approved):

giseleblair:
I see that this has been merged but I just tested this and issue #520 still persists in production. @d88naimi can you walk me through what this pr resolves?

d88naimi:
@giseleblair this is working for me https://www.screencast.com/t/czh3pGNU, however if you are looking at a JS file then yes it's still a bug and that's reported here https://github.com/zesty-io/manager-ui/issues/982

giseleblair:
@d88naimi got it, I was using the dropdown on the left. Some additional indication as to which dropdown will load the file might be useful

PR #989, opened 2022-03-04
Fix Schema Editor not handling empty states

agalin920:
Refactor schema structure to match content editor and add loading states to properly handle empty states

https://www.loom.com/share/4ccbdefb489a4e03b3fe6c256e76296d

### DIFF
--- a/src/apps/schema/src/app/main.js
+++ b/src/apps/schema/src/app/main.js
@@ -22,11 +22,13 @@ export default function SchemaBuilder() {
 
   useEffect(() => {
     setLoading(true);
-    Promise.all([dispatch(fetchModels()), dispatch(fetchSettings())]).then(
-      () => {
+    Promise.all([dispatch(fetchModels()), dispatch(fetchSettings())])
+      .then(() => {
         setLoading(false);
-      }
-    );
+      })
+      .catch((e) => {
+        throw e;
+      });
   }, []);
 
   return (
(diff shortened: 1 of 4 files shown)


src/apps/schema/src/app/main.js
```diff
@@ -1,30 +1,44 @@
...
   const match = useRouteMatch("/schema/start");
   const showNav = match && match.isExact;
+  const navSchema = useSelector((state) => state.navSchema);
+  const dispatch = useDispatch();
+
+  const [loading, setLoading] = useState(true);
+
+  useEffect(() => {
+    setLoading(true);
+    Promise.all([dispatch(fetchModels()), dispatch(fetchSettings())]).then(
```
shrunyan, line 25: Should we include a `catch` clause to handle if an API request fails?

agalin920: Sure. What is our standard to handle API failures? Can you point me to an existing case to base myself off on? Should we also add some to the content-editor .then chain?

shrunyan: This is probably a good example: https://github.com/zesty-io/manager-ui/blob/master/src/apps/audit-trail/src/views/app/index.js#L26

In an api request failure we'd want to stop the loading state and show an error notification

agalin920: Sounds good. On it

shrunyan (approved):

PR #1001, opened 2022-03-17
Fix hardcoded url to match appropriate file type

agalin920:
Bug was caused by hardcoded 'view' path pushed after loading a version causing the following endpoint

https://8-f48cf3a682-7fthvk.api.zesty.io/v1/web/scripts/zuid/versions/ to be called as https://8-f48cf3a682-7fthvk.api.zesty.io/v1/web/views/10-5c7db4-cv90bt/versions/

returning an empty version array

### DIFF
--- a/src/apps/code-editor/src/app/components/FileActions/components/DifferActions/DifferActions.js
+++ b/src/apps/code-editor/src/app/components/FileActions/components/DifferActions/DifferActions.js
@@ -45,7 +45,7 @@ export const DifferActions = memo(function DifferActions(props) {
       );
     }
 
-    history.push(`/code/file/views/${props.fileZUID}`);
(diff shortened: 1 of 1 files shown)


src/apps/code-editor/src/app/components/FileActions/components/DifferActions/DifferActions.js
```diff
@@ -73,12 +73,13 @@ export const DifferActions = memo(function DifferActions(props) {
   }
 
   useEffect(() => {
+    console.log("im in", props.fileType);
```
shrunyan, line 76: please remove console logs

shrunyan (changes requested):

shrunyan (approved):

PR #1003, opened 2022-03-22
Update file upload error to include file size limit

agalin920:


src/shell/store/media.js
```diff
@@ -545,7 +545,13 @@ export function uploadFile(file, bin) {
   return (dispatch, getState) => {
     function handleError() {
       dispatch(fileUploadError(file));
-      dispatch(notify({ message: "Failed uploading file", kind: "error" }));
+      dispatch(
+        notify({
+          message:
+            "Failed uploading file. Individual file size must be 500MB or smaller",
```
shrunyan, line 551: it is currently limited to 32mb

shrunyan (changes requested):

PR #1005, opened 2022-03-22
fix workflows

fuskovic:
# What this does

Fixes this error for both CI and CD:

```
Error: On 2022-04-05, the default branch will be renamed from "master" to "main". Your action is currently pinned to "@master". Even though GitHub creates redirects for renamed branches, testing found that this rename breaks existing GitHub Actions workflows that are pinned to the old branch name.
```


fuskovic:
Closing because local branch was in bad state

PR #1021, opened 2022-04-12
Refactor Analytics class to function component and use useDomain hook

agalin920:
NOTE: Requires in app testing.

When I attempted the auth flow I got a "No Account for this domain error"


shrunyan:
closing in favor of https://github.com/zesty-io/app-google-analytics

PR #1029, opened 2022-04-21
added relevant data set icon for schema tabs

### LINKED ISSUE #1018: Icon Update for Model Schema tabs in the navigation
**User Request:** 
The icon presented in the navigation tab, which indicates a tab is a Model schema, is the same as the Headless dataset icon. When viewed as an open tab it is confusing to users which is the schema tab and which is the content item table tab 

**User** would like a different icon indicating a tab is a Model schema file.

**Additional context**

**Model schema Icon:**
image

**Headless Dataset Icon:**
image

d88naimi:
fixes #1018 

Adding icons to match schema data set 

Screen Shot 2022-04-21 at 10 58 08 AM

### DIFF
--- a/src/shell/components/global-tabs/GlobalTabs.js
+++ b/src/shell/components/global-tabs/GlobalTabs.js
@@ -7,19 +7,7 @@ import debounce from "lodash/debounce";
 
 import { AppLink } from "@zesty-io/core/AppLink";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
-import {
-  faTimesCircle,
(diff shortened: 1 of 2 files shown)


agalin920:
@d88naimi I don't think this is the best solution for this. Essentially you are OVERRIDING our current tab icon calculation and only on the schema view. So this begs two questions

1. Why only schema view
2. Shouldn't our existing icon calculation logic be updated instead of overridden?

The existing logic can be found in

/Users/andresgalindo/Repos/manager-ui/src/shell/store/ui.js

PR #1033, opened 2022-04-29
Support directories in stylesheet and script code nav sections

agalin920:
Screen Shot 2022-04-28 at 7 52 26 PM

### DIFF
--- a/src/apps/code-editor/src/app/components/FileList/FileList.js
+++ b/src/apps/code-editor/src/app/components/FileList/FileList.js
@@ -22,25 +22,22 @@ export const FileList = memo(function FileList(props) {
     props.navCode.tree.sort(byLabel)
   );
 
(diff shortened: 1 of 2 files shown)


shrunyan:
@agalin920 

The directory css/js files are that way because they are custom file types. These files do not get included in the site.css and site.js bundles which are automatically built by concatenating the relative file types.

My concern is by showing the directories below those bundle titles it will create an incorrect expectation that they will be present in those bundles. Will this maybe the case, it is also confusing seeing those same resource types listed elsewhere. Begging the question, "why are these not included with the site.js & site.css bundles"

agalin920:
@shrunyan Confirmed that custom css still gets filed under Views and not in site.css bundle
Screen Shot 2022-05-02 at 12 23 17 PM

shrunyan (approved):

PR #1047, opened 2022-05-10
Convert embedded infotips to MUI tooltips

rjpals:
This converts infotips that are embedded within other components to MUI tooltips.


rjpals:
Not necessary -- this will be done as part of the conversion for text fields.

PR #1048, opened 2022-05-10
Touchups code mui buttons review request

d88naimi:
made all CircularProgress from 1rem to 20px

fixed extra large icons in code header diff version 

Before
Screen Shot 2022-05-10 at 12 15 47 PM

Fixed 
Screen Shot 2022-05-10 at 12 00 35 PM

swapped out code header Zuid to DS copyButton 
Screen Shot 2022-05-10 at 12 23 07 PM

### DIFF
--- a/src/apps/code-editor/src/app/components/FileActions/FileActions.less
+++ b/src/apps/code-editor/src/app/components/FileActions/FileActions.less
@@ -25,6 +25,9 @@
 
       display: flex;
       align-items: center;
+      button {
+        margin-right: 8px;
+      }
 
       span {
         word-break: break-all;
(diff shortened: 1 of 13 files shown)


src/apps/code-editor/src/app/components/FileActions/FileActions.less
```diff
@@ -25,6 +25,9 @@
 
       display: flex;
       align-items: center;
+      button {
```
agalin920, line 28: @d88naimi I see you did this to apply a margin on the design-system bound CopyButton. I would like to see this case being solved in an alternative fashion to ease its eventual transition to 'material'. Adding this here makes it untransparent and bug-prone when we swap it out to 'material' because this style can easily bleed into it.

One alt solution that comes to mind is to wrap the CopyButton in the wrapper and applying the margin on there. This will give us more transparency of what is being applied and why as well as avoiding the style bleed we have encountered with styles selecting native elements.

agalin920 (approved):
Approved

PR #1050, opened 2022-05-11
Media mui buttons migration

d88naimi:


src/apps/media/src/app/components/MediaSelected.js
```diff
@@ -88,9 +89,22 @@ export function MediaSelected(props) {
...
                           />
                         </figure>
-                        <button className={styles.Check} aria-label="Checked">
-                          <FontAwesomeIcon icon={faCheck} />
-                        </button>
+                        <Button
+                          variant="contained"
+                          size="small"
+                          aria-label="Checked"
+                          sx={{
```
shrunyan, line 96: Is there a reason we are using inline styles for these?

src/apps/media/src/app/components/MediaWorkspaceItem.js
```diff
@@ -65,9 +70,27 @@ export const MediaWorkspaceItem = memo(function MediaWorkspaceItem(props) {
...
           {props.modal ? (
-            <button className={styles.Check} aria-label="Checked">
-              <FontAwesomeIcon icon={faCheck} />
-            </button>
+            <Button
+              variant="contained"
+              className={styles.Check}
+              aria-label="Checked"
+              size="small"
+              sx={{
```
shrunyan, line 78: Same. What's the reason for using inline styles?

src/apps/media/src/app/components/MediaSelected.js
d88naimi: @shrunyan this is the MUI convention and the method we used Marketing website. In order to clean up classname styling I used this method. If we want to move away from this method we can also write it with styled component. 

Side notes * The sx isn't inline styling it does not show the styles in the dom.

src/apps/media/src/app/components/MediaWorkspaceItem.js
d88naimi: @shrunyan this is the MUI convention and the method we used Marketing website. In order to clean up classname styling I used this method. If we want to move away from this method we can also write it with styled component. 

Side notes * The sx isn't inline styling it does not show the styles in the dom.

shrunyan: Sorry not inline. Internal. They are added as `style` tags at the `head`. We should avoid this approach in general. It  can become problematic with how many components can be rendered on screen.

/cc @agalin920

d88naimi: @shrunyan I think I understand your question now, with the use of sx styling combined with `styles.Check` this is an extreme case with the on hover of the parent Card then will show child level check box icon. This was the best solution I came up with out having to refactor the whole layout. I am open to suggestions.
 
https://www.screencast.com/t/anER7fU2Dp1

shrunyan: Umm I think what you are describing is different than my concern. Spoke with @agalin920 and talked through my perf concerns. They seem to be non-issues with how emotion handles `style` tag addition to the DOM. We are good to use sx or styled components. There is an exit strategy in mind if we do hit unknown perf issues. 

@agalin920 will discuss sx vs styled components with you and @rjpals tomorrow. Gist is we should pick one and as a rule of thumb use that technique.

Marking as resolved

PR #1056, opened 2022-05-12
Fix tooltip icon alignment

### LINKED ISSUE #1055: Tooltip icons are misaligned
**Describe the bug**
MUI tooltip icons are misaligned with text

See this infotip from the Content editor, meta section.
Screen Shot 2022-05-12 at 11 20 40 AM

rjpals:
# Aligns tooltip icons with label text

Closes #1055 

### Changing content editors meta info page to have more consistent styling:

Old:
Screen Shot 2022-05-12 at 11 27 47 AM

New:
Screen Shot 2022-05-12 at 11 27 17 AM

### Changing SEO page styling so table headers don't wrap on small screens

Old:
Screen Shot 2022-05-12 at 11 31 55 AM

New:
Screen Shot 2022-05-12 at 11 32 50 AM

 This layout is not perfect, (you can see the titles are slightly misaligned) but because the table was implemented with `<span>`s and 3 separate css `<grid>`s, it is the best we can get without a complete refactor.

### DIFF
--- a/src/apps/code-editor/src/app/components/FileList/FileList.less
+++ b/src/apps/code-editor/src/app/components/FileList/FileList.less
@@ -53,6 +53,8 @@
     h1 {
       color: @zesty-light-blue;
       text-transform: uppercase;
+      display: flex;
+      align-items: center;
       margin-right: auto;
     }
   }
--- a/src/apps/content-editor/src/app/views/ItemEdit/Content/Actions/Widgets/WidgetListed/WidgetListed.less
+++ b/src/apps/content-editor/src/app/views/ItemEdit/Content/Actions/Widgets/WidgetListed/WidgetListed.less
@@ -2,4 +2,8 @@
   article {
     margin-bottom: 16px;
   }
+  span {
+    display: flex;
+    align-items: center;
+  }
 }
--- a/src/apps/content-editor/src/app/views/ItemEdit/Meta/ItemSettings/settings/CanonicalTag.less
+++ b/src/apps/content-editor/src/app/views/ItemEdit/Meta/ItemSettings/settings/CanonicalTag.less
@@ -1,6 +1,7 @@
 .CanonicalTag {
   label {
     display: flex;
+    align-items: center;
   }
   .settings {
     & > * {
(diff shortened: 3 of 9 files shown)


src/apps/code-editor/src/app/components/FileList/FileList.less
```diff
@@ -53,6 +53,8 @@
     h1 {
       color: @zesty-light-blue;
       text-transform: uppercase;
+      display: flex;
```
agalin920, line 56: This is very smelly to me. I would prefer to see a flex container wrap the h1 and tooltip separately. As opposed to converting h1 into a flex container with tooltip and text child. MUI provides quick and easy way to declare a flex container

src/apps/content-editor/src/app/views/ItemEdit/Content/Actions/Widgets/WidgetListed/WidgetListed.less
```diff
@@ -2,4 +2,8 @@
   article {
     margin-bottom: 16px;
   }
+  span {
```
agalin920, line 5: Same comment here as avove

src/apps/content-editor/src/app/views/ItemEdit/Meta/ItemSettings/settings/CanonicalTag.less
```diff
@@ -1,6 +1,7 @@
 .CanonicalTag {
   label {
     display: flex;
```
agalin920, line 3: Same comment here but seems like it has been previously set that way here :/

src/apps/media/src/app/components/MediaDetailsModal.less
```diff
@@ -32,6 +32,7 @@
     .FieldsContainer {
       label {
         margin-bottom: 16px;
+        display: flex;
```
agalin920, line 35: Same comment regarding flex on traditionally inline elements

agalin920 (approved):
Approved but I would prefer to see MUI declarative flex containers in the future to avoid setting them through stylesheets

PR #1075, opened 2022-05-16
Fix failing Create Model tests

rjpals:


rjpals:
Should be resolved by #1050

PR #1090, opened 2022-05-19
Fetch both dev and live files using query params

agalin920:


shrunyan:
This combines `live` and `dev` files into the same state slice. Are we confident that the UI is accounting for that and showing them in isolation? We must not have live files loading as editable.

shrunyan (approved):

PR #1091, opened 2022-05-19
Create warning popup when changing the reference name of a schema field

### LINKED ISSUE #1052: On Schema Field Change, Create a Warning popup!
When schema fields are about to change, a warning should pop up letting the user know this change could **break their experience**, and make them accept the fact (OK) or give the the option to cancel (CANCEL) and revert the change. On clicking OK the request would fire, on clicking cancel the change would revert

image

rjpals:
Create warning popup when changing the reference name of a field, so that users are reminded that the change can break production code.

Closes #1052 

Video of proposed UX:

https://user-images.githubusercontent.com/16678143/169360240-f2005833-57d9-48fc-b8fd-a731a6393845.mov

Note that the popup **only** occurs when the reference name is changed, and will not occur if the field is changed back to what it was originally.


src/apps/schema/src/app/views/SchemaEdit/FieldEdit/FieldEdit.js
```diff
@@ -93,15 +101,19 @@ function Header(props) {
 
 export function Footer(props) {
   const [loading, setLoading] = useState(false);
+  const [warningIsOpen, setWarningIsOpen] = useState(false);
+
+  //const [name, setName] = useState(props.field.name)
```
agalin920, line 106: Leftover comment

agalin920 (approved):
LGTM

PR #1101, opened 2022-05-23
Settings FieldTypeBinary to MUI Toggle Button 

d88naimi:
FieldTypeBinary 
to MUI

```
FormLabel
ToolTip
ToggleButtonGroup
ToggleButton
```
Screen Shot 2022-05-23 at 1 31 59 PM


src/apps/settings/src/app/views/Instance/Instance.js
```diff
@@ -195,31 +205,61 @@ export default connect((state) => {
               } else if (field.key === "preferred_domain_prefix") {
                 return (
                   <div key={field.ZUID} className={styles.column}>
-                    <FieldTypeBinary
-                      key={field.ZUID}
-                      name={field.key}
+                    <FormLabel sx={{ color: "primary.dark" }}>
```
agalin920, line 208: Whats the default coloring for this? This should be a theme change

src/apps/settings/src/app/views/Instance/Instance.js
```diff
@@ -195,31 +205,61 @@ export default connect((state) => {
...
+                        direction="row"
+                        alignItems="center"
+                        sx={{
+                          my: 1,
+                        }}
+                      >
+                        <Tooltip
+                          title={`Activating the WWW setting requires DNS setup of both the apex domain and www sub-domain.`}
+                        >
+                          <InfoIcon fontSize="small" sx={{ mr: 1 }} />
```
agalin920, line 219: If you are using Stack might as well use the spacing attribute as opposed to margins

src/apps/settings/src/app/views/Robots/Robots.js
```diff
@@ -154,15 +164,30 @@ export default connect((state) => {
...
+            </Stack>
+          </FormLabel>
+          <ToggleButtonGroup
+            color="secondary"
+            size="small"
+            value={robotOn.value}
+            exclusive
+            onChange={(evt, val) => handleToggle(val)}
+          >
+            <ToggleButton value={0}>No </ToggleButton>
```
agalin920, line 188: These values are numeric and the other two are strings. Is this intended? I would expect them all to be the same type since they were all coming from the FieldTypeBinary component

agalin920 (changes requested):
comments

agalin920:
This PR should link to the theme change for ToggleButton white background

src/apps/settings/src/app/views/Robots/Robots.js
d88naimi: Okay converted to string  and cleaned up state conversion to numbers and then back to strings. Now strings are being passed all the way through as the API expects strings

agalin920 (approved):
LGTM

PR #1119, opened 2022-05-26
Use event handler on form submission

### LINKED ISSUE #1112: media:group:create
**Describe the bug**
When creating a group if you use the enter key command it will reload the app.

**To Reproduce**
1. Go to /media
2. Click on "Create Sub Group"
3. Type in new group name
4. Press "enter" key
5. See error

**Expected behavior**
App should not reload and pressing enter should create the new group.

**Screenshots**
image

**Smartphone (please complete the following information):**

**Additional context**

rjpals:
Closes #1112


agalin920:
The button has an onClick handler that also calls handleCreateGroup. A better solution would be to give the button type of 'submit'  and remove its onClick handler so only the onSubmit handler gets fired

agalin920 (changes requested):
The button has an onClick handler that also calls handleCreateGroup. A better solution would be to give the button type of 'submit' and remove its onClick handler so only the onSubmit handler gets fired

src/apps/media/src/app/components/MediaCreateGroupModal.js
```diff
@@ -45,7 +45,7 @@ export function MediaCreateGroupModal(props) {
           <Button
             variant="contained"
             color="success"
-            onClick={handleCreateGroup}
+            type="save"
```
agalin920, line 48: Why is this being added?

rjpals: This was a typo, my bad. Interestingly, it did actually fix the issue tho.

agalin920:
@rjpals MediaEditGroupModal has the same issue could you also include it in this PR

shrunyan (approved):

PR #1122, opened 2022-05-26
Content Item Versioning small screen fix

d88naimi:
This PR needs to be deployed first https://github.com/zesty-io/manager-ui/pull/1116/files I believe before this one 

For small viewports the content Item versioning header is still very tight and wraps onto a new line.

Screen Shot 2022-05-26 at 2 15 08 PM

This PR does some html clean up and I introduced MUI media query to check if  the window is below 1400px then collapse the Nav bar and Content Nav bar see video https://www.screencast.com/t/6kJvR0YI1Y

I may want to increase the media query to 1680px as this be a helpful feature for medium to small laptop sizes please let me know what you think.


agalin920 (approved):
Approved

shrunyan:
Is this a pre-existing issue? Did switching to MUI components change something?

d88naimi:
@stuart legacy was fine I believe till 1280px however now 1300px the header will wrap, I think the increase in icon size from 16px to font-size 20px may be the culprit.

PR #1123, opened 2022-05-27
Feature/large file upload

shrunyan:
By using a signed URL we can bypass the Google App Engine 32mb limit at their load balancer infrastructure and post a file directly to a GCP cloud storage bucket.


agalin920 (approved):
LGTM

PR #1147, opened 2022-06-07
MUI Zesty FieldTypeOneToOne integration

agalin920:
DEPENDANT ON @zesty-io/material release with 6/7/2022 updates

Fully virtualized and now searchable

Screen Shot 2022-06-07 at 12 06 11 AM

Screen Shot 2022-06-07 at 12 05 52 AM

Screen Shot 2022-06-09 at 4 21 23 PM

### DIFF
--- a/cypress/integration/content/content_spec.js
+++ b/cypress/integration/content/content_spec.js
@@ -286,19 +286,18 @@ describe("Content Specs", () => {
       });
 
       it("One to one Field", () => {
-        cy.get("#12-edee00-6zb866").find(".Select").click({ force: true });
-        //allow relationships to load
         cy.get("#12-edee00-6zb866")
-          .find(".Select")
-          .find('[data-value="7-480ab4-wg7x7j"]')
-          .last()
+          .find(".MuiAutocomplete-popupIndicator")
+          .click({ force: true });
+        //allow relationships to load
+        cy.get("[role=presentation]")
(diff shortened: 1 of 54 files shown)


shrunyan:
While reviewing noticed rendering perf issues. It seems to be related to the `ui` state slice selection and unrelated to this body of work. Going to hold off on merge until we have a better idea on the root cause of the issue. Investigation work happening here: https://github.com/zesty-io/manager-ui/tree/perf/ui-state-selection

This has been discussed with @agalin920 IRL. He'll be taking over finding a solution.

shrunyan:
@agalin920 a couple of things

- I am not seeing a search field appear when a list has more than 50 items, this was the previous behavior. I believe we should be able to stick to the prior filter logic as the same expectation that the full related data set has been loaded is still being met.
- The following maybe MUI details that we don't have control over but pointing out as they seem odd to me
  - When click a dropdown it selects the dropdown text
  - The trigger for opening the dropdown is not the whole input including border. i.e. click the border of the list causes selection of input text but doesn't trigger opening the dropdown

agalin920:
Sort items by created_on

agalin920:
After discussing with Zosh about concerns with search visibility. He investigated and ultimatley just requested a placeholder for when  relationship has not been selected (We still support 'None' where value is 0)

shrunyan (approved):

PR #1154, opened 2022-06-08
Rebuild tabs optimization

agalin920:
Some views are selecting the full ui store. The full ui store was propagating an update and therefore a rerender upon every single slice change content/model/file/mediaGroups thus causing massive amounts of parent-level rerenders.

This PR reduces the full UI store propagation by doing a diff check before setting new tabs.

Future notes:
Running the rebuildTabs on any slice change is overkill and can be further looked in to be refactored to only target specific changes within those slices.

It is still best practice to use selectors to select the smallest and most atomic data required in the component from a store to avoid subscribing to unutilized data.

### DIFF
--- a/src/apps/content-editor/src/app/components/Editor/Editor.js
+++ b/src/apps/content-editor/src/app/components/Editor/Editor.js
@@ -15,7 +15,6 @@ export default memo(function Editor({
   onSave,
   itemZUID,
 }) {
-  const ui = useSelector((state) => state.ui);
   const dispatch = useDispatch();
   const firstTextField = fields.find((field) => field.datatype === "text");
   const firstContentField = fields.find(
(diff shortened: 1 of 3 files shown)


src/apps/content-editor/src/app/components/Editor/Editor.js
```diff
@@ -15,7 +15,6 @@ export default memo(function Editor({
   onSave,
   itemZUID,
 }) {
-  const ui = useSelector((state) => state.ui);
```
agalin920, line 18: This was un-used here

shrunyan:
In terms of refactoring, we should wait till https://github.com/zesty-io/manager-ui/discussions/1054 gets going. As that will most likely necessitate a refactor.

PR #1155, opened 2022-06-08
cypress fix for content_spec toggle YES/NO

d88naimi:


shrunyan:
Resolved https://github.com/zesty-io/manager-ui/pull/1154

PR #1174, opened 2022-06-14
Refactor SEO redirects table to be implemented with MUI data grid

agalin920:


shrunyan:
Nice! Looking good. Few questions/feedback.

1. Can we truncate the incoming path so that the end path slug is shown versus, in the loom, the starting slug. As when urls get long the parent paths an item belong to is all that shown and you start to loss context on the item which is being redirected. 
2. Instead of "Redirect Target Not Found" lets go with "Redirect target missing meta title" and make that a link to the exact items `/meta` tab so it can be quickly set if desired.

agalin920:
@shrunyan 
Point 1 and 3 are in. 
Screen Shot 2022-06-14 at 7 08 28 PM

For point 2.
It is my understanding that it is not that the Redirect target is missing a meta title but that the redirect target AKA ZUID is not found by the content search meaning that the item that it was targeting has been deleted. When not found and trying to redirect to /content/<redirect target zuid>/meta I still get a 404.

agalin920:
Message updated to : "Redirect Target has been deleted"

shrunyan:
Table must take the available height
image

Is there any way to set initial widths on the columns? In real world use the incoming paths are long and should be afforded more space

PR #1178, opened 2022-06-16
Mui card leftovers swap

d88naimi:


src/shell/components/Head/HeadTag/HeadTag.js
```diff
@@ -217,7 +233,10 @@ export const HeadTag = (props) => {
           );
         })}
       </CardContent>
-      <CardFooter className={styles.CardFooter}>
+      <CardActions
+        className={styles.CardFooter}
```
agalin920, line 237: Looks like you deleted this class but still have it here

src/shell/components/GlobalHelpMenu/GlobalHelpMenu.less
```diff
@@ -2,7 +2,11 @@
 @import "~@zesty-io/core/typography.less";
 
 .helpModules {
-  display: flex;
+  display: grid;
```
agalin920, line 5: Since all columns are 1fr. Why can't we keep flex?

agalin920 (approved):
Approved

PR #1183, opened 2022-06-17
Append site name onto Google preview when show_in_title setting is selected

agalin920:
Screen Shot 2022-06-16 at 10 59 15 PM

### DIFF
--- a/src/apps/content-editor/src/app/views/ItemEdit/Meta/ItemSettings/ItemSettings.js
+++ b/src/apps/content-editor/src/app/views/ItemEdit/Meta/ItemSettings/ItemSettings.js
@@ -49,8 +49,6 @@ export const ItemSettings = memo(
       [meta.ZUID]
     );
 
-    console.log("testing", data, meta, web);
-
     return (
       <section className={styles.Meta}>
         <main className={styles.MetaMain}>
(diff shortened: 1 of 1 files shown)


src/apps/content-editor/src/app/views/ItemEdit/Meta/ItemSettings/ItemSettings.js
```diff
@@ -20,6 +21,12 @@ import styles from "./ItemSettings.less";
 
 export const ItemSettings = memo(
   function ItemSettings(props) {
+    const showSiteNameInMetaTitle = useSelector(
```
agalin920, line 24: This only subscribes to a specific objects value. If the specific object is not found (not all instances may have this setting key) then the variable is a still falsy value and does not render the title

src/apps/content-editor/src/app/views/ItemEdit/Meta/ItemSettings/ItemSettings.js
```diff
@@ -20,6 +21,12 @@ import styles from "./ItemSettings.less";
...
+      (state) =>
+        state.settings.instance.find(
+          (setting) => setting.key === "show_in_title"
```
agalin920, line 27: Selecting on 'key' === 'show_in_title' because it provided the most amount of context.

shrunyan (approved):

PR #1204, opened 2022-06-24
Fix cypress error due to new sidebar

rjpals:


agalin920:
@rjpals Failing tests

PR #1214, opened 2022-06-29
1133 MUI Zesty FieldTypeDate and Zesty FieldTypeDateTime

agalin920:
DEPENDANT ON https://github.com/zesty-io/material/pull/29

Screen Shot 2022-06-28 at 8 30 22 PM
Screen Shot 2022-06-28 at 8 30 35 PM
Screen Shot 2022-06-28 at 8 36 39 PM
Screen Shot 2022-06-28 at 8 36 45 PM
Screen Shot 2022-06-28 at 8 36 51 PM
Screen Shot 2022-06-28 at 9 09 25 PM
Screen Shot 2022-06-28 at 9 26 23 PM


cypress/e2e/content/content.spec.js
```diff
@@ -96,28 +96,22 @@ describe("Content Specs", () => {
...
-      cy.get("#12-63ab04-0nkwcc input").click();
+      cy.get("#12-63ab04-0nkwcc button").click();
 
-      cy.get(
-        '.flatpickr-calendar:not(.hasTime) [aria-label="March 5, 2019"]'
-      ).click();
+      cy.get("[role=dialog]").find('[aria-label="Mar 5, 2019"]').click();
 
-      cy.get("#12-63ab04-0nkwcc input").should("have.value", "2019-03-05");
+      cy.get("#12-63ab04-0nkwcc input").should("have.value", "03/05/2019");
```
shrunyan, line 103: Date format is REALLY important. we seem to be changing from a `-` delimiter to a `/` one. This could have significant implications to end consumption of these values. e.g. a customer has an API endpoint which relies on the date format. Can we specific the delimiter and keep it the same?

cypress/e2e/content/content.spec.js
```diff
@@ -96,28 +96,22 @@ describe("Content Specs", () => {
...
 
-      cy.get(".flatpickr-confirm").click();
-
-      // TODO get aria-label and capture in variable
-      // cy.get("#12-f3db44-c8kt0q .flatpickr-input.form-control.input").click();
-
-      // TODO compare aria-label value with input value
+      cy.get("#12-f3db44-c8kt0q  input").should(
+        "have.value",
+        "03/05/2019 02:30 pm"
```
shrunyan, line 113: It looks like the previous test was incomplete and did not have an expectation on the value defined. As mentioned in the prior comment format is extremely important. We need to determine the current format date/time values are sent and persisted at the API. We should then ensure the UI maintains that format and the test verifies the format.

src/apps/content-editor/src/app/components/Editor/Field/Field.js
```diff
@@ -829,21 +829,65 @@ export default function Field({
...
+                  placement="top-start"
+                  arrow
+                  title={settings.tooltip ? settings.tooltip : " "}
+                >
+                  <InfoIcon fontSize="small" sx={{ mr: 1 }} />
+                </Tooltip>
+              ) : (
+                " "
+              )}
+
```
shrunyan, line 883: Does this line break add white space to the UI?

cypress/e2e/content/content.spec.js
shrunyan: Pulled from the [source code formatting string](https://github.com/zesty-io/manager-ui/pull/1214/files#diff-f46e1bada931bfb3d22a8c7e5b48da51563447b4b5576e8f19489a1002477620R865) the format should be `YYYY-MM-DD HH:mm:ss`

src/apps/content-editor/src/app/views/ItemEdit/components/Header/ItemVersioning/ScheduleFlyout/ScheduleFlyout.js
```diff
@@ -200,14 +200,12 @@ export default class ScheduleFlyout extends Component {
                   />
                 </div>
                 <div className={styles.Row}>
-                  <FieldTypeDate
-                    type="date"
+                  <FieldTypeDateTime
                     name="publish"
                     label="Publish date and time"
-                    future={true}
```
shrunyan, line 207: Does the MUI component allow selection of future values? We should include a test which confirms this, along side the existing past value test.

src/apps/content-editor/src/app/views/ItemEdit/components/Header/ItemVersioning/ScheduleFlyout/ScheduleFlyout.js
```diff
@@ -200,14 +200,12 @@ export default class ScheduleFlyout extends Component {
...
                     value={this.state.selectedTime}
-                    datatype={"datetime"}
                     onChange={this.handleChangePublish}
+                    disablePast
```
shrunyan, line 208: Interesting. Adding note for myself to revisit and view this behavior in app. Should be fine but it is a change.

shrunyan (changes requested):

cypress/e2e/content/content.spec.js
agalin920: @shrunyan Yea so this was my thought process.

On current prod we show the following
Screen Shot 2022-07-06 at 4 40 49 PM

which then upon being sent to API gets resolved to:
date: "2019-03-05 00:00:00"
date_time: "2019-03-05 14:30:00.000000"

Given that our date time displays differently I figured there was no issue having what is displayed as long as it was being sent to the BE all the same.

All that being said I now realize it makes more sense for the value being sent directly to the BE to be displayed exactly how it is on the content editor so that users can see what date format to expect

src/apps/content-editor/src/app/views/ItemEdit/components/Header/ItemVersioning/ScheduleFlyout/ScheduleFlyout.js
agalin920: Yep it matches the behavior of future={true} which avoids the selection of a past value

src/apps/content-editor/src/app/views/ItemEdit/components/Header/ItemVersioning/ScheduleFlyout/ScheduleFlyout.js
agalin920: It always allows the selection of future values. future={true} was a prop for the old component that avoided the selection of a past value

But yea I'll write some tests for this

agalin920:
@shrunyan 
Updated date display format to display exactly what BE returns

Screen Shot 2022-07-06 at 5 28 29 PM

src/apps/content-editor/src/app/views/ItemEdit/components/Header/ItemVersioning/ScheduleFlyout/ScheduleFlyout.js
shrunyan: So the experience change introduced with this component swap is that past values are now selecteable? Do we care about this change? Can we prevent the new MUI component from allowing past value selections?

**Side note**: in regards to testing, what timezone is the browser test runner in github actions set to? What does that mean for our tests?

shrunyan:
> Updated date display format to display exactly what BE returns.

I like this direction. Only thing which is incorrect is the trailing `.000000` after the seconds value. We do not currently send time stamps structured like that to the API.

example update on zesty.pw date&time field
image

p.s. Typically that slot in the format string is timezone info. All though I have not seen it like the value the MUI component is providing. We drop timezone info as all values are stored in UTC and then converted to the local timezone for display.

src/apps/content-editor/src/app/views/ItemEdit/components/Header/ItemVersioning/ScheduleFlyout/ScheduleFlyout.js
agalin920: Correct. Im assuming the rational with future={true} was that the component can only select future values.

agalin920:
@shrunyan Yea oddly enough they don't get sent on the POST/PATCH but the BE returns .000000 upon GET the newly modifed record

So we can either send the trailing zeros to API via client to match it or modify the API

src/apps/content-editor/src/app/views/ItemEdit/components/Header/ItemVersioning/ScheduleFlyout/ScheduleFlyout.js
agalin920: No the experience is the same past values are still not selectable, this is set by a different prop 'disablePast'

shrunyan (approved):

PR #1234, opened 2022-07-07
Fix error where Metrics crashes when no media was requested

### LINKED ISSUE #1225: TypeError: Cannot read properties of null (reading 'map')
Sentry Issue: [MANAGER-UI-1JX](https://sentry.io/organizations/zestyio/issues/3358604213/?referrer=github_integration)

```
React ErrorBoundary Error: Cannot read properties of null (reading 'map')
  at requestData (./src/apps/reports/src/app/views/Metrics/index.js:169:3)
  at Ru (./src/apps/reports/src/app/views/Metrics/index.js:157:17)
  at o (/vendors.adf937ee00c199fc8603.js:102:436680)
  at Metrics (./src/apps/reports/src/app/views/Metrics/index.js:63:16)
  at t (/vendors.adf937ee00c199fc8603.js:102:5890381)
...
(22 additional frame(s) were not displayed)

TypeError: Cannot read properties of null (reading 'map')
  at Mu (./src/apps/reports/src/app/views/Metrics/index.js:496:39)
  at ss (/vendors.adf937ee00c199fc8603.js:102:5670460)
  at Ua (/vendors.adf937ee00c199fc8603.js:102:5722509)
  at Ml (/vendors.adf937ee00c199fc8603.js:102:5709783)
  at Sl (/vendors.adf937ee00c199fc8603.js:102:5709711)
...
(15 additional frame(s) were not displayed)
```

rjpals:
Closes #1225

### DIFF
--- a/src/apps/reports/src/app/views/Metrics/index.js
+++ b/src/apps/reports/src/app/views/Metrics/index.js
@@ -58,12 +58,12 @@ const numberWithCommas = (x) => {
 
 const floatWithCommas = (x) => {
   x = x.toFixed(2);
-  let parts = x.toString().split(".");
+  const parts = x.toString().split(".");
   parts[0] = parts[0].replace(/\B(?=(\d{3})+(?!\d))/g, ",");
   return parts.join(".");
 };
 
-export default function Metrics(props) {
+export default function Metrics() {
   const zuid = useSelector((state) => state.instance.ZUID);
   const [timePeriod, setTimePeriod] = useState(30);
   const { start, end } = getDates(timePeriod);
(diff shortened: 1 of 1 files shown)


src/apps/reports/src/app/views/Metrics/index.js
```diff
@@ -58,12 +58,12 @@ const numberWithCommas = (x) => {
 
 const floatWithCommas = (x) => {
   x = x.toFixed(2);
-  let parts = x.toString().split(".");
+  const parts = x.toString().split(".");
```
shrunyan, line 61: No changed needed but want to provide feedback for future reference. With an error fix it is best to avoid making additional changes which are not directly related to the error at hand. In this case the change from `let` to `const`. Can make if difficult to pin point the exact change which solves the issue. Obviously with this one it's not all that difficult but it's a good habit to follow so when there are errors that require cross cutting changes the PR is easier to understand knowing that every change is necessary. 

Additionally in terms of the change from let to const this particular line and it's following line is a good example of the debate on where it is appropriate to use const in javascript. While the assignment of `parts` is not changed on line 62 the underlying value is. i.e. The array the const is referring to has its values modified. While technically okay, like the other point, it is best to avoid making a habit of describing structures which have their values modified with `const`.

PR #1241, opened 2022-07-22
Typescript

rjpals:
This adds the ability to use TypeScript and converts a single file (`src/shell/notifications.js`) to TS. 

With this configuration, TS is entirely optional. The source tree consist of TS and JS interoperating with each other and developers will be able to freely choose between them (for now.) This will allow us to incrementally adopt TS without an all-out rewrite.

For information, see [the spec](https://docs.google.com/document/d/1e_ALIynxoEWcxB4pb6zl_tqzUKIpnxjS4oag2BAhScI)

### DIFF
--- a/package.json
+++ b/package.json
@@ -101,7 +101,6 @@
     "@babel/plugin-transform-runtime": "^7.10.4",
     "@babel/preset-env": "^7.14.7",
     "@babel/preset-react": "^7.14.5",
-    "@babel/preset-typescript": "^7.18.6",
     "@foreachbe/cypress-tinymce": "^1.0.0",
     "@sentry/cli": "^2.2.0",
     "@sentry/webpack-plugin": "^1.18.9",
--- a/package.json
+++ b/package.json
@@ -101,6 +101,7 @@
     "@babel/plugin-transform-runtime": "^7.10.4",
     "@babel/preset-env": "^7.14.7",
     "@babel/preset-react": "^7.14.5",
+    "@babel/preset-typescript": "^7.18.6",
     "@foreachbe/cypress-tinymce": "^1.0.0",
     "@sentry/cli": "^2.2.0",
     "@sentry/webpack-plugin": "^1.18.9",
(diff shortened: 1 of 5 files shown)


src/shell/store/notifications.ts
```diff
@@ -1,4 +1,10 @@
-export function notifications(state = [], action) {
+type Action = {
```
agalin920, line 1: Are we gonna go with 'type' over 'interface' ?

src/shell/store/notifications.ts
```diff
@@ -8,15 +14,21 @@ export function notifications(state = [], action) {
...
+        return notification;
       });
 
     default:
       return state;
   }
 }
 
-export function notify(data) {
+type Data = {
```
agalin920, line 25: I think Data is too arbitrary and not indicative enough of what this is. Shouldn't we should name this Notification? Especially in the case we ever need to export this type to a component

agalin920 (changes requested):

package.json
```diff
@@ -101,6 +101,7 @@
     "@babel/plugin-transform-runtime": "^7.10.4",
     "@babel/preset-env": "^7.14.7",
     "@babel/preset-react": "^7.14.5",
+    "@babel/preset-typescript": "^7.18.6",
```
agalin920, line 104: Is this needed?

src/shell/store/notifications.ts
```diff
@@ -1,4 +1,10 @@
-export function notifications(state = [], action) {
+type Action = {
+  type: "NEW_NOTIFICATION" | "REMOVE_NOTIFICATION";
+  data: Data;
+  epoch: Date;
+};
+
+export function notifications(state = [], action: Action) {
```
agalin920, line 7: State should be defined

src/shell/webpack.config.js
```diff
@@ -74,8 +74,10 @@ module.exports = async (env) => {
         utility: path.resolve(__dirname, "../utility"),
         apps: path.resolve(__dirname, "../apps"),
       },
+      extensions: [".ts", ".js"],
     },
     plugins: [
+      /*
```
shrunyan, line 80: why has this been commented out?

shrunyan (approved):

PR #1242, opened 2022-07-26
Activity Log

agalin920:


cypress/e2e/reports/activity-log/activity-log.spec.js
```diff
@@ -0,0 +1,232 @@
...
+  });
+
+  describe("Resource Tab", () => {
+    it("Navigates to Resource Detail on Resource Item click", () => {
+      cy.waitOn("/v1/env/audits*", () => {
+        cy.visit(
+          "/reports/activity-log/resources?from=2022-07-14&to=2022-07-16"
+        );
+      });
+      cy.wait(1000);
```
shrunyan, line 117: why an additional `wait` after the `waitOn`?

cypress/e2e/reports/activity-log/activity-log.spec.js
```diff
@@ -0,0 +1,232 @@
...
+        "/reports/activity-log/resources/7-f28fd4d4a9-qtjb66"
+      );
+    });
+  });
+
+  describe("Skeletons", () => {
+    it("Displays All Skeletons while initial API is called", () => {
+      cy.intercept("/v1/env/audits*", (req) => {
+        req.continue(async (res) => {
+          await new Promise((resolve) => setTimeout(resolve, 4000));
```
shrunyan, line 130: This is subtle complex. Please add a comment explaining. My understanding is we are inserting an explicit 4 seconds of time to ensure skeletons are rendered to test for their presence.

src/apps/content-editor/src/app/views/ItemEdit/Content/Actions/Widgets/WidgetDraftHistory/WidgetDraftHistory.js
```diff
@@ -53,21 +48,35 @@ export default connect((state, props) => {
...
+            </ul>
+            {/* TODO: Replace this link with Zesty MUI AppLink */}
+            <Link
+              color="#FF5D03"
+              underline="hover"
+              href="#"
+              onClick={(evt) => {
+                evt.preventDefault();
+                history.push(
+                  `/reports/activity-log/resources/${props.itemZUID}`
```
shrunyan, line 71: I won't mark this as blocking to deploy but we need a better abstraction for internal app link routing than supplanting default a tag behavior.

src/apps/reports/src/app/views/ActivityLog/components/ActionTimelineItem/TimelineItem.js
```diff
@@ -0,0 +1,179 @@
...
+          variant="caption"
+          component="div"
+          color="text.secondary"
+          sx={{
+            whiteSpace: "nowrap",
+            overflow: "hidden",
+            textOverflow: "ellipsis",
+          }}
+        >
+          {" "}
```
shrunyan, line 152: Is this spacing intentional?

src/shell/index.js
```diff
@@ -9,6 +9,8 @@ import { Router } from "react-router-dom";
 import { ThemeProvider, createTheme } from "@mui/material/styles";
 import { theme } from "@zesty-io/material";
 
+import "chart.js/auto";
```
shrunyan, line 12: Why does this happen in the shell?

src/utility/resolveResourceType.js
```diff
@@ -0,0 +1,11 @@
+export const resolveResourceType = (ZUID) => {
```
shrunyan, line 1: This should probably exist in this dep: https://github.com/zesty-io/zuid-js

src/utility/instanceZUID.js
```diff
@@ -0,0 +1,2 @@
+const instanceZUID = window.location.host.split(".")[0];
```
shrunyan, line 1: This happens on app boot and is added to the store. Can the store value be selected instead of adding this util?

shrunyan (changes requested):
The `.DS_Store` files should not be part of the PR. Please check the .gitignore and ensure those OS files are excluded.

src/apps/content-editor/src/app/views/ItemEdit/Content/Actions/Widgets/WidgetDraftHistory/WidgetDraftHistory.js
agalin920: Agreed I believe its marked with a TODO to replace with the yet to be completed Zesty MUI AppLink

src/utility/instanceZUID.js
agalin920: This was added in the previous RTK Query implementation and it is used to construct the RTK Query base url so I don't think I would be able to select it from the store

shrunyan:
@zcolah it looks like we do not have hover states on the resource list view items. Are these accounted for in the design?

src/shell/index.js
agalin920: Ah this is required for v3 and added it here to verify all charts. Will move

cypress/e2e/reports/activity-log/activity-log.spec.js
agalin920: Testing purposes. Will remove

shrunyan:
@agalin920 functional feedback.

- view button on the resource > item view is not deep linking me to the item edit view. seems to link to /content regardless of log type
- when clicking into an item from the resource list view it seems the query parameters are added introducing a additional browser history state causing the back button to not take me back to the resource list view, unless pressed multiple times.

agalin920:
@shrunyan 
"view button on the resource > item view is not deep linking me to the item edit view. seems to link to /content regardless of log type "

This is caused by newly added log items not having proper URLs in the API response. Need to log a bug for it and transition to FE URL construction  

"when clicking into an item from  the resource list view it seems the query parameters are added introducing a additional browser history state causing the back button to not take me back to the resource list view, unless pressed multiple times"

This should be fixed now

zcolah (approved):
**Great job Andres. Future fixes we will make are:** 

1. Identifying a manner to incorporate a scroll bar into our UI
2. Reviewing text updating below Activity Log
3. Finalizing name (b/w Activity Log and Audit Log)
4. Calling out items with Missing Meta Titles properly 
5. Handle edge case of created not coming up for old items (before 2020)

shrunyan (approved):

PR #1359, opened 2022-09-15
Tabs code changes modal

rjpals:
PR to add a modal when closing a Code tab with unsaved changes


src/apps/code-editor/src/store/files.js
```diff
@@ -174,10 +174,17 @@ export function files(state = [], action) {
           file.ZUID === action.payload.ZUID &&
           file.status === action.payload.status
         ) {
+          /*
+            The code comparison could potentially be very expensive, so we check
+            the dirty flag in the store and only do the comparison if needed
+            (taking advantage of lazy evaluation)
+          */
+          const dirty = file.dirty || file.code !== action.payload.code;
+          console.log({ dirty });
```
shrunyan, line 183: remove

src/apps/code-editor/src/store/files.js
```diff
@@ -214,6 +221,29 @@ export function files(state = [], action) {
 
       return files;
 
+    case "UNMARK_FILE_DIRTY":
+      console.log({ state });
```
shrunyan, line 225: remove

src/apps/code-editor/src/store/files.js
```diff
@@ -214,6 +221,29 @@ export function files(state = [], action) {
...
+      files = state.map((file) => {
+        console.log({ file, p: action.payload });
```
shrunyan, line 227: remove

src/shell/components/global-tabs/GlobalTabs.tsx
```diff
@@ -103,6 +104,7 @@ export default memo(function GlobalTabs() {
 
   return (
     <ThemeProvider theme={theme}>
+      <GlobalDirtyCodeModal />
```
shrunyan, line 107: These should probably be in the code app so it is co-located towards the code it operates on.

src/shell/components/global-tabs/components/Tab.tsx
```diff
@@ -188,7 +188,8 @@ export const ActiveTab: FC<ActiveTab> = ({ tabWidth }) => {
       tabWidth={tabWidth}
       tab={activeTab}
       onClick={() => {
-        if (isPinned) dispatch(unpinTab(activeTab));
+        // force unpin because we don't want to show the modal on the active tab
+        if (isPinned) dispatch(unpinTab(activeTab, true));
```
shrunyan, line 192: style: please use `{}` and denote code blocks

src/apps/code-editor/src/app/components/PendingEditsModal/PendingEditsModal.js
```diff
@@ -0,0 +1,119 @@
+import { memo, useState, useEffect } from "react";
```
shrunyan, line 1: I'm really confused on which component is the dirty code state modal. 

- src/shell/components/global-tabs/components/DirtyCodeModal.tsx
- src/apps/code-editor/src/app/components/PendingEditsModal/PendingEditsModal.js

These 2 components seem to be doing the same job but are loaded into the runtime at different places. What am I missing?

src/shell/store/ui.ts
```diff
@@ -335,11 +358,31 @@ export function pinTab({ pathname, search }: TabLocation) {
...
 
-export function unpinTab({ pathname, search }: TabLocation) {
+// TODO this should probably be broken up, or have a name like "attempt unpin"
+export function unpinTab({ pathname, search }: TabLocation, force = false) {
   return (dispatch: Dispatch, getState: () => AppState) => {
     const state = getState();
     const parsedPath = parsePath({ pathname, search });
     const tab = createTab(state, parsedPath);
+    const { parts } = parsedPath;
+    console.log({ parsedPath });
```
shrunyan, line 368: remove. please audit this PR for `console.logs`

shrunyan (changes requested):

src/shell/components/global-tabs/GlobalTabs.tsx
rjpals: This is the modal that activates when you try to unpin an dirty code tab. It is NOT the modal that activates when you navigate away from a dirty code file. It is necessary for this modal to be available anywhere in the app where a user could possibly unpin a tab. For that reason it cannot be in code; because it would be unmounted and then unavailable outside of Code. 

According to our convo earlier today, you don’t believe this modal should exist. I will wait for direction on that based on Zosh’s user feedback sessions this week.

src/apps/code-editor/src/app/components/PendingEditsModal/PendingEditsModal.js
rjpals: They are logically different modals. I am in the process of refactoring so that they use the same component for *presentation,* but they are quite different logically:

`PendingEditsModal` is part of the Code app and is activated by the `getUserConfirmation` part of the `react-router` API whenever a user navigates away from the Code app. It operates on the code file the user is editing, getting this context from the URL path parts.

`DirtyCodeModal` exists globally. This can be triggered from anywhere in manager when a user unpins a tab. It operates on the URL that is saved in the pinned tab.

PR #1380, opened 2022-09-26
Implement folder actions

agalin920:
New folder functionality
Rename folder functionality
Delte Folder functionality

### DIFF
--- a/src/apps/media/src/app-revamp/components/Header.tsx
+++ b/src/apps/media/src/app-revamp/components/Header.tsx
@@ -24,11 +24,10 @@ interface Props {
   binId?: string;
   groupId?: string;
 }
+type Dialogs = "delete" | "rename" | "new" | null;
 
 export const Header = ({ title, id, binId, groupId }: Props) => {
-  const [openRenameDialog, setOpenRenameDialog] = useState(false);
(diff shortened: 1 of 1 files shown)


src/apps/media/src/app-revamp/components/Header.tsx
```diff
@@ -1,30 +1,138 @@
...
+  id?: string;
+  binId?: string;
+  groupId?: string;
 }
 
-export const Header = ({ title }: Props) => {
+export const Header = ({ title, id, binId, groupId }: Props) => {
+  const [openRenameDialog, setOpenRenameDialog] = useState(false);
+  const [openNewFolderDialog, setOpenNewFolderDialog] = useState(false);
+  const [openDeleteDialog, setOpenDeleteDialog] = useState(false);
```
rjpals, line 31: Since these dialogs are mutually exclusive, I'd prefer something like this stylistically:

```ts
type Dialogs = 'delete' | 'rename' | 'new' | null
const [openDialog, setOpenDialog] = useState<Dialogs>(null)
```

Wondering what your thoughts on this would be?

rjpals (approved):

PR #1388, opened 2022-09-28
Update: draggable children from useState to useRef to avoid causing rerender

### LINKED ISSUE #38: Rearranging schema items has a lag
**Describe the bug**
Dragging items in a schema to rearrange their order has horrible response times. I often drag and when I let go, I wait 20 seconds for it to refresh with the change.
**To Reproduce**
1. Go to Schema
2. Click on a content model
3. Click on a field and drag to reorder it
4. See slow response time

**Expected behavior**
Expected behavior is to be able to drag and drop to reorder without any lag.

**Additional context**
This was originally reported by Dan Peliter from GroupBy on Slack. Original message:
@here not sure if it's server issues happening now, or if it's just the way the feature works, but dragging items in a schema to rearrange their order has horrible response times. I often drag and when I let go, I wait 20 seconds for it to refresh with the change. The CMS has a simple `Sort Order` field that works beautifully for end users....not sure why the schema builder isn't just using the same thing. It would be much easier/faster.

JacobDiazCruz:
Closes #38

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,6 +5,7 @@
   "requires": true,
   "packages": {
     "": {
+      "name": "manager-ui",
       "version": "1.0.1",
       "license": "Commons Clause License Condition v1.0",
       "dependencies": {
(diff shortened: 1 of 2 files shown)


JacobDiazCruz:
Difference between the old and fixed version:
old version: https://www.screencast.com/t/6puMgMZVmb
fixed version: https://www.screencast.com/t/0xOyxo9oF6oa

JacobDiazCruz:
@agalin920 Requesting for review again.

shrunyan:
closed in favor of https://github.com/zesty-io/manager-ui/pull/1866

PR #1418, opened 2022-10-12
Update dropzone to support controlled inputs

agalin920:
Screen Shot 2022-10-12 at 1 07 06 PM

Dropzone clones children before children actually get processed un-updated. Therefore on an expected update, it would clone the previous state of children before cloning the new state of children. This was causing unintended side effects when working with controlled inputs


shrunyan:
closed in favor of https://github.com/zesty-io/manager-ui/pull/1866

PR #1444, opened 2022-10-18
Feature/dam thumbnail

shrunyan:


shrunyan:
While I would ideally like to see our front-end code follow more of a Higher Order Component code structure this PR was not finished and is now aged out.

PR #1506, opened 2022-11-07
Convert spaces to dashes when searching

agalin920:


shrunyan:
I have concerns with this change. There are inconsistency with the experience. Currently the UI allows editing of a filename and including spaces within that name but it is not the true state of the file as once you save it and the new file url is generated it replaces the spaces with hypens. 

Essentially there are 2 concerns I have. 

1. an unclear and inconsistent expiration on how this should all behave
2. a judgement and forcing users out of using spaces. I don't think we should do this.

PR #1512, opened 2022-11-08
Store media folder sort order in localstorage

agalin920:


shrunyan (approved):

shrunyan:
Are we using the key `zesty:navMedia:sort` to persist the selected sort order? If so this needs to be per instance. I forget, are instances scoped by idb database or key?

PR #1580, opened 2022-12-01
fix: added space support for content nav filter

### LINKED ISSUE #1568: Navigation Filter Doesn't Allow for Spaces
image

theofficialnar:
Added support for spaces in content nav filter
Closes #1568 

Screenshot:
nav-filter-fix


theofficialnar:
> Good solution. Leaves user input intact and only transforms for the purpose of filtering.
> 
> Can we go one step beyond to ensure the above for all input and not just spaces? In this case moving the toLowerCase?

Sure! PR updated.

PR #1660, opened 2022-12-21
Do not throw an exception when a duplicate request occurs on ItemEdit.js

### LINKED ISSUE #1617: TypeError: Cannot read properties of undefined (reading 'status')
Sentry Issue: [MANAGER-UI-1MA](https://sentry.io/organizations/zestyio/issues/3427548125/?referrer=github_integration)

```
TypeError: Cannot read properties of undefined (reading 'status')
  at call (./src/apps/content-editor/src/app/views/ItemEdit/ItemEdit.js:133:24)
...
(1 additional frame(s) were not displayed)
```

rjpals:
If a user switches between items too fast, a duplicate request will be triggered for an item. As described in #1215 the request middleware will simply return `undefined` for the duplicate request. This causes an exception and therefore a Sentry error, but it **does not** break the user experience.

The solution is to simply return early if the middleware returns `undefined` -- this will prevent an error from being sent to Sentry & keep the loading component displayed until the request is resolved.

Closes #1617

### DIFF
--- a/src/apps/content-editor/src/app/views/ItemEdit/ItemEdit.js
+++ b/src/apps/content-editor/src/app/views/ItemEdit/ItemEdit.js
@@ -130,6 +130,12 @@ export default function ItemEdit() {
     try {
       const itemResponse = await dispatch(fetchItem(modelZUID, itemZUID));
 
+      // Likely due to a duplicate request which causes middleware to return undefined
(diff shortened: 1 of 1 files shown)


shrunyan:
Close in favor of https://github.com/zesty-io/manager-ui/pull/1696

PR #1696, opened 2023-01-05
Update api middleware to return a promise to duplicate requests

rjpals:
This is an experimental solution to the issue described in #1215 

The idea is to keep a map of `uri` to request promise for each inflight request. If duplicate requests occur, the second and subsequent requests will return a reference to the promise representing the first request. This ensures that all requests with the same URI will be resolved at the same time with the same data.

### DIFF
--- a/src/shell/store/middleware/api.js
+++ b/src/shell/store/middleware/api.js
@@ -3,25 +3,30 @@ import { request } from "utility/request";
 import { fetchItems } from "shell/store/content";
 import { fetchFields } from "shell/store/fields";
 
-const inflight = [];
+// Map of URI to inflight request promise
+const inflight = {};
 export const fetchResource =
   (store) =>
   (next) =>
   (action = {}) => {
     if (action.type === "FETCH_RESOURCE") {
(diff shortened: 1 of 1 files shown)


shrunyan:
@rjpals would you please include a test case which simulates the multiple requests scenario. Ensure the test case triggers this code path. Run the test case against the existing solution and your proposed solution, to ensure there isn't a deviation in results.

shrunyan:
@agalin920 per point 2, I believe this code path should be independent of RTK. As any slices of the store converted to the required RTK shape get auto generated action names therefore the existing actions/reducers should be isolated from the RTK logic and state. Are there any known places where we have intermingled the store state between legacy and RTK actions?

shrunyan:
@agalin920 as a whole is this solution an improvement over the issues it would introduce. i.e. do the pros out weigh the cons? If so lets move forward with this solution.

shrunyan:
related https://github.com/zesty-io/manager-ui/pull/1660

shrunyan (approved):

PR #1740, opened 2023-01-18
Add date rang filter to media insgihts

agalin920:
Screen Shot 2023-01-18 at 10 34 01 AM

### DIFF
--- a/src/apps/media/src/app/views/InsightsMedia.tsx
+++ b/src/apps/media/src/app/views/InsightsMedia.tsx
@@ -1,10 +1,17 @@
 import { FC, useState, useMemo } from "react";
-import { Box, Typography, Card } from "@mui/material";
-import { useSelector } from "react-redux";
 import {
-  useGetUsageQuery,
-  useGetRequestsQuery,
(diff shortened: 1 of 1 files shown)


src/apps/media/src/app/views/InsightsMedia.tsx
```diff
@@ -41,16 +57,17 @@ export const InsightsMedia: FC = () => {
...
-    ]);
+    useGetUsageQuery([priorStartDate.format(), priorEndDate.format()]);
   const { data: usage, isFetching: isUsageFetching } = useGetUsageQuery([
-    moment(date).subtract(1, "months").format(),
-    moment(date).format(),
+    startDate.format(),
+    endDate.format(),
   ]);
 
+  const [anchorEl, setAnchorEl] = useState<null | HTMLElement>(null);
```
shrunyan, line 68: Is there a potential for a memory leak holding a reference to an element in state? If the button was clicked, setting state to reference the DOM element, and then the view is changed unmounting this component wouldn't this create a detached DOM node?

shrunyan (approved):

PR #1762, opened 2023-01-24
Group network errors by a custom fingerprint on sentry

agalin920:
Due to stack traces differing when triggering a network error exception, Sentry is not grouping them properly. This change captures the error with a custom fingerprint so they get aggreagted.

This has been tested and confirmed on Sentry
Screen Shot 2023-01-23 at 9 07 31 PM

### DIFF
--- a/src/utility/request.js
+++ b/src/utility/request.js
@@ -1,7 +1,8 @@
 import Cookies from "js-cookie";
 import { rest } from "lodash";
 import { Sentry } from "./sentry";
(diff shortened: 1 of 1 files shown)


shrunyan:
Lets include a notification with this informing, via UI, the user the network connection failed.

shrunyan:
Related https://github.com/getsentry/sentry-javascript/discussions/5334

shrunyan (approved):

PR #1655, opened 2022-12-21
FIX: Add an error message screen when model fails to load

JacobDiazCruz:
Closes #1618 
Closes #2000 
Closes #1809 
Closes #1739 

image

PR #1768, opened 2023-01-25
Fixed: undefined model zuid

### LINKED ISSUE #1739: TypeError: Cannot read properties of undefined (reading 'ZUID')
Sentry Issue: [MANAGER-UI-1TE](https://sentry.io/organizations/zestyio/issues/3696977529/?referrer=github_integration)

```
React ErrorBoundary Error: Cannot read properties of undefined (reading 'ZUID')
  at xu (./src/apps/content-editor/src/app/components/Editor/Editor.js:10:3)
  at Content (./src/apps/content-editor/src/app/views/ItemEdit/Content/Content.js:12:14)
  at ItemEdit (./src/apps/content-editor/src/app/views/ItemEdit/ItemEdit.js:65:20)
  at Pp (./src/apps/content-editor/src/app/ContentEditor.js:29:25)
  at Mb (./src/shell/views/Shell/Shell.js:29:20)
...
(22 additional frame(s) were not displayed)

TypeError: Cannot read properties of undefined (reading 'ZUID')
  at xu (./src/apps/content-editor/src/app/components/Editor/Editor.js:155:43)
  at call (./src/apps/content-editor/src/app/views/ItemEdit/ItemEdit.js:162:9)
...
(18 additional frame(s) were not displayed)
```

JacobDiazCruz:
Closes #1739

### DIFF
--- a/src/apps/content-editor/src/app/components/Editor/Editor.js
+++ b/src/apps/content-editor/src/app/components/Editor/Editor.js
@@ -152,8 +152,10 @@ export default memo(function Editor({
           <h1 className={styles.Display}>No fields have been added</h1>
           <h2 className={styles.SubHead}>
             Use the{" "}
-            <AppLink to={`/schema/${model.ZUID}`}>Schema Builder</AppLink> to
-            define your items content
+            <AppLink to={`/schema/${model?.ZUID || ""}`}>
+              Schema Builder
+            </AppLink>{" "}
+            to define your items content
           </h2>
         </div>
       )}


src/apps/content-editor/src/app/components/Editor/Editor.js
```diff
@@ -152,8 +152,10 @@ export default memo(function Editor({
           <h1 className={styles.Display}>No fields have been added</h1>
           <h2 className={styles.SubHead}>
             Use the{" "}
-            <AppLink to={`/schema/${model.ZUID}`}>Schema Builder</AppLink> to
-            define your items content
+            <AppLink to={`/schema/${model?.ZUID || ""}`}>
```
shrunyan, line 155: I do not think we should render a non-functional link if there is no model data present

shrunyan (changes requested):

JacobDiazCruz:
> @JacobDiazCruz Have you been able to replicate this error? What are the steps to get into this state?

No, I haven't successfully replicated it on dev, all content items have model ZUID. But I just added the fix in case this happens on prod.

src/apps/content-editor/src/app/components/Editor/Editor.js
JacobDiazCruz: Do you think it's better to just show an error message if there's no model ZUID? Since right now, what I did is redirect the users to /schema in case the ZUID is not present.

JacobDiazCruz: Makes sense to just use the modelZUID from the url instead of the model data. I will revise it on my next commit.

shrunyan:
closed in favor of https://github.com/zesty-io/manager-ui/pull/1655

PR #1808, opened 2023-02-08
Show both sass and less variable indicators

agalin920:
Screen Shot 2023-02-08 at 10 33 39 AM


giseleblair (changes requested):
@zcolah let's review how we can make this UI update so it's clearer to the user how style settings can be used

PR #1998, opened 2023-04-17
FIX: Date range filter fix

### LINKED ISSUE #1959: Activity Log: Date Filter - Change the title from Date to Select Date Range
Change the title from Date to Select Date Range
CleanShot 2023-03-29 at 16 05 33@2x

### LINKED ISSUE #1960: Activity Log - Date Range Filter is not in active state when it should be in activity log
The date filter should be active (orange) with the date range (see query parameter) on all 4 tabs of Activity Log. This is because they are only for the last 90 days in Activity Log. Provided more detail on a call with Nar. 

CleanShot 2023-03-29 at 16 06 59@2x

theofficialnar:
Closes #1959 
Closes #1960 

image

### DIFF
--- a/cypress/e2e/reports/activity-log/activity-log.spec.js
+++ b/cypress/e2e/reports/activity-log/activity-log.spec.js
@@ -90,13 +90,13 @@ describe("Reports > Activity Log > Home", () => {
         );
       });
 
-      const from = moment()
+      const from = moment("2022-07-14")
         .hours(0)
         .minute(0)
         .second(0)
         .millisecond(0)
         .format("x");
-      const to = moment()
+      const to = moment("2022-07-14")
         .add(1, "day")
         .hours(0)
         .minute(0)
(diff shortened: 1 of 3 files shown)


cypress/e2e/reports/activity-log/activity-log.spec.js
```diff
@@ -90,13 +90,13 @@ describe("Reports > Activity Log > Home", () => {
         );
       });
 
-      const from = moment()
+      const from = moment("2022-07-14")
```
shrunyan, line 93: Why are we setting an initial value of july 14th 2022?

theofficialnar: This is because previously before this fix was applied, the initial date value was not passed to the datepicker, but now that we are passing the initial date value, we had to set the date for the test so that we can easily let cypress runner to pick the date on the datepicker, otherwise we had to figure out how many times cypress has to click the next month button in order to get to the current date.

shrunyan: excellent. Please add that explanation as a comment. That is going to be important information if any revisits this test.

shrunyan (approved):

shrunyan (approved):

PR #2008, opened 2023-04-18
Global Search: Add search page filters

theofficialnar:
- Added sort by, user and date filter
- Updated date filter to include a date range picker

image
image
image


theofficialnar:
> @theofficialnar The filters look a bit awkward with no white space below it. Is that according to the design?

Yeah, I just followed what was on Figma

theofficialnar:
@agalin920  Actually, now that you've mentioned it. I just realized that the filters needed to have a background color. Wasn't really that visible on the laptop monitor when I was looking at figma, I'll fix the background color on my next PR.

PR #2034, opened 2023-05-09
Global Search: Include models in top search suggestions

finnar-bin:
- Include schema models in top suggestions
- Allow users to search schema medals via the ff:
  - Model name
  - Model ZUID
  - Model data type
  - User ZUID

### Model name
image
image

### Model ZUID
image

### Model data type
image
image
image

### User ZUID
image

### User name
image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,7 +5,6 @@
   "requires": true,
   "packages": {
     "": {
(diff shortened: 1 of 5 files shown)


theofficialnar:
Included in #2041

PR #2072, opened 2023-06-14
Feat: Global search phase 6

theofficialnar:
- Add media results in global search
- Search media via the ff:
  - File name
  - File zuid
  - Folder name
- Add media option in resource type filters

### File / Folder name
Screenshot

### File ZUID
Screenshot(1)

### Filters
Screenshot(2)
Screenshot(4)

### Search Page
Screenshot(3)


src/shell/views/SearchPage/SearchPage.tsx
```diff
@@ -197,6 +270,29 @@ export const SearchPage: FC = () => {
     return _results;
   }, [results, params]);
 
+  const renderBody = () => {
```
shrunyan, line 273: why is this extracted into a function versus being inline with a ternary operator?

theofficialnar: Just thought it would be a bit cleaner to extract it to a function. I can make it inline instead if you think it's better that way.

theofficialnar: @shrunyan @agalin920 already updated, moved the code inline

PR #2096, opened 2023-06-28
Fix media app jitter in modal when zoom enabled

agalin920:


src/apps/settings/src/app/views/Styles/Styles.js
```diff
@@ -23,6 +23,7 @@ import CloseIcon from "@mui/icons-material/Close";
 
 import styles from "./Styles.less";
 import { useMetaKey } from "../../../../../../shell/hooks/useMetaKey";
+import { MemoryRouter } from "react-router";
```
shrunyan, line 26: Why the addition of `MemoryRouter`? Does not seem like that is referenced anywhere in this change.

shrunyan (approved):

PR #2106, opened 2023-07-08
Feature/content analytics dashboard

agalin920:
Screenshot 2023-07-07 at 8 04 20 PM

Screenshot 2023-07-07 at 8 00 48 PM

closes https://github.com/zesty-io/manager-ui/pull/2106


zcolah:
1. Recent Edits Table should not show Stats and Views columns
https://www.figma.com/file/O14FlIdJBki8a816pP7i4t/Content-App?type=design&node-id=835%3A147447&mode=design&t=eMU3fI0T93lhSQ5n-1

zcolah:
VQA: https://www.figma.com/file/O14FlIdJBki8a816pP7i4t/Content-App?type=design&node-id=1399%3A151736&mode=design&t=eMU3fI0T93lhSQ5n-1

shrunyan (commented):
I am marking this PR as approved as all my following comments are more about structure and not relating to the functionality of the code. I am not requesting any changes. Just providing some thoughts around code choices

1)
The newly introduced components are either using a 
``` 
if {
return ...
} 

return ...
```
or 
```
if() return ...
return ...
``` 

I would prefer these to be full 
```
if () { 
return ...
} else {
return ...
}
```

I believe it helps with legibility and understanding the interconnection of the statements.

2)
There has been a util.js file introduced. These always strike me as a code smell. We should prefer to find another pattern for sharing logic when needed. i.e. external package, pushing logic to the store layer, etc.

3)
I struggle with legibility of a lot of these components. I think the primary issue is all of the style props passed to the MUI components. It takes up a lot of visual space which has to be mentally parsed to get to the logicl parts which are of concern. e.g. event handlers, operations and state references.

Not sure there is much to do about that. The only thing I would suggest is maybe a need to break components down further. The file: https://github.com/zesty-io/manager-ui/pull/2106/files#diff-40d5a35a5a4c5896734415c0377c0c0f5af189a1e0da3f628da7b248d4ae1ab7 possibly being an example where that could help.

shrunyan (approved):

shrunyan:
> ....what the main argument is in order to avoid it so really curious.

My main argument is it creates a space to dump things. Making it easy to avoid dealing with hard questions of why isn't the functionality slipping nicely into place? Is there a code structure issue? etc... i.e. "Not sure how to work a piece of functionality into the code structure... throw it in the utils."

The concern also trends towards large teams with a wide set of skill levels. Creating a scenario that lets folks avoid learning where to properly place code vs throwing it into the utils. 

**It's not a strong argument** which is why I'm not totally opposed to them just more so if we can we should avoid. 

---

I generally agree with your philosophy on RTK. Which as you note leads to;

> ...moving it to a shareable location makes sense.

Begging the question what is an ideal shareable location? util? Is there a single util across the entire code base? Does it contain all shareable functions regardless of relation? Do we split grouped sets of shareable functions into separate logic util packages? Are these externalized from the project and included as dependencies? What we would be a good DX after these choices are made and have to reasoned about months later? e.g. Where do I go look to find that shared functionality? How do I identify if it is the shared functionality I need?

I feel these questions can create paralysis for devs and lead people to just writing the functionality they need and then code duplication.

PR #2232, opened 2023-08-16
Only provide the option to parent Homepage if model is headless

agalin920:


src/apps/schema/src/app/components/CreateModelDialogue.tsx
```diff
@@ -138,7 +132,9 @@ export const CreateModelDialogue = ({ onClose, modelType = "" }: Props) => {
               canonicalTagMode: 1,
               metaLinkText: model.label,
               metaTitle: model.label,
-              parentZUID: model.parentZUID || "0",
+              parentZUID: model.parentZUID.startsWith("7-")
```
shrunyan, line 135: This seems like it would prevent models from being nested to other valid models.

src/apps/schema/src/app/components/SelectModelParentInput.tsx
```diff
@@ -0,0 +1,74 @@
...
+export const SelectModelParentInput = ({
+  value,
+  onChange,
+  modelType,
+}: SelectModelParentInputProps) => {
+  const { data: navItems } = useGetContentNavItemsQuery();
+
+  const parents = useMemo(() => {
+    if (navItems) {
+      const _navItems = cloneDeep(navItems);
```
shrunyan, line 29: Why clone? This seems to only operate on the array and not object elements. Would creating a new array be more efficient?

src/apps/schema/src/app/components/CreateModelDialogue.tsx
shrunyan: okay. maybe a detail worth of a comment so we remember the expectation/behavior combo when we come back to this code

shrunyan (approved):

PR #2235, opened 2023-08-18
Fix: Unable to re-order child tree

theofficialnar:
Removed `pointer-events: none` which prevents the element onClick

[reorder-fix.webm](https://github.com/zesty-io/manager-ui/assets/28705606/41c57092-7d3f-4506-86e7-a269c3e46ab1)


theofficialnar:
> @theofficialnar Was there a reason this was added? Was it to avoid something else?

I was trying several things when I was fixing the bug with the reorder randomly jumping around and this was one of the things I remember adding and forgot to remove once I found the actual fix.

PR #2238, opened 2023-08-22
Feature/content title bar

agalin920:
Dependent on https://github.com/zesty-io/material/pull/79


zcolah (changes requested):
1. Unsaved Changes modal is not loading the new one which is [shown in Figma](https://www.figma.com/file/zOI7oSH3fG1XDmmPzsP6PX/App-Navigation?type=design&node-id=410%3A162105&mode=design&t=LNoc2n1skf1B4bch-1) but the old one. 

Please look into it. 

CleanShot 2023-08-23 at 03 16 11@2x

2. As per definition, the breadcrumb in this case should not show the more button. It should directly display categories. [See Figma](https://www.figma.com/file/zOI7oSH3fG1XDmmPzsP6PX/App-Navigation?type=design&node-id=410%3A159409&mode=design&t=LNoc2n1skf1B4bch-1). 

Note: We only show the more button when there are more than 4 parents. 

CleanShot 2023-08-23 at 03 17 40@2x

3.

zcolah:
6. The Order of Modals in the Duplicate Content Item flow does not match spec defined in [Figma](https://www.figma.com/file/zOI7oSH3fG1XDmmPzsP6PX/App-Navigation?type=design&node-id=410%3A164477&mode=design&t=Hn2cw0dF8PkTKJwk-1). Ideally the user should first get the warning, and then be asked if the want to duplicate. This is to avoid dirty state. 

Also the modals should not overlay each other but rather replace each other.

zcolah (changes requested):
6. The Order of Modals in the Duplicate Content Item flow does not match spec defined in [Figma](https://www.figma.com/file/zOI7oSH3fG1XDmmPzsP6PX/App-Navigation?type=design&node-id=410%3A164477&mode=design&t=Hn2cw0dF8PkTKJwk-1). Ideally the user should first get the warning, and then be asked if the want to duplicate. This is to avoid dirty state.
Also the modals should not overlay each other but rather replace each other.

zcolah (changes requested):
VQA Notes: 
https://www.figma.com/file/zOI7oSH3fG1XDmmPzsP6PX/App-Navigation?type=design&node-id=1519%3A206389&mode=design&t=Hn2cw0dF8PkTKJwk-1

zcolah (changes requested):
1. Publish Status padding on left and right should be 32px 
2. Was looking at the Analytics app and after we applied the white background and 32 px padding and the metrics seem to no more be center aligned to the donut chart in the metric bar area. 

CleanShot 2023-08-24 at 11 52 59@2x

zcolah:
Hi Andres, we need to build a small screen version of the title bar that wraps as it will help users at Blue Pixel. See [Figma design here](https://www.figma.com/file/zOI7oSH3fG1XDmmPzsP6PX/App-Navigation?type=design&node-id=1725%3A103896&mode=design&t=apsJ4YeoifmFcihw-1).

zcolah (changes requested):
Hi Andres, we need to build a small screen version of the title bar that wraps as it will help users at Blue Pixel. See [Figma design here](https://www.figma.com/file/zOI7oSH3fG1XDmmPzsP6PX/App-Navigation?type=design&node-id=1725%3A103896&mode=design&t=apsJ4YeoifmFcihw-1).

shrunyan:
Manage publish state link does not work.
image

PR #2244, opened 2023-09-05
Fix: App crashing when creating single page model with no parent

### LINKED ISSUE #2243: Schema: App crashes when creating a single page model with no parent model
**Describe the bug**
The app crashes when a user creates a new single page model with no parent model but the new model still gets created when the user refreshes the page.

**To Reproduce**
1. Go to Schema
2. Click on Create Model
3. Select single page model
4. Add in the new model details WITHOUT selecting a parent model
5. See error

**Expected behavior**
The app should not crash when a new model without a parent is created

**Screenshots**
[schema-create-bug.webm](https://github.com/zesty-io/manager-ui/assets/28705606/42a52401-6fda-42f5-ab6f-23ba1e04ab2c)

**Additional context**
The new model still gets created when the user refreshes the page.

theofficialnar:
Closes #2243


shrunyan (approved):

shrunyan:
Are we concerned at all on why the `model.parentZuid` is not present?

PR #2252, opened 2023-09-11
Change sales image

agalin920:


shrunyan (changes requested):
Please include width/height parameters to match the needed size. e.g. `https://zestyio.media.zestyio.com/gisele-blair-zestyio.jpeg?width=50&height=50`

shrunyan (approved):

PR #2288, opened 2023-10-05
Fix: Update cypress version

theofficialnar:
Updating Cypress to version 12 because of an issue with Chrome 117 where Cypress tests are failing.
For reference: https://github.com/cypress-io/cypress/issues/22163#issuecomment-1725418438


theofficialnar:
> @theofficialnar Wouldn't we expect this to pass?

Yeah, I'm actually still looking into why that specific test is failing since it passes on local.

PR #2290, opened 2023-10-05
Feature/content revamp

### LINKED ISSUE #2125: Content: Edit Content Tab Revamp

agalin920:
Dependent on https://github.com/zesty-io/material/pull/81


zcolah (changes requested):
1. Scrolling does not work as expected. Please see Loom for more context.
https://www.loom.com/share/800b54be52ac4fffb73a2afc76ee85fc

2. Some more scroll issues
https://www.loom.com/share/f2dec3edf1144cecb8d128ba5e8a136d

3. As I use DUO mode more, I realize scroll can be frustrating inside of it as well. Similar issues as above. 
https://www.loom.com/share/f4be95c05776413089a7c568828a33ab?from_recorder=1&focus_title=1

zcolah (changes requested):
1. Field Container Width is Off for Dataset Items: 
https://www.loom.com/share/f0620904e8964d80ae7003a1d181d62c

2. New Field Design and shell is not being automatically reflected in SEO and Head Tags tab
https://www.loom.com/share/7c5e5ccc6ecc4048b7236db2753c6cc3

3. Please make the current device selection be in active state in the menu options always: 
https://www.loom.com/share/d515e96003ba46dc972fb6aa281e4273

4. Old Loading screen is still appearing in DUO mode along with the new loading screen: https://www.loom.com/share/c19770b57ec24e34ab4a49e37152d32d?from_recorder=1&focus_title=1

5. Duo Mode Container is much larger than it should be: 
https://www.loom.com/share/c4591f7f761f4abc9c4c908ab6f79fb4

zcolah (changes requested):
[Watch Loom for More Context ](https://www.loom.com/share/b26d26e8108d4bcbac061d273d2c6ebc)
1. [Copy ZUID and Copy Field ID is not working (no tick mark and no copying happening)](https://www.figma.com/file/zOI7oSH3fG1XDmmPzsP6PX/Zesty-3.0-Revamp?type=design&node-id=1482%3A44574&mode=dev&t=7v8vl3ET5xhWX3fV-1)
2. Add the DUO Mode Toggle back to the title bar now. This is so that the customers have a more familiar experience to access DUO mode still available to them.
3. A Red Border must appear on every field that has an error. This includes WYSIWYG, Media, etc.
4. When DUO Mode is opened in a new tab, please have the scale set to 100%.
5. The link color for JSON does not match the link color of the other links. 
6. The padding on left and right of the main container is 24px. It should be 32px.
7. Return the Delete Section to the Right Sidebar. This is for familiarity.
8. The DUO Mode link is not stretching and filling the entire DUO Mode tool bar when we are on a wide screen
9. Errors are being shown in descending order of the field order. They should be in ascending order.
10. View Model Activity Log should open Activity Log in Schema

zcolah (changes requested):
May you please look into the Menu Item Height inconsistencies between production and stage? 
The Menu Item Height in our Figma Design System and our current Manager UI is 36px.
However in Stage the Menu Item height is 40px. Because this is 40px this has distorted multiple menus with the options being cropped out and looking larger than normal. I created an issue for this but then realized it cannot be an issue as it is only in stage atm. https://github.com/zesty-io/manager-ui/issues/2302
Here is a Figma screen from the past that shows that our menu item heights have always been 36px in Figma: https://www.figma.com/file/zOI7oSH3fG1XDmmPzsP6PX/Zesty-3.0-Revamp?type=design&node-id=2894%3A286049&mode=dev&t=n5MwysEP0xrSo3QF-1
Could we look into the old menu item height component and the new one? I am not sure what is happening. Not sure how they are becoming 40px.

theofficialnar:
@giseleblair @b-estevez empty fields **except** uuid and sort are now null when creating new content items. UUID has its value automatically generated and sort is always 0 by default.

b-estevez:
> @giseleblair @b-estevez empty fields **except** uuid and sort will be null. UUID has its value automatically generated and sort is always 0 by default.

@theofficialnar @agalin920 @giseleblair 

Upon testing all fields with empty values, it seems all fields have `null` values as expected except for `uuid` and `sort`

Screenshot from Chrome's developer tools:

image

image

Screenshot from Postman:

image


I also created a scenario:

* I created a single content item while the `uuid` and `sort` fields doesn't exist yet from the schema.
* After the first content item was created, I added `uuid` and `sort` fields to the schema and then I created a new content item (meaning there are 2 total content items now).
* The second content item have default values for `uuid` and `sort` fields and the first existing content item has `null` values for both `uuid` and `sort` fields.

First Content Item
image

Second Content Item
image


Meaning the `uuid` and `sort` value will still have `null` as default value in some cases which is what we want.

shrunyan (approved):

PR #2305, opened 2023-10-25
Prevent infinite loop if a model's parent is itself

theofficialnar:
Context: https://zesty-io.slack.com/archives/C050120G5/p1698191024727889

Prevent the model breadcrumbs component from going on an infinite loop if a model's parent is itself.

### Preview
[breadcrumb-infinite-loop-fix.webm](https://github.com/zesty-io/manager-ui/assets/28705606/b67cf463-a210-4b0f-b13a-407337a676c9)


shrunyan:
@theofficialnar is it possible to include a test which validates the prior issue and the fix for it?

theofficialnar:
@agalin920 filtered the options to make sure that a model cannot parent itself
@shrunyan added the test to check if a model that has previously parented itself can be rendered properly

PR #2309, opened 2023-10-26
Fix use-domain hook logic to not include any paths on domain

agalin920:
Also added edge case handling in the case a domain already has a protocol set and if the domain is an invalid URL


shrunyan:
I don't see how the logic provided solves the linked issue. Seems like the logic is accounting for protocol not a domain having a path part.

shrunyan (changes requested):
I don't see how the logic provided solves the linked issue. Seems like the logic is accounting for protocol not a domain having a path part.

shrunyan:
> I don't see how the logic provided solves the linked issue. Seems like the logic is accounting for protocol not a domain having a path part.

Nevermind. Got it now. Took a second read.

shrunyan (approved):

PR #2444, opened 2024-01-09
Remove content item data keys that are no longer valid when item is being created

agalin920:
See ticket for better context


shrunyan:
@agalin920 how does this interact with prior versions which contain the prior field name?

shrunyan (approved):

PR #2446, opened 2024-01-10
Limit media upload on content edit image field to locked group

agalin920:
Screenshot 2024-01-10 at 11 55 28 AM


shrunyan:
@agalin920 This seems to have broken a test

shrunyan (approved):

Issue #2502, opened 2024-01-30
Content: SEO Tab - Unclear if save is successful when I try to save with Meta Title of more than 160 characters


zcolah:
**Describe the bug**
1. I open a content item and go to the SEO Tab
2. In the SEO tab, I enter a Meta Title that is more than 160 characters (basically beyond it's limit)
3. I click Save 
4. I see an error that says cannot save, but the title bar still indicates that the item is saved. 
5. This is a confusing experience for me

See video: https://www.loom.com/share/392958068935400bb6e47b659f272ce0

**Solution**
1. Show the red border around the field that has the error
2. If the Meta Title has more than 160 characters, and tries to save the item: trims it text content to 160 characters and then save it successfully. This is similar to the Meta Description experience.

PR #2534, opened 2024-02-05
fix: Content: SEO Meta Description Displays that there are excess characters when there are none

### LINKED ISSUE #2501: Content: SEO Meta Description Displays that there are excess characters when there are none
**Bug:**
SEO Meta Description Displays that there are excess characters when there are none
See Loom: https://www.loom.com/share/33c13fa684b94f3a88111e9c7bfad2f9

**Solution**
Do not display that there are excess characters unless there actually are excess characters

glespinosa:
This will also solve problem in Meta Title, Meta Keywords, Navigation Link Text not only in the Meta Descriptions because they share same issues regarding excess characters

[screencast-8-aaeffee09b-7w6v22.manager.dev.zesty.io_8080-2024.02.06-09_19_45.webm](https://github.com/zesty-io/manager-ui/assets/44116036/828dea4e-4a74-4d69-a4eb-39e715f7257c)

Closes #2501 #2502

### DIFF
--- a/src/apps/content-editor/src/app/views/ItemEdit/Meta/ItemSettings/settings/MetaDescription.js
+++ b/src/apps/content-editor/src/app/views/ItemEdit/Meta/ItemSettings/settings/MetaDescription.js
@@ -46,7 +46,7 @@ export default connect()(function MetaDescription({
         onChange(meta_description, "metaDescription");
       }
     }
-  }, [meta_description]);
+  }, [meta_description, isSaving]);
 
   if (error) {
     dispatch(
--- a/src/apps/content-editor/src/app/views/ItemEdit/Meta/ItemSettings/settings/MetaDescription.js
+++ b/src/apps/content-editor/src/app/views/ItemEdit/Meta/ItemSettings/settings/MetaDescription.js
@@ -12,7 +12,6 @@ export default connect()(function MetaDescription({
   onChange,
   dispatch,
   errors,
-  isSaving,
 }) {
   const [error, setError] = useState("");
 
@@ -41,12 +40,8 @@ export default connect()(function MetaDescription({
       }
 
       setError(message);
-
-      if (isSaving) {
-        onChange(meta_description, "metaDescription");
-      }
     }
-  }, [meta_description, isSaving]);
+  }, [meta_description]);
 
   if (error) {
     dispatch(
--- a/src/apps/content-editor/src/app/views/ItemEdit/Meta/ItemSettings/settings/MetaDescription.js
+++ b/src/apps/content-editor/src/app/views/ItemEdit/Meta/ItemSettings/settings/MetaDescription.js
@@ -12,6 +12,7 @@ export default connect()(function MetaDescription({
   onChange,
   dispatch,
   errors,
+  isSaving,
 }) {
   const [error, setError] = useState("");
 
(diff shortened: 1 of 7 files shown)


theofficialnar (commented):
@glespinosa did you check with @zcolah if the expected behavior here is to automatically trim the user input instead of retaining the excess characters and letting the user fix it themselves before it gets saved?

zcolah:
> @glespinosa did you check with @zcolah if the expected behavior here is to automatically trim the user input instead of retaining the excess characters and letting the user fix it themselves before it gets saved?

@theofficialnar currently this is the approach we are taking as it is the behavior that was already being followed for the other fields on the SEO Tab. We definitely want an error blurb like experience for SEO in the near future though.

theofficialnar (approved):

src/apps/content-editor/src/app/views/ItemEdit/Meta/ItemSettings/settings/MetaDescription.js
```diff
@@ -40,6 +41,10 @@ export default connect()(function MetaDescription({
       }
 
       setError(message);
+
+      if (isSaving) {
```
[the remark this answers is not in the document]
glespinosa: yes for now it's relying on the useEffect since everytime the user is saving that's the only time I can rerun the onChange function to check the characters condition on excess.

src/shell/store/content.js
```diff
@@ -406,7 +406,15 @@ export function saveItem(itemZUID, action = "") {
...
     }
-
+    if (item.web.metaTitle) {
+      item.web.metaTitle = item.web.metaTitle.slice(0, 150);
+    }
+    if (item.web.metaLinkText) {
+      item.web.metaLinkText = item.web.metaLinkText.slice(0, 150);
+    }
+    if (item.web.metaKeywords) {
+      item.web.metaKeywords = item.web.metaKeywords.slice(0, 250);
```
theofficialnar, line 416: Why is 250 here? The ui character counter says it's 255.
image

glespinosa: done updating changes

src/apps/content-editor/src/app/views/ItemEdit/Meta/ItemSettings/settings/MetaDescription.js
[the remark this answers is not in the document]
glespinosa: @agalin920 I removed now the useEffect on every component and let the useEffect inside of ItemSettings do the work of removing the errors while on there are on saving... Then i slice all the props base on the number of allowed character on the properties in the /shell/store/content.js

src/shell/store/content.js
```diff
@@ -406,7 +406,15 @@ export function saveItem(itemZUID, action = "") {
     if (item.web.metaDescription) {
       item.web.metaDescription = item.web.metaDescription.slice(0, 160);
     }
-
+    if (item.web.metaTitle) {
+      item.web.metaTitle = item.web.metaTitle.slice(0, 150);
```
shrunyan, line 410: why is this being sliced at 150 characters?

src/shell/store/content.js
shrunyan: @glespinosa which character count did we settle on and why?

src/shell/store/content.js
glespinosa: My basis is on their UI like how the metadescription and other fields were slice.
image

src/shell/store/content.js
glespinosa: @shrunyan  I'm basing on the UI
image

PR #2552, opened 2024-02-15
Sort field revamp

### LINKED ISSUE #2458: Content: Sort Field
# Problem
The current sort field has the following key issues: 

1. The Plus and Minus buttons are filled primary action buttons that grab too much attention. These are tertiary action buttons and hence do not need so much visual prominence. 
2. They allow for a user to type in "."s for decimal places despite not supporting them. 

CleanShot 2024-01-16 at 18 47 24

# Solution
The new sort field has muted plus and minus buttons and does not the support the input of "."s .
image

theofficialnar:
image

Closes #2458

### DIFF
--- a/cypress/e2e/content/content.spec.js
+++ b/cypress/e2e/content/content.spec.js
@@ -238,23 +238,17 @@ describe("Content Specs", () => {
     });
 
     it("Sort Field", () => {
-      cy.get("#12-4e1914-kcqznz input[type='number']")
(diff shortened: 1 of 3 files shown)


zcolah:
- [ ] Unable to type in a minus sign into the input
- [ ] The input should grow as the user enters more than 2 digit characters. Note: it should not exceed the bounds of the container
- [ ] Center Align the text
- [ ] User should be able to use the up and down arrow keys on the keyboard to increase and decrease the value in the input when selected

PR #2558, opened 2024-02-19
fix: Global Nav: Bug Fixes

glespinosa:
image


glespinosa:
The cause of the scrollbar is because of the paddingY of 20px, changing to 16px solve the remaining issues
image

theofficialnar (approved):

PR #2579, opened 2024-03-07
Allow spaces and capital letters in dropdown value

### LINKED ISSUE #1343: Schema App: Dropdown field type does not allow for spaces in value
When the user selects the Dropdown field type in schema and they are configuring the values for each dropdown option, the UI does not allow the user to have spaces or capital letters in the value field. 

The API does allow for capital letters and spaces and so the UI should also allow for the same formatting options. 

Screen Shot 2022-09-08 at 2 35 56 PM

theofficialnar:
[Screencast_20240307_102024.webm](https://github.com/zesty-io/manager-ui/assets/28705606/0c60724d-56bd-43ce-b381-d0534a56297d)

Fixes #1343


giseleblair:
Looks good thank you!

Issue #2435, opened 2024-01-09
Media: Preview Modal - Disable the Arrow Key Image Changing Functionality when a user is typing in the title input and rename modal input


zcolah:
**Describe the bug**
Preview Modal - Disable the Arrow Key Image Changing Functionality when a user is typing in the title input. 

https://www.loom.com/share/38192d8f864848859d5b71f4d524014a

This issue also happens when the user is renaming a file. See Loom. 

https://www.loom.com/share/6fc3a420ba504bd79a6ceb2827d7bd11

PR #2584, opened 2024-03-12
fix: Preview Modal - Disable the Arrow Key Image Changing Functionality

glespinosa:
Will close https://github.com/zesty-io/manager-ui/issues/2435
This will prevent unnecessary navigation of images while the user is focus on an input/textarea when the only purpose of user is to navigate through texts.


glespinosa:
@zcolah needs your vqa tag here on the PR

PR #2668, opened 2024-04-10
Fix/Conditionally call function that is an optional prop

agalin920:


src/apps/media/src/app/components/FileModal/index.tsx
```diff
@@ -70,7 +70,7 @@ export const FileModal: FC<Props> = ({
 
   useEffect(() => {
     if (isError) {
-      onSetIsFileModalError(true);
+      onSetIsFileModalError && onSetIsFileModalError(true);
```
shrunyan, line 73: Total nit pick, feel free to disregard if you disagree. As both `isError` and `onSetIsFileModalError` having values set is a condition for the function being invoked I would rather see `onSetIsFileModalError` as part of the `if` conditional statement.

theofficialnar (approved):

PR #2677, opened 2024-04-15
Change internal link none value from "0" to null

### LINKED ISSUE #2667: Internal Link field: when "None" is selected the value should be null
When an "internal link" field type is configured to a model, the initial value of this field is `null`. When a user selects  `None` from the dropdown in the UI, this value is then returned as "0" as opposed to `null` 

If a user selects `None` then the value returned should be `null` to keep the consistency of when a user selects or does not select a value from this field type. 

Screenshot 2024-04-08 at 11 56 40 AM

Also note in the screenshot, that when a user does explicitly select `None` from the dropdown, then the dropdown will display an item that says `Selected item not found: 0`

finnar-bin:
Fixes #2667 

image
image


shrunyan:
@agalin920 add the review request as I'm trying to determine why we are not seeing the auto-merge option. Not sure if once all conditions are meet it is no longer available and must be added before

PR #2693, opened 2024-04-23
Add media app and listeners on Marketplace and Content Freestyle apps

agalin920:
https://github.com/zesty-io/manager-ui/assets/10054410/41ce6b94-a10b-4001-b8d4-122bf6027bfa

### DIFF
--- a/src/apps/content-editor/src/app/views/ItemEdit/components/ItemEditHeader/DuplicateItemDialog.tsx
+++ b/src/apps/content-editor/src/app/views/ItemEdit/components/ItemEditHeader/DuplicateItemDialog.tsx
@@ -40,11 +40,22 @@ export const DuplicateItemDialog = ({ onClose }: DuplicateItemProps) => {
     if (modelFields?.[0]?.datatype === "text") {
       fieldToChange = modelFields[0].name;
     }
+
+    const uuidFieldKeys = modelFields
+      ?.filter((field) => field.datatype === "uuid")
+      ?.map((field) => field.name);
+
+    const newData = { ...item.data };
+
+    uuidFieldKeys?.forEach((key) => {
+      newData[key] = null; // Set each uuid key to null
+    });
+
     createContentItem({
       modelZUID,
       body: {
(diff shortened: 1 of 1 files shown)


src/apps/content-editor/src/app/views/ItemEdit/FreestyleWrapper.tsx
```diff
@@ -15,6 +19,21 @@ export const FreestyleWrapper = () => {
   // @ts-expect-error CONFIG not typed
   const [sessionToken] = useState(Cookies.get(CONFIG.COOKIE_NAME));
 
+  const [showZestyDAM, setShowZestyDAM] = useState(false);
+
+  const handleZestyDAMRequest = (event: MessageEvent) => {
+    if (event.data.type === "ZESTY_DAM_REQUEST") {
```
shrunyan, line 25: As I am looking at this I have a thought on how we can further secure our windowing messages. In the vein of what has been discussed in the past with a need to make a robust messaging system as we further expand on embedding external applications. When manager-ui mounts an application it hands that application the current users session token. Additional if a standalone application is running it assumedly used the app loader and obtained a zesty user session token. Either way there should be an active user session token present within both the parent and child application.

If the child application was to provide the session token back to the parent with every window message the parent could then check for a string match between the parents session token and child provided session token. This would confirm that the user session being used between the parent and child is the same.

Thoughts?

src/apps/marketplace/src/app/view/CustomApp.js
```diff
@@ -54,6 +59,19 @@ function LoadApp(props) {
     }
   }, [frame.current, app]);
 
+  const handleZestyDAMRequest = (event) => {
```
shrunyan, line 62: the same state, handler, useeffect and router wrapping the media component exists within the `CustomApp` and `FreestyleWrapper` components. Would it make sense to extract these out into a High order Component. Which could then wrap any other component.

shrunyan (approved):

PR #2704, opened 2024-05-08
Avoid UUID duplication

agalin920:


src/apps/content-editor/src/app/views/ItemEdit/components/ItemEditHeader/DuplicateItemDialog.tsx
```diff
@@ -45,6 +45,8 @@ export const DuplicateItemDialog = ({ onClose }: DuplicateItemProps) => {
       body: {
         data: {
           ...item.data,
+          // Remove the UUID to avoid duplication
+          ...(item.data.uuid !== undefined && { uuid: null }),
```
shrunyan, line 49: The property name, in this case `uuid`, on the data object is a determined by the user provided field name. As such; there is no guarantee the field will be called "uuid". We need to find fields with the type of uuid and `null` their values.

shrunyan (changes requested):

finnar-bin (approved):

PR #2711, opened 2024-05-13
Comment System

finnar-bin:
**This is dependent on BE Change: [insert pr here]**

https://github.com/zesty-io/manager-ui/assets/28705606/80fd0937-9c61-427d-b1da-7e3f34263776


finnar-bin:
### Note: need to make some payload changes. Not yet ready for merging.

src/apps/content-editor/src/app/views/ItemEdit/ItemEdit.js
```diff
@@ -443,7 +443,11 @@ export default function ItemEdit() {
                 />
                 <Route
                   exact
-                  path="/content/:modelZUID/:itemZUID"
+                  path={[
```
shrunyan, line 446: @agalin920 please weigh in on the URL design for deep links to comments.

finnar-bin: @agalin920 this was originally done via query params. @shrunyan suggested that there are other options to achieving this and I thought that using the paths would make the deeplink more consistent with our existing route patterns.

shrunyan: @agalin920 I'm good with using query parameters but would want a clear and written explanation of when to use paths vs parameters. 

My thinking is the following. _Paths are used to deep link to a view(read: a combination of UI layout and elements rendered on screen) and parameters are used to refine a view_. 

**e.g.** 
Table view: /content/6-000-0000
Table view filtered to published items: /content/6-000-0000?status=published

Content edit view: /content/6-000-0000/7-000-0000
Content edit view that displays deactivated field:  /content/6-000-0000/7-000-0000?showDeactivated=true

Comment in content edit view: /content/6-000-0000/7-000-0000/comment/12-000-0000/24-000-0000
Comment in content edit view that displays deleted replies:  /content/6-000-0000/7-000-0000/comment/12-000-0000/24-000-0000?showDeleted=true

PR #2800, opened 2024-07-11
Update marketplace json endpoint

agalin920:


finnar-bin (approved):
Are we planning on utilizing this url moving forward or is this going to be a temp fix?

finnar-bin:
Gotcha, I thought this a dev preview link. Was concerned that unpublished announcements would be accidentally displayed.

shrunyan:
I concur with @finnar-bin. I do not think this URL is what we want as it is a preview link and will ignore publishings data

shrunyan:
I believe we want close this PR in favor of this API change https://github.com/zesty-io/webengine/pull/388

@agalin920 let me know if you disagree or if I'm incorrect with the linked PR

PR #2849, opened 2024-07-18
[Media] File replacement

### LINKED ISSUE #2661: Media: Allow a User to Replace File
# Problem
Key Customer Request: A user should be able to replace a file and keep the same same ZUID, same file name, and same URL. This is very important for critical files being replaced. 

# Solution

A user should be able to select a file, and then click on a replace file and then select a file from their computer they would like to replace it with

image

image

## Add Tooltips for Media Preview
- Tooltip should be bottom positioned start: https://mui.com/material-ui/react-tooltip/#positioned-tooltips

- More Icon: "More"
- Edit Icon: "Rename File"
- Replace Icon: "Replace File"
- Delete Icon: "Delete File"
- Close Icon: "Close Preview"

image

# Figma
https://www.figma.com/file/tnYqXziKnSW0XQcXdHD2zP/Media%3A-Replace-File-Feature?type=design&node-id=0%3A1&mode=dev&t=IOHLnDJxoDiGQbd5-1

# Developer Notes
1. When replacing a file, ensure it keeps the same ZUID, same file name, and same URL. 
2. Allow a user to update the alt text of a file when going through the replace flow. They should not be able to update the file name though.
3. Linking API request here: https://github.com/zesty-io/manager-ui/issues/136

# Future Exploration (Out of Scope in this ticket)
1. A user should be able to add an item to Media  via drag and drop or the upload button. If it has the same name as a file that exists, then we should ask if we want to replace it or keep both. Similar to Google Drive. 

CleanShot 2024-04-12 at 18 53 44

finnar-bin:
Allows users to replace an existing file with a new file of the same file type without changing the file's name and url in our DAM.
Resolves #2661 

### Dependencies
- Backend Change: https://github.com/zesty-io/media-manager-service/pull/150
- New icon: https://github.com/zesty-io/material/pull/105

### Demo

https://github.com/user-attachments/assets/5e68c830-e512-42b9-b6e1-15f6405d5671


finnar-bin:
@agalin920 I have moved the NoPermission component out of the user modal folder so that it can be its own reusable component since it's pretty much the same to what I needed for the File replacement no permission.

PR #2867, opened 2024-07-22
Schema rules limit file type for media field

agalin920:
Screenshot 2024-07-22 at 2 07 05 PM


finnar-bin (dismissed):

finnar-bin:
Btw, I vaguely remember Bynder's  having a way to limit which file types will be available in the popup but I'm not sure if that's something that would be beneficial here, just wanted to point it out.

finnar-bin (approved):

PR #3030, opened 2024-10-22
[Content] Enable ai assistant for all instances

finnar-bin:
Context: https://zesty-io.slack.com/archives/C04U75EEV/p1729628914170349?thread_ts=1729562214.091069&cid=C04U75EEV

### DIFF
--- a/src/shell/components/withAi/index.tsx
+++ b/src/shell/components/withAi/index.tsx
@@ -111,114 +111,110 @@ export const withAI = (WrappedComponent: ComponentType) =>
       }
     };
 
-    if (isEnabled) {
-      return (
-        <>
-          <WrappedComponent
-            {...props}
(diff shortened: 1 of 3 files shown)


src/shell/components/withAi/index.tsx
```diff
@@ -111,114 +111,110 @@ export const withAI = (WrappedComponent: ComponentType) =>
       }
     };
 
-    if (isEnabled) {
```
agalin920, line 114: Do we still need this variable?

finnar-bin: I left it in case we decide to lock this again to certain instances so we can just re-add the variable without having to rewrite the logic again. But I can remove it if it's preferred.

agalin920 (approved):

PR #3036, opened 2024-10-25
fix: have workflows create necessary branch and clean up prior branches

shrunyan:
- add job to remove prior env branch
- add job on stable workflow to tag release and push to origin. Tag will follow the format `release-GIT_SHORT_HASH`
- update job to ensure branch is present at origin before checking/creating PR
- rename jobs for consistency

### DIFF
--- a/.github/workflows/cd-beta.yaml
+++ b/.github/workflows/cd-beta.yaml
@@ -1,5 +1,9 @@
 name: cd-beta
 
+concurrency:
+  group: beta
+  cancel-in-progress: true
+
 on:
   push:
     branches:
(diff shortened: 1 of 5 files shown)


.github/workflows/cd-beta.yaml
```diff
@@ -77,12 +81,25 @@ jobs:
...
     runs-on: ubuntu-latest
-    # needs: [deploy_beta]
     steps:
       - name: Checkout Repo
-        uses: actions/checkout@v2
+        uses: actions/checkout@v4
+
+      - name: Ensure origin/stable branch exists
+        if: git branch -r --list origin/stable != 'origin/stable'
+        run: it checkout beta && git switch --create stable && git push -u origin stable
```
finnar-bin: Not sure if this is how it's supposed to be but it seems like it's missing a `g` in the git checkout

shrunyan: Good catch! Thanks @finnar-bin

PR #3044, opened 2024-10-28
[Content] Support dedicated TC Image field in SEO tab

### LINKED ISSUE #3021: Content: SEO Tab - TC Image Field
# Problem
Currently a user cannot see the TC Image field in the SEO tab and instead they see it in the Edit tab. This confuses and frustrates the user for two key reasons: 

1. All other TC meta fields are present in the SEO such as TC Title and TC Description. They need to do continuous back and forth to update just the TC image field.
2. The TC image field value is not reflected in the Twitter Preview Card. 

# Solution

1. If the model of a content item has a TC Image field, then please display this field in the SEO Tab instead of the Edit Tab for the Content item. 
2. For the Twitter Card social media preview image, please always display the file in the TC Image field. However if the TC image is blank or is not of an image file type, then please use the Meta Image field file. 

# Figma
https://www.figma.com/design/Oc42SSkzPDozxO7zZRCUJT/SEO-Tab-%2B-AI-App?node-id=668-43036&m=dev&t=q6KEMTga12olZ0Qr-1

finnar-bin:
Resolves #3021 

### Preview
image

### DIFF
--- a/src/apps/content-editor/src/app/views/ItemEdit/Meta/SocialMediaPreview/TwitterPreview.tsx
+++ b/src/apps/content-editor/src/app/views/ItemEdit/Meta/SocialMediaPreview/TwitterPreview.tsx
@@ -6,8 +6,6 @@ import { useSelector } from "react-redux";
 
 import { useDomain } from "../../../../../../../../shell/hooks/use-domain";
 import { AppState } from "../../../../../../../../shell/store/types";
-import { useLazyGetFileQuery } from "../../../../../../../../shell/services/mediaManager";
-import { fileExtension } from "../../../../../../../media/src/app/utils/fileUtils";
 
 type TwitterPreviewProps = {
   imageURL: string;
@@ -19,7 +17,6 @@ export const TwitterPreview = ({ imageURL }: TwitterPreviewProps) => {
   }>();
   const domain = useDomain();
   const location = useLocation();
-  const [getFile] = useLazyGetFileQuery();
   const [tcImageURL, setTcImageURL] = useState<string>(null);
   const isCreateItemPage = location?.pathname?.split("/")?.pop() === "new";
   const item = useSelector(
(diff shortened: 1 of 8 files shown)


src/apps/content-editor/src/app/views/ItemEdit/Meta/SocialMediaPreview/TwitterPreview.tsx
```diff
@@ -17,12 +18,38 @@ export const TwitterPreview = ({ imageURL }: TwitterPreviewProps) => {
...
   );
 
+  useEffect(() => {
+    if (!!item?.data?.tc_image) {
+      const tcImage = String(item?.data?.tc_image)
+        ?.split(",")
+        ?.filter((el: string) => el)?.[0];
+
+      if (tcImage.startsWith("3-")) {
+        getFile(String(item.data.tc_image), true)
```
agalin920: Cant we use the resolver here?

agalin920 (changes requested):

src/apps/content-editor/src/app/views/ItemEdit/Meta/SocialMediaPreview/TwitterPreview.tsx
finnar-bin: Oh yeah, I needed to check if the zuid actually contains an image first, if not use the existing meta image. This reminds me to actually add that check which I forgot, I'll be making another commit to add it.

finnar-bin: Updated to add the check I mentioned above

agalin920: Hmm why not just use html image onError callback? that would suffice to tell if its not an image right?

agalin920 (changes requested):

finnar-bin:
> Hmm why not just use html image fallback?

That something I actually haven't thought off, lemme check that out

finnar-bin:
@agalin920 updated to use the image resolver

agalin920 (approved):

PR #3061, opened 2024-11-18
CSV Download

### LINKED ISSUE #2675: Content: Export CSV
# Intro 

We are adding an "Export CSV" button to the multi page table title bar. When this button is clicked it should allow a user to download a CSV file that contains all the field columns and rows. In addition to the field columns we will also add the following columns: 

1. Created At (Date Time)
2. Created By 
3. Updated At (Date Time)
4. Content Item ZUID

# Questions to Investigate 

1. Describe a situation of why and where the export csv feature will be used? How will this CSV data get consumed by them after downloading it? Are they archiving it? Are they trying to duplicate a model and also have the items carry over with it? Are they trying to transfer data between two instances? 

2. How does the user expect to see: 
- One to Many, One to One, 
- Internal Link relationships
- Do they want all the columns of the related fields?
- Media Field (are they expecting the file name, URL, thumbnail, or a ZUID)

3. How does the user expect the media field 
4. Are they comfortable with ZUIDs? Do they wish to see the display labels?
5. Where would they expect to see this feature? Schema or Content? 

# Figma 
https://www.figma.com/file/zOI7oSH3fG1XDmmPzsP6PX/Zesty-3.0-Revamp?type=design&node-id=4288%3A191852&mode=dev&t=qXhv5y0NQio0iSqe-1

allenpigar:
Should be deploy the gcp-cf first before deploying this pr : https://github.com/zesty-io/gcp-cf/pull/179

added `export csv` function on content item.
csv-download

### DIFF
--- a/src/shell/services/cloudFunctions.ts
+++ b/src/shell/services/cloudFunctions.ts
@@ -23,10 +23,37 @@ export const cloudFunctionsApi = createApi({
         };
       },
     }),
+
+    downloadCsv: builder.query<
+      string[],
+      {
+        modelZUID: string;
+      }
+    >({
+      query: ({ modelZUID }) => {
+        return {
+          url: `downloadCSV`,
+          method: "GET",
+          params: {
+            instanceZUID,
+            modelZUID,
+          },
+          responseHandler: async (response) =>
+            window.location.assign(
+              window.URL.createObjectURL(await response.blob())
+            ),
+          cache: "no-cache",
+        };
+      },
+    }),
   }),
 });
 
 // Export hooks for usage in functional components, which are
 // auto-generated based on the defined endpoints
-export const { useRefreshCacheMutation, useAiGenerationMutation } =
-  cloudFunctionsApi;
+export const {
+  useRefreshCacheMutation,
+  useAiGenerationMutation,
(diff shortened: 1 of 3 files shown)


finnar-bin (approved):

src/shell/services/cloudFunctions.ts
```diff
@@ -23,10 +23,37 @@ export const cloudFunctionsApi = createApi({
...
     }),
+
+    downloadCsv: builder.query<
+      string[],
+      {
+        modelZUID: string;
+      }
+    >({
+      query: ({ modelZUID }) => {
+        return {
```
shrunyan, line 34: @allenpigar the users session is not being sent with this request causing it to fail.

allenpigar: user session is added using the `preparedHeaders` in utils:  https://github.com/zesty-io/manager-ui/pull/3061/files#diff-205e3a012094d2e6f027f951e9e9e609afaccb2cc2bbe29850f56c52d8e4db9dR2

i already made adjustment on the env variable in dev. seems like the env was not set properly on lower env which causes the invalid session error.

PR #3083, opened 2024-11-25
Do not log network errors to sentry

### LINKED ISSUE #3074: TypeError: Failed to fetch
Sentry Issue: [MANAGER-UI-1Z5](https://zestyio.sentry.io/issues/3893430801/?referrer=github_integration)

```
TypeError: Failed to fetch
  at request (./src/utility/request.js:52:10)
  at action (./src/shell/store/auth.js:89:12)
  at next (./src/shell/store/middleware/session.js:11:22)
  at dispatch (./src/shell/store/middleware/local-storage.js:6:18)
  at checkSession (./src/shell/components/private-route/index.js:19:15)
...
(2 additional frame(s) were not displayed)
```

finnar-bin:
Resolves #3074

### DIFF
--- a/src/utility/request.js
+++ b/src/utility/request.js
@@ -100,8 +100,8 @@ export function request(url, opts = {}) {
         );
         throw err;
       } else {
+        // Network errors
         store.dispatch(notify({ message: err.message, kind: "warn" }));
-        Sentry.captureException(err, { fingerprint: ["network_error"] });
         return Promise.reject(err);
       }
     });
--- a/src/utility/request.js
+++ b/src/utility/request.js
@@ -1,6 +1,5 @@
 import Cookies from "js-cookie";
 import { rest } from "lodash";
-import { Sentry } from "./sentry";
 import { store } from "shell/store";
 import { notify } from "../shell/store/notifications";
 // import { endSession } from "shell/store/auth";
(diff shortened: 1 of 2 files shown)


src/utility/request.js
```diff
@@ -100,8 +100,8 @@ export function request(url, opts = {}) {
         );
         throw err;
       } else {
+        // Network errors
         store.dispatch(notify({ message: err.message, kind: "warn" }));
-        Sentry.captureException(err, { fingerprint: ["network_error"] });
```
agalin920, line 104: This line is meant  to fingerprint the error in order to aggregate them. The error is still being thrown below this line.

An unhandled promise rejection is typically treated as an unhandled error by Sentry so this doesn't resolve what the PR aims to resolve.

All this change would do is de-aggregate them and make more noise

agalin920 (changes requested):

src/utility/request.js
finnar-bin: @agalin920 I see. I assume removing the throw error below this line would be enough? We're already dispatching a notification in the UI when this happens anyway.

agalin920: Right. I believe so

finnar-bin: Updated

agalin920 (approved):

PR #3086, opened 2024-12-02
[Content] Handle error when onOpen does not return a Promise

### LINKED ISSUE #3070: TypeError: Cannot read properties of undefined (reading 'then')
Sentry Issue: [MANAGER-UI-27W](https://zestyio.sentry.io/issues/5659024907/?referrer=github_integration)

```
TypeError: Cannot read properties of undefined (reading 'then')
  at <object>.onOpen (./src/shell/components/FieldTypeOneToOne/index.tsx:67:7)
...
(15 additional frame(s) were not displayed)
```

finnar-bin:
Fixes #3070


src/shell/components/FieldTypeOneToOne/index.tsx
```diff
@@ -64,10 +64,14 @@ export const FieldTypeOneToOne = ({
 
   const handleOpen = () => {
     if (!loaded && onOpen) {
-      onOpen().then(() => {
-        setLoading(false);
-      });
       setLoading(true);
+      if (onOpen() instanceof Promise) {
```
agalin920: Why/When would onOpen not be a promise?

agalin920 (changes requested):

src/shell/components/FieldTypeOneToOne/index.tsx
finnar-bin: I've updated my solution to this. So, what happens is that for the one to one field type, onOpen returns a dispatch action, but that dispatch action returns an empty object if the modelZUID is null/undefined. So, what I did was when onOpen is triggered, check if there's a modelZUID before running the dispatch, if none return a Promise.reject. This way onOpen will always be a promise.

src/apps/content-editor/src/app/components/Editor/Field/Field.tsx
```diff
@@ -703,11 +703,15 @@ export const Field = ({
 
     case "one_to_one":
       const onOneToOneOpen = useCallback(() => {
-        return dispatch(
-          fetchItems(relatedModelZUID, {
-            lang: getSelectedLang(allLanguages, langID),
-          })
-        );
+        if (!!relatedModelZUID) {
```
shrunyan: What are we confirming with this check? That it is not blank? It is not `0`? It not `null`? Would this be clear as `Boolean()`? Should this be using our https://github.com/zesty-io/zuid-js lib to validate the ZUID?

finnar-bin: Yeah, it's just checking if it has a value. I wasn't aware that we have a zuid validator I could make use, I guess it makes sense to be using that validator instead.

finnar-bin: @shrunyan Updated to use `zuid.isValid`

shrunyan (approved):

agalin920 (approved):

PR #3087, opened 2024-12-03
Feature/blocks

agalin920:


src/apps/blocks/index.tsx
```diff
@@ -0,0 +1,60 @@
...
+          id="blocksNav"
+          defaultWidth={220}
+          minWidth={220}
+          maxWidth={360}
+        >
+          <Sidebar />
+        </ResizableContainer>
+        <Switch>
+          <Route exact path="/blocks" render={() => <AllBlocks />} />
+          {/* <Route path="/schema/search" render={() => <SearchModels />} />
```
finnar-bin: Probably worth removing if unused

src/apps/blocks/views/AllBlocks.tsx
```diff
@@ -0,0 +1,230 @@
...
+                  <Box
+                    component="ul"
+                    color="info.dark"
+                    sx={{
+                      fontSize: "14px",
+                      listStylePosition: "inside",
+                    }}
+                  >
+                    <li>
+                      <Link color="secondary">How to Create a Block</Link>
```
finnar-bin, line 180: Are these links supposed to not have hrefs?

agalin920: Yea don't have them yet from Zosh

src/apps/blocks/index.tsx
agalin920: Will remove

src/apps/content-editor/src/app/components/Editor/Field/Field.tsx
```diff
@@ -283,10 +283,16 @@ export const Field = ({
       return (
         <AIFieldShell
           ZUID={fieldData?.ZUID}
-          name={fieldData?.name}
```
shrunyan, line 286: This change strike me as odd. How does this relate to blocks? How is the `name` variable available? If it was already available, why add it now as an or condition?

shrunyan: Same questions for label

src/apps/content-editor/src/app/components/Editor/PreviewMode/PreviewMode.js
```diff
@@ -33,11 +33,17 @@ export default function PreviewMode(props) {
       if (itemZUID) {
         const item = content[itemZUID];
 
-        let url = item?.web?.path
-          ? `${item.web.path}`
-          : `/-/instant/${item.meta.ZUID}.json`;
+        let url = "";
 
-        url = `${url}?_bypassError=true&__version=${props.version}`;
```
shrunyan, line 40: Does this need to be extracted? Are there other code paths that need item URL determination?

src/apps/content-editor/src/app/components/Editor/Field/Field.tsx
shrunyan: I'm seeing that this component is being used in the blocks app. I guess I'm not sure why there is a difference in the props.

src/apps/content-editor/src/app/views/ItemEdit/components/BlockTabs.tsx
```diff
@@ -0,0 +1,270 @@
...
+import { useHistory, useParams } from "react-router";
+import { useGetContentModelItemsQuery } from "../../../../../../../shell/services/instance";
+import { ContentItem } from "../../../../../../../shell/services/types";
+import moment from "moment-timezone";
+import { useGetUsersQuery } from "../../../../../../../shell/services/accounts";
+import { AddRounded, SearchRounded } from "@mui/icons-material";
+import noSearchResults from "../../../../../../../../public/images/noSearchResults.svg";
+import blockPlaceholder from "../../../../../../../../public/images/blockPlaceholder.png";
+export const BlockTabs = (props: any) => {
+  const [value, setValue] = useState(0);
```
shrunyan, line 27: This is hard to understand immediately. "Value" is to generic. What is the context of this value? 

This jumped out to me when it is equated later. `value == 0`. Very hard to know what that means beyond it's literal evaluation.

src/apps/content-editor/src/app/views/ItemEdit/components/ItemEditHeader/DuplicateItemDialog.tsx
```diff
@@ -78,7 +79,13 @@ export const DuplicateItemDialog = ({ onClose }: DuplicateItemProps) => {
     })
       .unwrap()
       .then((res) => {
-        history.push(`/content/${modelZUID}/${res.data.ZUID}`);
+        history.push(
```
shrunyan, line 82: This is the second place I've seen this logic. Should this be extracted?

src/apps/content-editor/src/app/components/Editor/Field/Field.tsx
agalin920: Good question. Yes the name and label are passed as props from parent but in this datatype field we ignore those props and get the same data directly from the field data. 

Why this was problematic for blocks is because there is a custom field for "variant title" and it should respect the parent props as it is not a field actually existent in the model

Reasoning why I didn't opt to go fully for dropping fieldData?.name in favor for name is to minimize code change footprint and any potential edge cases

src/apps/content-editor/src/app/views/ItemEdit/components/BlockTabs.tsx
agalin920: Noted. Used for active tab state but yea not clear.

src/apps/content-editor/src/app/views/ItemEdit/components/ItemEditHeader/DuplicateItemDialog.tsx
agalin920: @shrunyan Good point. Extracting these into url generator helper functions would be a great idea

PR #3106, opened 2024-12-18
[3096] Content-Create-Item: Added an identifier when user inputs meta…

geodem127:
… discription manualy


agalin920 (changes requested):
.

agalin920:
Is there a way to do this without muddying up the content item state? Im worried that it might have unintended sideffects

PR #3108, opened 2024-12-20
[#3048] - Added logic to properly handle evaluated values that are no…

### LINKED ISSUE #3048: TypeError: Cannot read properties of undefined (reading 'toLowerCase')
Sentry Issue: [MANAGER-UI-292](https://zestyio.sentry.io/issues/6023544719/?referrer=github_integration)

```
TypeError: Cannot read properties of undefined (reading 'toLowerCase')
  at parsedDate (./src/shell/components/FieldTypeDate/index.tsx:57:33)
  at <object>.{children#3}.slotProps.field.onChange (./src/shell/components/FieldTypeDate/index.tsx:210:42)
...
(14 additional frame(s) were not displayed)
```

geodem127:
…t strings

### DIFF
--- a/cypress/e2e/blocks/tests/blocks.spec.js
+++ b/cypress/e2e/blocks/tests/blocks.spec.js
@@ -49,6 +49,8 @@ describe("All Blocks Tests", () => {
   it("creates a variant", () => {
     cy.contains(CypressTestBlock).click();
     BlockPage.createVariant(CypressTestVariant);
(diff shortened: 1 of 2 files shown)


finnar-bin (commented):
Why is it necessary to check if this is not a number?

geodem127:
In this line of code, isNaN isn’t strictly necessary since the input is defined as a string, and ?? would handle undefined or null values. However, it ensures that the .toLowerCase() method is not executed on numeric inputs, which would otherwise cause an error.

PR #3110, opened 2024-12-27
[3096] Content-Create-Item: Added an identifier when user inputs meta…

geodem127:
… discription manualy

### DIFF
--- a/src/apps/content-editor/src/app/components/Editor/Editor.js
+++ b/src/apps/content-editor/src/app/components/Editor/Editor.js
@@ -289,13 +289,17 @@ export default memo(function Editor({
               ?.replaceAll("&nbsp;", " ")
               ?.slice(0, 160) || ""
           );
-
-          dispatch({
-            type: "SET_ITEM_WEB",
-            itemZUID,
-            key: "metaDescription",
-            value: cleanedValue,
-          });
+          const userInputMetaData = JSON.parse(
+            sessionStorage.getItem("userInputMetaData") || "[]"
+          );
+          if (!userInputMetaData.includes("metaDescription")) {
+            dispatch({
+              type: "SET_ITEM_WEB",
+              itemZUID,
+              key: "metaDescription",
+              value: cleanedValue,
+            });
+          }
 
           if ("og_description" in metaFields) {
             dispatch({
--- a/src/apps/content-editor/src/app/components/Editor/Editor.js
+++ b/src/apps/content-editor/src/app/components/Editor/Editor.js
@@ -46,7 +46,12 @@ export default memo(function Editor({
   const dispatch = useDispatch();
   const isNewItem = itemZUID.slice(0, 3) === "new";
   const { data: fields } = useGetContentModelFieldsQuery(modelZUID);
+  const { data, web } = useSelector(
+    (state) => state.content[isNewItem ? `new:${modelZUID}` : itemZUID]
+  );
   const [isLoaded, setIsLoaded] = useState(false);
+  const [metaDescription] = useState(web?.metaDescription);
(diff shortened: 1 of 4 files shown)


src/apps/content-editor/src/app/components/Editor/Editor.js
```diff
@@ -46,7 +46,12 @@ export default memo(function Editor({
   const dispatch = useDispatch();
   const isNewItem = itemZUID.slice(0, 3) === "new";
   const { data: fields } = useGetContentModelFieldsQuery(modelZUID);
+  const { data, web } = useSelector(
```
agalin920: The item is already in scope e.g the item prop. Don't see the reason for this selector

agalin920 (changes requested):

src/apps/content-editor/src/app/components/Editor/Editor.js
```diff
@@ -355,6 +361,32 @@ export default memo(function Editor({
     }
   }, [isNewItem, setIsLoaded, applyDefaultValuesToItemData]);
 
+  useEffect(() => {
```
agalin920: This logic is really hard to follow. I would expect this logic to exist within the call stack of the change handler. But will let @finnar-bin weigh in as well

src/apps/content-editor/src/app/components/Editor/Editor.js
```diff
@@ -355,6 +361,32 @@ export default memo(function Editor({
...
+    if (!isNewItem) return;
+    const isMetaDescriptionInitState = metaDescription === web?.metaDescription;
+    const isDescriptionChanged = data?.description !== web?.metaDescription;
```
agalin920: Careful here, this is incorrect. Data key is the set fields on a model. Therefore data.description is not a guaranteed field nor is the set logic looking for this field. Fields are defined by the user e.g myDescription, someDescription, mySecondTextField etc...

agalin920 (changes requested):

src/apps/content-editor/src/app/components/Editor/Editor.js
finnar-bin: Yeah I kind of agree with what @agalin920 commented.

Correct me if I'm wrong but wouldn't it be easier to just have 3 states (metaDescription, og_description and tc_description) that holds a boolean value, then set the state to true if the user has done a manual input on those fields (the field names are static so it should be easy to find out through the onChange handler if any of those fields have been manually updated). Then on [these lines](https://github.com/zesty-io/manager-ui/blob/bd3beb6baf5e1b54c8d0d9b32e1378a59e68ceb4/src/apps/content-editor/src/app/components/Editor/Editor.js#L293C1-L316C12) do not do the dispatch if the state is true since that would mean the user has performed a manual update.

finnar-bin (changes requested):

src/apps/content-editor/src/app/components/Editor/Editor.js
geodem127: @finnar-bin , this would be the case if "description" and "metaDescription" consistently matched after the page load (without persisting any data). I had a similar thought initially, but things get more complex when the user refreshes the page or navigates to a different one. In such scenarios, metaDescription retains its previous value even if the meta component hasn’t fully loaded in the view, causing it to mistakenly assume that metaDescription has changed. That said, I might have missed something, so a walkthrough would be helpful to clarify.

src/apps/content-editor/src/app/components/Editor/Editor.js
geodem127: Sorry, I totally missed it. Updated this already

PR #3114, opened 2025-01-02
[Content] Workflow status

### LINKED ISSUE #2132: Content: Workflow Statuses
# Problem
Many current customers have asked for the ability to add custom statuses to the content items they are working on so as to allow for them to have a more streamlined efficient workflow. Having the ability to set and see an item's current statuses will allow an organization and it's members to: 

1. Know what stage of the content lifecycle a content item is at and which team is working on it
2. Whether a content item should be published and prevent accidental publishing of items which are still a work in progress
3. Add and remove statuses to indicate whether a stage is complete or not only if they have permission to do

# Solution

## Users can view ,and manage custom statuses from Settings
A user can create, re-order, edit, and de-activate statuses via the Workflows tab in Settings

Image

## Users can create custom statuses from here
For any custom status a user can set it's 

1. Name
2. Description
3. Color
4. Which roles can add this status? 
5. Which roles can remove this status?
6. Allow publish with this status? 

Image

## Users can add a status to a version of a content item via the version dropdown when editing a content item
Users (with the appropriate permissions) can immediately add and remove statuses via the version dropdown(e.g. Draft)

Image

## Users cannot publish items unless they have a status that allows for the same
Failure to have the appropriate status, will display an error similar to the one in the image below. This helps prevent accidental publishes of items still in progress. 
image

### Users cannot add or remove a status unless they have permission to do so
This ensures only the team with the appropriate permissions can mark whether an item is approved or ready to be reviewed or needs changes, etc. 

Image

## Users without the permission to edit Workflow Statuses will not be able to access the Workflows Admin Panel
This will ensure only users with the role Owners and Administrators can edit Workflow statuses and prevent users from other roles from accidentally editing or modifying statuses and status orders. 

image

# Figma
Phase 1: https://www.figma.com/file/8wqowEpgXM6Ru4BVdnVyXx/Workflows?type=design&node-id=117%3A42581&mode=dev&t=hXgnY8e7ynA4pDsL-1

# Developer Notes
All developer notes have been put as sticky notes next to each flow. I have also listed critical notes here as well for immediate context that can help before reviewing the design. 

## Workflow Admin Panel in Settings

1. In the Settings App, create a route that says Workflows under a section called User Settings (as indicated here in this mockup)
2. By default there should be 3  custom statuses created for all instances that have workflows switched on by default. These statuses should only have a name and description. There should be no rules applied to them. A user should be able to edit these and also deactivate them. They should be in the order: Draft, Needs Review, Approved.
3. Needs confirmation: Workflows should be loaded for any instance that already exists and for every new instance that is created. 
4. Sort statuses in the sorting order set by the user. 
5. Only users with the role: Owner, Staff, or Admin should be able to view this page. If they do not have permission, they will see the screen on the right. 
6. A user should be able to re-order statuses using the drag indicator on the left of the list item. Similar to the experience we have for re-ordering fields for a model in Schema. This sorting order should be remembered and reflected in the dropdown in the Content Editing experience as well. 
7. A new status created by the user appears at the end of the list. As soon as a new status is created, please put it into Active state (semi transparent orange bg) so that is easy for the user to see where it is. This is similar to the experience of creating a new folder in Media. Clicking anywhere outside of the list item will immediately make it go into a default state. 

## Status Data we need to store

When a user creates a status we need to store the following information: 

1. Name
2. Description
3. Color
4. Which roles can add this status? - this ensures a status can only be added by one or more role types
5. Which roles can remove this status? - this ensures a status can only be removed by one or more role types
8. Allow publish with this status - this ensures a content item can only be published once it gets a status that allows for publishing
9. Key note: a user should be able to publish a content item with any status until it has atleast one status that has “Allow publish with this status” checked

## Color Options
The Color Options in the dropdown should be: 

1. Red: Red 500 
2. Blue - Blue 500 
3. Green - Green 500
4. Yellow - Yellow 500 
5. Orange: Orange 500
6. Pink: Pink 500 
7. Rose: Rose 500 
10. Deep Purple: Deep Purple 500 
11. Purple: Purple 500 
12. Grey: Grey 500 

### Adding a Status
Once a status is added: 
1. Send an email to all users who have the roles which have the ability to remove the status. See the email format to follow on the right.
2. Add a log to the content item’s activity log - “[Status Name] status added to [Version Number]” by [User Name] on [Date Time]

#### Email Format 

From: 
[Status Added By Name]

Subject: 
[Instance Name] [Status Name] status added to [Content Item Meta Title] [Version Number]

Email Content: 
[Status Added By Name] added a [Status Name] status added to [Content Item Meta Title] [Version Number]

Open Content Item on Zesty [link to content item] | You have received this email because your role has the ability to remove this status

Note: If no Content Item Meta Title is available, then use the Navigation Text

#### Email Example

From
Gisele Blair <notifications@zesty.io> 
Subject
[Sony Alpha Universe] Draft status added to What's In My Bag: A Hybrid Sony Alpha Kit For High-Level (v12)
Email Content
Gisele Blair added a Draft status added to What's In My Bag: A Hybrid Sony Alpha Kit For High-Level (v12)

Open Content Item on Zesty | You have received this email because your role has the ability to remove this status

### Removing a Status

1. Send an email to all users who have the roles which have the ability to add and remove the status. See the email format to follow on the right.
2. Add a log to the content item’s activity log - “[Status Name] status removed from [Version Number]” by [User Name] on [Date Time]

#### Email Format

From: 
[Status Added By Name]

Subject: 
[Instance Name] [Status Name] status added to [Content Item Meta Title] [Version Number]

Email Content: 
[Status Added By Name] added a [Status Name] status added to [Content Item Meta Title] [Version Number]

Open Content Item on Zesty [link to content item] | You have received this email because your role has the ability to remove this status

Note: If no Content Item Meta Title is available, then use the Navigation Text

#### Email Example
From
Gisele Blair <notifications@zesty.io> 
Subject
[Sony Alpha Universe] Draft status added to What's In My Bag: A Hybrid Sony Alpha Kit For High-Level (v12)
Email Content
Gisele Blair added a Draft status added to What's In My Bag: A Hybrid Sony Alpha Kit For High-Level (v12)

Open Content Item on Zesty | You have received this email because your role has the ability to remove this status

## Publishing Rules
If any of the statuses created by the user have a rule that says Needs Status to Publish, then check if that status is applied and also check if the user has permissions to publish this content item. If the answer is yes to both only then allow them to publish. Else show the error as shown down below. 

image

## Adding a Status and removing a status rules
If a user Tries to Add a Status (but they do not have permission to do so) then show the following tooltip. Tooltip should follow cursor. https://mui.com/material-ui/react-tooltip/#follow-cursor
image

If a user Tries to Remove a Status (but they do not have permission to do so) then show the following tooltip. Tooltip should follow cursor. https://mui.com/material-ui/react-tooltip/#follow-cursor

image

finnar-bin:
Resolves #2132 

### Dependencies

- [x] https://github.com/zesty-io/instances-api/pull/598
- [x] https://github.com/zesty-io/gcp-cf/pull/192

### Preview
[screen-recorder-thu-jan-02-2025-12-59-14.webm](https://github.com/user-attachments/assets/10125d23-3a6b-4c50-8b2c-3c63ef494ba6)


cypress.config.js
```diff
@@ -5,6 +5,7 @@ module.exports = defineConfig({
   viewportWidth: 1920,
   viewportHeight: 1080,
   video: false,
+  defaultCommandTimeout: 30000,
```
agalin920: This is a pretty large increase. Any reasoning?

finnar-bin: I'm not sure if it's just on my end but when I'm running the tests locally, sometimes the ui takes some time to update which causes the tests to fail prematurely even if I've already made sure the api calls have succeeded via intercept and wait. I didn't want to add timeouts to each cy.get hence decided to make it a global setting.

It may not be needed for the CI though

agalin920: it takes up to 30 seconds?

src/shell/services/instance.ts
```diff
@@ -607,6 +615,65 @@ export const instanceApi = createApi({
...
     }),
+    getItemWorkflowStatus: builder.query<
+      ItemWorkflowStatus[],
+      { modelZUID: string; itemZUID: string }
+    >({
+      query: ({ modelZUID, itemZUID }) =>
+        `/content/models/${modelZUID}/items/${itemZUID}/labels`,
+      transformResponse: getResponseData,
+      providesTags: (_, __, { itemZUID }) => [
+        { type: "ItemWorkflowStatus", id: itemZUID },
```
agalin920, line 630: Confused as this provides tags with id but all the mutations invalidate ItemWorkflowStatus without id. What does this mean in terms of cache invalidation

src/apps/content-editor/src/app/views/ItemEdit/components/ItemEditHeader/VersionSelector/VersionItem.tsx
```diff
@@ -0,0 +1,396 @@
...
+
+import {
+  useGetWorkflowStatusLabelsQuery,
+  useUpdateItemWorkflowStatusMutation,
+} from "../../../../../../../../../shell/services/instance";
+import {
+  User,
+  WorkflowStatusLabel,
+} from "../../../../../../../../../shell/services/types";
+// import { WORKFLOW_LABELS as statusLabels } from "./mocks";
```
agalin920: Shouldnt this comment and file be removed?

agalin920 (changes requested):

src/apps/content-editor/src/app/views/ItemEdit/components/ItemEditHeader/VersionSelector/VersionItem.tsx
```diff
@@ -0,0 +1,396 @@
...
+import { AppState } from "../../../../../../../../../shell/store/types";
+import { useGetUsersRolesQuery } from "../../../../../../../../../shell/services/accounts";
+import { NoResults } from "./NoResults";
+
+export const BG_COLOR_MAPPING: Record<string, string> = {
```
agalin920, line 47: What is the reasoning for this map?

src/apps/settings/src/app/views/User/Workflows/RestrictedPage.tsx
```diff
@@ -0,0 +1,171 @@
...
+      >{`${role} • ${email}`}</Typography>
+    </Box>
+  </Box>
+);
+
+const RestrictedPage = () => {
+  const { isLoading, isError, data } = useGetUsersRolesQuery();
+
+  const profileList = data
+    ?.filter((profile) => AUTHORIZED_ROLES.includes(profile?.role?.name))
```
agalin920: @finnar-bin is this correct? Should we be checking system role instead?

cypress.config.js
finnar-bin: When I run it locally on my machine it was what works for me. I'll remove it so it goes back to the default 10secs and hopefully it still passes on the CI

src/apps/settings/src/app/views/User/Workflows/RestrictedPage.tsx
finnar-bin: Yeah, @geodem127 instead of checking the user's role name, check their system role zuid instead and make sure it's either 31-71cfc74-0wn3r (owner) or 31-71cfc74-4dm13 (admin)

src/shell/services/types.ts
```diff
@@ -589,3 +618,98 @@ export type GroupItem = {
   createdAt: string;
   updatedAt: string;
 };
+
+export type WorkflowStatusLabelQueryParams = {
+  showDeleted?: boolean;
+};
+
+export const COLOR_NAMES = {
```
agalin920: Dont really like this map nor the color_hex map. Especially on this file. This would pobably be better fitting in the view itself or as a utility. This file is mostly to hold API relevant types

src/shell/services/types.ts
```diff
@@ -589,3 +618,98 @@ export type GroupItem = {
...
+  ORANGE: "#FF5C08",
+  PINK: "#EE46BC",
+  PURPLE: "#7A5AF8",
+  RED: "#F04438",
+  ROSE: "#F63D68",
+  YELLOW: "#F79009",
+  GREY: "#667085",
+} as const;
+
+export const AUTHORIZED_ROLES: string[] = ["Admin", "Owner"];
```
agalin920: Same here. There isnt enough context in this constant for it to be used globally

agalin920 (changes requested):

src/apps/content-editor/src/app/views/ItemEdit/components/ItemEditHeader/VersionSelector/VersionItem.tsx
finnar-bin: So, when a label is created, the "color" data saved in the backend is the hex value keys, this map is used to determine what background color will be used for the Chip component since the hex values stored in the backend is used for the text color.

src/shell/services/types.ts
```diff
@@ -589,3 +618,98 @@ export type GroupItem = {
...
+  label: ColorNameTypes;
+  value: ColorHexTypes;
+};
+
+export type StatusLabelsRoleMenu = {
+  label: string;
+  value: string;
+};
+
+export const colorMenu: StatusLabelsColorMenu[] = Object.keys(COLOR_NAMES).map(
```
agalin920: This is not a type and this file for defining types.

agalin920 (changes requested):

src/shell/services/instance.ts
finnar-bin: Oh yeah good catch, I'll update the mutations

src/apps/content-editor/src/app/views/ItemEdit/components/ItemEditHeader/VersionSelector/VersionItem.tsx
agalin920: @finnar-bin If its being set by the backend why are we mapping it to something else? Shouldnt we be respecting what BE defines? Otherwise whats the purpose of it being defined in the BE

finnar-bin: So what the backend stores is the color for the text inside the chip. The chip then needs to have different background colors based on the text color. That's where this map comes into play to easily determine what the chip background color needs to be based on what the text color data was provided.

src/shell/services/types.ts
geodem127: I'll be placing these, along with everything else unrelated, into the "workflows" directory.

src/shell/services/types.ts
geodem127: I'll be placing these, along with everything else unrelated, into the "workflows" directory.

src/shell/services/types.ts
geodem127: Will update this and relocate to the "workflows" dir

src/apps/content-editor/src/app/views/ItemEdit/components/ItemEditHeader/VersionSelector/VersionItem.tsx
finnar-bin: Removed on latest commit

src/shell/services/instance.ts
finnar-bin: Updated on latest commit

cypress.config.js
finnar-bin: Removed on latest commit

src/apps/settings/src/app/views/User/Workflows/RestrictedPage.tsx
geodem127: Got it. I'll update this accordingly.

src/apps/content-editor/src/app/views/ItemEdit/components/ItemEditHeader/VersionSelector/VersionItem.tsx
agalin920: @finnar-bin gotcha thanks

src/apps/settings/src/app/views/User/Workflows/index.tsx
agalin920, line 1: This view path is kind of awkward. This make me think there is a User view and then a sub view within the user view called workflows. But it looks like everything is just inside the nested /views/user/workflows folder. Whats the purpose of having the /User path part?

geodem127: The "Workflows" section is located under User Settings. As per the design, there is likely to be additional subviews within this section.


### **FIGMA DESIGN:**


image

agalin920 (approved):

geodem127 (approved):

PR #3119, opened 2025-01-04
Refactor hover effect: Use mouse events to trigger hover state for be…

### LINKED ISSUE #3098: Global: Sidebar Nav Items - Retain the Hover Background on the item after more menu is opened
# Bug Description
Watch Loom: https://www.loom.com/share/dca2f7e0d50f45c9900b8934d68da107

# Expected behavior

1. **If the nav list item is not in active state:** Retain the hover background of the list item
2. **If the nav list Item is in active state:** Retain the active state background of the list item

**See Reference:** 
CleanShot 2024-12-15 at 20 26 11

### LINKED ISSUE #3102: Global: Bugs & Enhancements
# Bugs
- [x] https://github.com/zesty-io/manager-ui/issues/3098

# Enhancements
- [x] https://github.com/zesty-io/manager-ui/issues/2928

geodem127:
…tter control

### DIFF
--- a/src/shell/components/NavTree/components/NavTreeItem.tsx
+++ b/src/shell/components/NavTree/components/NavTreeItem.tsx
@@ -145,7 +145,7 @@ export const NavTreeItem: FC<Props> = React.memo(
         }}
         ContentProps={{
           className: itemId === hoveredItemId ? "is-hovered" : "",
-          onMouseOver: (event: any) => {
+          onMouseEnter: (event: any) => {
             setHoveredItemId(itemId);
           },
           onDragOver: (event: any) => {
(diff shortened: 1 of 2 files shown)


finnar-bin (approved):

src/shell/components/NavTree/components/NavTreeItem.tsx
```diff
@@ -136,6 +144,10 @@ export const NavTreeItem: FC<Props> = React.memo(
           },
         }}
         ContentProps={{
+          className: itemId === hoveredItemId ? "is-hovered" : "",
+          onMouseOver: (event: any) => {
```
agalin920: concerned with setting state on mouse over. Wouldnt it be better to set on menu click? Since that is the only time we need it to show?

agalin920 (changes requested):

src/shell/components/NavTree/components/NavTreeItem.tsx
geodem127: > concerned with setting state on mouse over. Wouldnt it be better to set on menu click? Since that is the only time we need it to show?

I tried this approach, but it causes an issue where the hovered background color flickers when the dialog is closed, as the element gets obscured by the backdrop. To minimize the triggered events, I could switch the "onMouseOver" event to "onMouseEnter". This approach requires the least amount of changes.

agalin920 (approved):

PR #3122, opened 2025-01-07
[3116]#SEO-Tab-Page-Parent-Dropdown: Enhanced sorting logic with adde…

### LINKED ISSUE #3116: Content: SEO Tab - Page Parent Dropdown UX Enhancements
# Problem

**Assume an instance has the following URLs:** 

- /abc/
- /qza/
- /disaster-relief/
- /disaster-relief/abc/
- /abc/disaster-relief
- /abc/qza/disaster-relief/

**Now assume a user searches for  "/disaster-relief/"**

Currently the system will show the results in the following order (A to Z): 

- /abc/disaster-relief
- /abc/qza/disaster-relief/
- /disaster-relief/
- /disaster-relief/abc/

**The user finds this sorting order unintuitive because the exact match (/disaster-relief/) does not appear first. They expect the most relevant results (e.g., the exact match and closely related URLs) to be prioritized over less relevant, nested results.**

The user also would like to search by ZUID which is not possible. 

# Solution

## Improvement 1: Improved Sorting Logic
The solution involves a sorting mechanism that improves the order of search results by prioritizing relevance based on the following criteria:

1. Exact Match First: Display the URL that matches the search query exactly at the top of the results.
2. Hierarchy of Nested Levels: Sort the remaining results based on the number of nested levels (i.e., the depth of the URL structure). URLs with fewer nested levels are considered more relevant.
3. Alphabetical Order: For URLs at the same nested level, sort them with the exact match first and then alphabetically (A to Z).

### Example
If a user searches for /disaster-relief/, the results would be displayed in this improved order:

- /disaster-relief/ (Exact match)
- /disaster-relief/abc/ (One level deeper, sorted alphabetically)
- /abc/disaster-relief (Same nested level as above but sorted with exact match first and then alphabetically)
- /abc/qza/disaster-relief/ (Deeper nested level)

This ensures the results feel logical and aligned with user expectations, with the exact match and closest matches appearing first.

## Improvement 2: Allow Users to Search by ZUID

If a user types in the ZUID, then show the matching content item's URL. 
If the ZUID is of the item the user is on, then do not show any results as an item cannot be it's own parent.

geodem127:
…d support for ZUID-based search.

### DIFF
--- a/src/apps/content-editor/src/app/views/ItemEdit/Meta/settings/ItemParent.tsx
+++ b/src/apps/content-editor/src/app/views/ItemEdit/Meta/settings/ItemParent.tsx
@@ -13,6 +13,7 @@ import { notify } from "../../../../../../../../shell/store/notifications";
 import { searchItems } from "../../../../../../../../shell/store/content";
 import { useGetContentNavItemsQuery } from "../../../../../../../../shell/services/instance";
 import { sortMostRelevantSearch } from "../../../../../../../../shell/components/GlobalSearch/utils";
+import zuid from "zuid";
 
 type ParentOption = {
   value: string;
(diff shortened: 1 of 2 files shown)


finnar-bin (approved):

src/apps/content-editor/src/app/views/ItemEdit/Meta/settings/ItemParent.tsx
```diff
@@ -213,6 +217,25 @@ export const ItemParent = ({ onChange }: ItemParentProps) => {
     }
   }, [rawNavData]);
 
+  useEffect(() => {
+    const normalizedSearch = searchText.toLowerCase();
+    const isZuid = /^[a-z0-9]+-[a-z0-9]+-[a-z0-9]+$/i.test(normalizedSearch);
```
agalin920: we should have an isZUID utility already in the codebase

agalin920 (approved):
@geodem127 this is a lot of extra logic. Can you confirm this has no major performance impact

src/apps/content-editor/src/app/views/ItemEdit/Meta/settings/ItemParent.tsx
geodem127: > we should have an isZUID utility already in the codebase

Got it! I'll use the existing utility for this.

geodem127:
> @geodem127 this is a lot of extra logic. Can you confirm this has no major performance impact

Performance-wise, this actually eliminates the laggy behavior when typing in the input search, with no changes to behavior or API calls. However, I made a mistake in filtering the results and have reached out to Lunar about the correct filtering behavior, so further adjustments are needed.

PR #3130, opened 2025-01-10
1260-Invalid-attempt-to-spread-non-iterable-instance: Added additiona…

### LINKED ISSUE #1260: TypeError: Invalid attempt to spread non-iterable instance.
Sentry Issue: [MANAGER-UI-1P0](https://sentry.io/organizations/zestyio/issues/3464819542/?referrer=github_integration)

```
TypeError: Invalid attempt to spread non-iterable instance.
In order to be iterable, non-array objects must have a [Symbol.iterator]() method.
  at contentVersions (./src/shell/store/contentVersions.js:13:30)
  at next (./src/shell/store/middleware/api.js:102:12)
  at next (./src/shell/store/middleware/api.js:38:14)
  at next (./src/shell/store/middleware/nav.js:15:10)
  at next (./src/shell/store/middleware/app-bus.js:17:12)
...
(10 additional frame(s) were not displayed)
```

geodem127:
…l handler to check if data is an array

### DIFF
--- a/src/shell/store/contentVersions.js
+++ b/src/shell/store/contentVersions.js
@@ -1,7 +1,9 @@
 import { request } from "utility/request";
 
 export function contentVersions(state = {}, action) {
-  const existingVersions = state[action.itemZUID] || [];
+  const existingVersions = Array.isArray(state[action.itemZUID])
+    ? state[action.itemZUID]
+    : [];
 
   switch (action.type) {
     case "FETCH_ITEM_SUCCESS":


agalin920 (changes requested):
Can you please describe in what scenarios this was failing? Want to avoid arbitrary guards without fully understanding the error case

geodem127:
> Can you please describe in what scenarios this was failing? Want to avoid arbitrary guards without fully understanding the error case

This issue occurs when contentVersion returns an object (e.g., {...}). It seems to happen in scenarios such as when creating new content. However, when I attempted to replicate the issue during the creation of new content through the UI flow, I encountered a different error:
Uncaught TypeError: undefined is not iterable (cannot read property Symbol(Symbol.iterator)).

Since I was unable to reproduce the exact error via the UI, I manually modified the passed prop and was then able to replicate the same error:
`Uncaught TypeError: Invalid attempt to spread non-iterable instance. In order to be iterable, non-array objects must have a [Symbol.iterator]()` method.

agalin920 (approved):

PR #3156, opened 2025-01-22
starter-blocks: created starter blocks feature

### LINKED ISSUE #3057: Block: Starter Blocks
# Intro
Users will be able to select from a selection of pre designed blocks when creating a Block Model. This will help accelerate their development process and ability to make quick on the fly changes to their web pages. 

Image

# Spec for Each Block
https://doc.clickup.com/20654803/d/h/kpapk-1291/e9e0ff21bdc81f3

# Figma
https://www.figma.com/design/oiesa7e6Q5mcGEAyOjYCtJ/Blocks-App?node-id=498-59910&t=rB2OuW2YNkcuPqOO-1

geodem127:


src/apps/schema/src/app/components/StarterBlockForm.tsx
```diff
@@ -0,0 +1,469 @@
...
+import {
+  useCreateContentModelMutation,
+  useBulkCreateContentModelFieldMutation,
+} from "../../../../../shell/services/instance";
+import { ContentModel } from "../../../../../shell/services/types";
+
+import { Field } from "./Field";
+import { BlockTypeProps } from "./StarterBlocksDialogue";
+
+const LinkStyle = styled(Link)(({ theme }) => ({
```
agalin920: Why is this link so heavily styled? We don't want to use style components and rely on our design system and only using sx for overrrides

agalin920 (changes requested):

src/apps/schema/src/app/components/StarterBlockForm.tsx
geodem127: Sorry, I assigned the wrong color, which is why the specified color didn't render. I'll be updating it now.

agalin920:
@geodem127 im not observing the html code file being created.

cypress/support/api.js
```diff
@@ -0,0 +1,97 @@
...
+      .forEach((forDelete) => {
+        cy.apiRequest({
+          url: `${API_ENDPOINTS.devInstance}/content/models/${forDelete.ZUID}`,
+          method: "DELETE",
+        });
+      });
+  });
+});
+
+Cypress.Commands.add("createModel", (payload) => {
```
agalin920, line 57: I personally prefer avoiding direct API calls in UI automation because they can become a crutch, potentially leading to unrealistic test scenarios. UI tests should ideally mimic real user interactions, ensuring end-to-end validation of workflows as they would occur in production. Relying too much on API calls within UI tests might bypass critical UI behaviors, making the tests less reflective of actual user experiences.

Curious to hear your thoughts on this: @finnar-bin @shrunyan

agalin920 (changes requested):

public/images/block_blank.png
agalin920, line 1: Don't we already have this image? blocksPlaceholder or something like that

public/images/sb__person__image__1.png
agalin920, line 1: Where are some of these images being used?

agalin920 (changes requested):

agalin920:
Screenshot 2025-02-03 at 11 54 32 AM
Screenshot 2025-02-03 at 11 54 21 AM

Can you fix this issue please. State change after component unmounts due to the history.push()

cypress/support/api.js
finnar-bin: Personally, I think it's fine for test setup and cleanup. Like in this case, the setup to create a model has it's own dedicated test. I think it makes it prevents the issue where during test setup and cleanup it sometimes fails which then causes a chain effect of failures if for example a cleanup was supposed to delete a model and it failed.

src/apps/schema/src/app/components/StarterBlocks/StarterBlockForm.tsx
```diff
@@ -0,0 +1,513 @@
...
+          throw new Error(errorMessage);
+        }
+        const ZUID = createStarterBlockRes?.data?.model?.ZUID;
+
+        setIsLoading(false);
+        onClose();
+        history.push(`/schema/${ZUID}`);
+      } catch (err) {
+        setIsLoading(false);
+        console.error("Error during form submission:", err);
```
finnar-bin, line 229: Is there any error feedback to let the user know that this process failed?

src/apps/schema/src/app/components/StarterBlocks/StarterBlocksSelection.tsx
```diff
@@ -0,0 +1,290 @@
...
+} from "@mui/material";
+import { useCallback, useEffect, useRef, useState } from "react";
+import CloseRoundedIcon from "@mui/icons-material/CloseRounded";
+import PlayCircleOutlineRoundedIcon from "@mui/icons-material/PlayCircleOutlineRounded";
+import SearchIcon from "@mui/icons-material/Search";
+
+import { NoResults } from "../NoResults";
+import { StarterBlockProps, STARTER_BLOCKS } from "./configs";
+
+const CardStyles = styled(Card)(({ theme }) => ({
```
finnar-bin, line 29: Why are we using styled components instead of sx overrides?

src/apps/schema/src/app/components/StarterBlocks/StarterBlocksSelection.tsx
```diff
@@ -0,0 +1,290 @@
...
+  selectBlockType,
+}) => {
+  const searchRef = useRef<HTMLDivElement>();
+  const [filteredBlockTypes, setFilteredBlockTypes] =
+    useState<StarterBlockProps[]>(STARTER_BLOCKS);
+  const [activeBlock, setActiveBlock] = useState(null);
+
+  const [search, setSearch] = useState("");
+
+  function handleBlockSelect(blockType: any) {
```
finnar-bin: Kind of a nitpick but can we stick to using either just regular or arrow functions and not mix them up?

finnar-bin (dismissed):

src/apps/schema/src/app/components/StarterBlocks/StarterBlocksSelection.tsx
geodem127: I need to manage custom card styling with different states using a `className`, along with handling the custom states of its child components. Writing separate CSS and logic for each child component feels redundant.

src/apps/schema/src/app/components/StarterBlocks/StarterBlockForm.tsx
geodem127: Yes. Any unknown errors will be handled in this "catch," while errors related to model creation will be managed within the mutation and included in the response.

public/images/sb__person__image__1.png
geodem127: These are used when creating image fields. Additionally, I’d like to ask if there’s a better approach to adding an image source to an image field. Where would be the best place to store default images?

public/images/block_blank.png
geodem127: yes there is. Updated it and removed the image.

src/apps/schema/src/app/components/StarterBlocks/StarterBlocksSelection.tsx
geodem127: updated

cypress/support/api.js
geodem127: Yes, each scenario has its own dedicated test. These are primarily used for setup and cleanup, ensuring reliable data for conditional testing scenarios. They also help with reusing names and labels when creating items/models.

public/images/sb__person__image__1.png
agalin920: Gotcha that should be fine

agalin920 (approved):

agalin920:
@geodem127 there is some undesirable overflowing scrolling behavior on the model on laptop screen

geodem127:
> @geodem127 there is some undesirable overflowing scrolling behavior on the model on laptop screen

This has been resolved

PR #3176, opened 2025-01-28
Implemented logic to accept '0' as a valid input value and set the de…

### LINKED ISSUE #3152: Schema: Number Field - Bugs
- [x] https://github.com/zesty-io/manager-ui/issues/3153
- [ ] https://github.com/zesty-io/manager-ui/issues/3154

### LINKED ISSUE #3153: Schema: Number Field - User Cannot Set Default Value as 0 unless they manually type it in
Problem: 
When setting a default value it auto filled 0 which is the value I wanted but the UI did not see that as the value and gave me a required value warning. 

Solution: 
If on the rules tab, the user does not type in a default value, please default to 0 and allow the user to save the number field

geodem127:
…fault value to 0

### DIFF
--- a/src/shell/components/FieldTypeNumber.tsx
+++ b/src/shell/components/FieldTypeNumber.tsx
@@ -71,17 +71,10 @@ export const FieldTypeNumber = ({
 
         onChange(+value?.toString()?.replace(/^0+/, "") ?? 0, name);
       }}
-      onKeyDown={(evt: any) => {
-        const inputLength = evt.target.value.length;
-        const allowedKeys =
-          /^\d$/.test(evt.key) ||
-          ["ArrowLeft", "ArrowRight", "Backspace", "."]?.includes(evt.key);
-
-        if (evt.key === "Backspace" && inputLength < 2 && !value)
+      onKeyDown={(evt) => {
+        if ((evt.key === "Backspace" || evt.key === "Delete") && value === 0) {
           evt.preventDefault();
-
-        if ((!allowedKeys || evt.key === "Delete") && !value)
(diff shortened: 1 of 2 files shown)


src/shell/components/FieldTypeNumber.tsx
```diff
@@ -71,10 +71,17 @@ export const FieldTypeNumber = ({
...
       }}
-      onKeyDown={(evt) => {
-        if ((evt.key === "Backspace" || evt.key === "Delete") && value === 0) {
+      onKeyDown={(evt: any) => {
+        const inputLength = evt.target.value.length;
+        const allowedKeys =
+          /^\d$/.test(evt.key) ||
+          ["ArrowLeft", "ArrowRight", "Backspace", "."]?.includes(evt.key);
+
+        if (evt.key === "Backspace" && inputLength < 2 && !value)
```
finnar-bin: Why do we need to check for the inputLength? Wouldn't checking for the value already suffice?

geodem127: Checking the input length is necessary to ensure proper handling of the backspace key in specific scenarios. Without this check, users are unable to delete characters when there are multiple zeros (e.g., "00000") because backspace gets disabled prematurely. By considering the input length, we allow users to backspace through excess characters, enabling them to clear the input correctly.

Additionally, the length check ensures backspace is disabled when there’s only one character left. This prevents the user from deleting the last valid digit, ensuring that invalid or empty inputs are replaced with the default value (0).

finnar-bin (approved):

shrunyan (approved):

PR #3177, opened 2025-01-28
[Content] Set number value to 0 if the api originally returned the value as null

### LINKED ISSUE #3154: Content: Item Cannot Be Saved if the Number Field has a Default Value of 0
Problem: 
If I increased and then decreased back to 0 it let me save the field.
A user set a number field as required and with a default value of 0. However when they tried to save the variant it treated the 0 as a missing required field and threw an error and did not allow the user to save the content item/ 

Solution: 
The user should be able to save the item, as the number field technically had a value aka 0. They do not need to manually re-enter 0 for it to save.

finnar-bin:
Makes sure that users are able to save a content item when a number field originally has a null value

Fixes #3154

### DIFF
--- a/src/shell/store/content.js
+++ b/src/shell/store/content.js
@@ -415,6 +415,17 @@ export function saveItem({
       )
       .map((fieldZUID) => state.fields[fieldZUID]);
 
+    Object.keys(item.data).forEach((key) => {
+      const field = fields.find((field) => field.name === key);
+      if (
+        field &&
+        field.datatype === "number" &&
+        (item.data[key] === null || item.data[key] === undefined)
+      ) {
+        item.data[key] = 0;
+      }
+    });
+
     // Check required fields are not empty strings or null valuess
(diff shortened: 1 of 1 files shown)


geodem127 (approved):

shrunyan:
@finnar-bin this doesn't feel like the correct solution to me. The issue isn't the value being returned by the API. It is how the UI is treating a default `0` value in a required context.

shrunyan (changes requested):
@finnar-bin this doesn't feel like the correct solution to me. The issue isn't the value being returned by the API. It is how the UI is treating a default 0 value in a required context.

shrunyan:
Now that I see @geodem127 PR https://github.com/zesty-io/manager-ui/pull/3176 I'm not sure what the purpose of this PR change would be. @geodem127 PR seems to have the intent of addressing the issue as described in the ticket.

finnar-bin:
Closing this for now as I feel like a better fix can be made to address the issue

PR #3179, opened 2025-01-30
[3096]: Content Create Item - Added logic to check if the meta descri…

### LINKED ISSUE #3096: Content: Create Item - System should not override custom Meta Description set by user
**Describe the bug**
Watch Loom: https://www.loom.com/share/c3d5d7fbd6cd4e698d0d05dd7bb4deff

**Expected behavior**
The system should stop automatically updating the Meta Description, once a user sets a custom meta description value.

geodem127:
…ption was manually entered

### DIFF
--- a/src/apps/content-editor/src/app/components/Editor/Editor.js
+++ b/src/apps/content-editor/src/app/components/Editor/Editor.js
@@ -47,7 +47,8 @@ export default memo(function Editor({
   const isNewItem = itemZUID.slice(0, 3) === "new";
   const { data: fields } = useGetContentModelFieldsQuery(modelZUID);
   const [isLoaded, setIsLoaded] = useState(false);
-  const [prevItemData, setPrevItemData] = useState(undefined);
+  const [prevFirstContentFieldValue, setPrevFirstContentFieldValue] =
+    useState(null);
 
   const metaFields = useMemo(() => {
     if (fields?.length) {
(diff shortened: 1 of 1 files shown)


src/apps/content-editor/src/app/components/Editor/Editor.js
```diff
@@ -297,12 +298,22 @@ export default memo(function Editor({
...
+            dispatch({
+              type: "SET_ITEM_WEB",
+              itemZUID,
+              key: "metaDescription",
+              value: cleanedValue,
+            });
+            setPrevItemData((prev) => {
+              return {
+                ...prev,
+                [name]: cleanedValue,
```
finnar-bin: I don't think it's necessary to save this as an object since the if statement makes it so that the value that goes through here always comes from only 1 field which is the `firstContentField`. This should still work fine if you just save the value as a string.

src/apps/content-editor/src/app/components/Editor/Editor.js
```diff
@@ -47,6 +47,7 @@ export default memo(function Editor({
   const isNewItem = itemZUID.slice(0, 3) === "new";
   const { data: fields } = useGetContentModelFieldsQuery(modelZUID);
   const [isLoaded, setIsLoaded] = useState(false);
+  const [prevItemData, setPrevItemData] = useState(undefined);
```
finnar-bin: Prefer initiating a state with null instead of undefined

finnar-bin (approved):

agalin920 (approved):

PR #3183, opened 2025-02-03
[#3169]: Content Auto Complete Input Autohighlight - added filter opt…

### LINKED ISSUE #3169: Content: Auto Complete Input - The first option sometimes gets into hover state when a search query is typed and sometimes it does not.
# Bug
The first option sometimes gets into hover state on search and sometimes it does not. https://www.loom.com/share/dd78ae7ad2ba439da00c1941dd92535d

# Solution
The first option should always get into hover state once the user types in a search query

geodem127:
…ions to ensure first item is always highlighted

### DIFF
--- a/src/shell/components/withAi/AIGenerator.tsx
+++ b/src/shell/components/withAi/AIGenerator.tsx
@@ -527,6 +527,12 @@ export const AIGenerator = ({
                     (option) => option.value === fieldData.tone
                   )}
                   options={TONE_OPTIONS}
+                  filterOptions={(options, state) =>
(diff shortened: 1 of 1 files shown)


agalin920:
@geodem127  Its not clear what this change is aiming to do. Are we removing full partial substring search for only words that start with the string. If so why is this the desired filtering behavior? Furthermore why would we only apply it to this specific Autocomplete?

geodem127:
Closing this PR for now as it has been deprioritized for further evaluation.

PR #3203, opened 2025-02-12
[Content] Create new item from one to many field

### LINKED ISSUE #2937: Phase 2: One to One & One to Many Field - Create & Add New Button
Users will now be able to create a new item in the related model directly from the one to many field, and then add it to the selected list of related items. This ticket is still WIP.

finnar-bin:
- Allows users to create a new content item directly from a one-to-many field
  Resolves #2937 
- Moved the entire internal link logic to a separate component to ensure that other field types do not unnecessarily re-render.

### Preview
[Screencast from 02-12-2025 01:36:52 PM.webm](https://github.com/user-attachments/assets/5e4fdbf1-4d8d-4ed3-ac17-22055b9761b0)


src/shell/components/RelationalFieldBase/CreateNewItemDialog.tsx
```diff
@@ -0,0 +1,57 @@
...
+  onClose: () => void;
+};
+export const CreateNewItemDialog = ({
+  modelZUID,
+  onItemCreated,
+  onClose,
+}: CreateNewItemDialogProps) => {
+  useEffect(() => {
+    window.addEventListener(CREATE_NEW_ITEM_DIALOG_EVENTS.CLOSE, onClose);
+    window.addEventListener(
```
agalin920: @finnar-bin are there any alternatives than the custom events and event listener approach. Couldn't we use a Context that can invoke this function?

agalin920 (changes requested):

src/shell/components/RelationalFieldBase/CreateNewItemDialog.tsx
finnar-bin: Yeah I guess that would work. I could just wrap the entire content app with the context.

finnar-bin: @agalin920 update to use the context api

agalin920 (approved):

PR #3224, opened 2025-02-17
[Content] Improve image thumbnails

### LINKED ISSUE #3206: Content: Multi Page Table - File Type Column - Incorrect File Thumbnails
# Problem
The file type column thumbnail in the multi-page table is not displaying correctly:

1. Multi-Page Table Thumbnail: Incorrectly displays the generic "Image Icon" even though a file is added. This should only happen when no file is added. 
2. Content Item Thumbnail: Currently shows "Failed to Load". This thumbnail should have appeared in the table as well. 

Loom Video: [Loom Recording](https://www.loom.com/share/ca2727742f8e48dc94f3bd349f300e48?from_recorder=1&focus_title=1)

# Expected Behavior

- If a file is uploaded, the table should display the appropriate thumbnail for that file, matching what is shown in the content item view.
- If no file is uploaded, display the "Image Icon" as a placeholder.

# Solution
Update Thumbnail Logic

1. Ensure the file type column in the multi-page table pulls the correct thumbnail for each file type.
2. Use the same thumbnail displayed in the content item view.
3. Handling Edge Cases
- If the file fails to load, show "Failed to Load" (consistent with the content item view).
- If no file is uploaded, display the default "Image Icon."
- If the URL is invalid, display the red error icon thumbnail as we do in the content item

# Testing
1. Please test with different file types (.png, .docx, .gif, .pptx, etc) to ensure thumbnails are displayed correctly.
2. Validate behavior for missing and unsupported files and invalid URLs.

finnar-bin:
Show the appropriate thumbnail for non-image files
Resolves #3206 

image


src/apps/content-editor/src/app/views/ItemList/TableCells/ImageCell.tsx
```diff
@@ -1,16 +1,19 @@
...
 
 type ImageCellProps = { params: GridRenderCellParams };
 export const ImageCell = ({ params }: ImageCellProps) => {
-  const [hasError, setHasError] = useState(false);
-  const handleImageError = () => {
-    setHasError(true);
-  };
+  const isFileZUID = !!params.value?.startsWith("3-");
+
+  const { data, isFetching } = useGetFileQuery(params.value, {
```
agalin920: Why? this will impact performance and this table is already very heavy

agalin920 (changes requested):

src/apps/content-editor/src/app/views/ItemList/TableCells/ImageCell.tsx
finnar-bin: Without fetching the media zuid's details, I can't get the need parameters to properly render the `FileTypePreview`. In this case, the filename, also I'm not sure if passing in the media-resolver as src would work fine with that component.

shrunyan: How is the filename used in this experience? Shown on hover?

finnar-bin: @shrunyan it's used to determine whether or not the thumbnail is going to render an image or check what file type icon will be rendered.

agalin920: @finnar-bin I see. Can we set up some time to run this locally and check performance dev tool tab to ensure we see no severe impacts.

finnar-bin: @agalin920 yeah I agree. Sure, do you have any specific time in mind?

finnar-bin: Actually, I just realized, I could probably just do a `find` on all media assets instead of doing an api call to get this data, since we're already fetching all media on app load.

I'll update the PR first to make that change.

finnar-bin: @agalin920 I have updated it to use the cached data from `useGetAllBinFilesQuery` which is ran on app load. This way the api call is only done once and the ImageCell component only needs to do a `.find` to get the File data that it needs

shrunyan: This sounds like a brittle solution. If what I'm understanding is we no longer make an API request and rely on the assumption that the data has been fetched already.

While we often do pre-fetch data. Sometimes a complete set. We do not have a guarantee we will have pre-fetched all data. As data sets can be come prohibitively large, to fetch all at once. So we have to strike a balance between checking whether we have data on hand and that it is not to stale versus always requesting the API for data. 

Meaning the strategy should, pretty much always, be 1) check the redux store 2) if no result present request the API

finnar-bin: @shrunyan So since I'm using the RTK query that retrieves all the media items on app load it already does that by default. It automatically refetches the data if it doesn't exist or if there are any updates due to it's default caching behavior and automated re-fetching.

finnar-bin: @shrunyan as we've discussed, I have updated the code so it attempts to search for the file data in case the cache doesn't have it.

src/apps/content-editor/src/app/views/ItemList/TableCells/ImageCell.tsx
```diff
@@ -1,16 +1,58 @@
...
+      { skip: !bins?.length }
+    );
+  useEffect(() => {
+    if (isFetchingAllMediaFiles || !isFileZUID) return;
+
+    const matchedFile = allMediaFiles?.find((file) => file.id === params.value);
 
-  if (!params.value || hasError) {
+    if (!matchedFile) {
+      // If cache doesn't have the file data, attempt to look it up from the server
```
agalin920: In what cases will it not? Its an RTK query hook, that data will always be present (fetched or cached)

agalin920 (changes requested):

src/apps/content-editor/src/app/views/ItemList/TableCells/ImageCell.tsx
finnar-bin: This was added to address @shrunyan 's concern in case the cache may have not have been updated

agalin920: Looking at the thread comment you are correcting the assumption that we do in fact make api call and hydrate the media on every load. Was there an additional discussion im not aware of? Im trying to understand in what real world cases this code path will go down. Bigger concern, what if the file has been deleted? Wouldn't this case cause the table cell to make a call every single time?

agalin920 (approved):

src/apps/content-editor/src/app/views/ItemList/TableCells/ImageCell.tsx
finnar-bin: Stuart and I had a discussion about this last week on a call. If I remember correctly, I think the reason was that it's just a backup in case the cache doesn't have it yet, since in the future what if we decide to limit the response of the api that fetches all the media files then there's a possibility that the file is missing in the cache.

Regarding the concern about a file being deleted, that's actually a good catch. I'll look into that today and see what the behavior is.

finnar-bin: @agalin920 so based on my testing the ImageCell does end up triggering the `getFile` rtk query every time it's rendered (scrolled in and out of view for example) when the file has been deleted. I can't think of any other way to prevent this aside from having a state on the parent that stores all image zuids with deleted files but I think that's just way too much.

shrunyan: Wouldn't the record in the local cache have a `deleted_at` value? That would then inform us the item is deleted. We could not make the API call and display an experience which indicates the fact the image is delete.

finnar-bin: @shrunyan there's a `deleted_at` field but as soon as a file is deleted and rtk query's cache gets invalidated and therefore has to do a get `/bin/1-xxxx-xxxx/files` to update the cache, the deleted file no longer gets returned in the response. Similarly, doing a get `/file/3-xxxx-xxxx` to fetch the file, returns a 404 since the file has already been deleted.

The downside to the code that I added above which @agalin920 is raising is that whenever a user is scrolling and a field with a media item that has been deleted is rendered, it will always attempt to do a get `/file/3-xxxx-xxxx` and will always return a 404, it doesn't loop per se but it does do the api call everytime that item gets rendered to the virtualized table.

shrunyan: With deleted files being requested every time they are rendered I'm not sure it matters that much. As it is about just as chatty with the API as files which are not deleted, those records have to fetched as well. A bit more as the user could scroll away and back which would then cause a network request versus pulling it from cache(_as you stated it wouldn't be present due to it being deleted and purged from the local rtk query cache_). The ratio of deleted images to non-deleted images I would have to imagine is very low. Additionally there could be an argument to be made about wanting to express the fact the associated image to the content record is deleted. Otherwise how else would a user know that. **All-in-all I am not sure I see the re-fetching of referenced images which have been deleted as something we should avoid**.

Towards @agalin920 primary point around what is the purpose of this check, as the assumption is the data should be present either from network or cache. I'm assuming that is in reference to the `isFetchingAllMediaFiles` call in this same component. That call seems to aggressive to me. If the component is provided a parameter of the media ZUID why are we not doing a single media file check against cache and then network? To jump ahead of a response of "it would cause to many requests" is that what the all media fetch at app load covers?

finnar-bin: @shrunyan I see what you mean. Waiting for @agalin920 to respond before merging this.

agalin920: @finnar-bin  it would cause to many requests" is that what the all media fetch at app load covers? 

To answer this: Yes. All media is loaded into cache on app load since it is within many apps and global search. So while the call seems aggressive its a call that is already made by other parts of the app so its not inherently making the call but hooking into the data.

My argument is that if the media fully loaded into cache on app load and we get a cache miss, it is very likely that the item has been deleted and a attempt to single fetch it would most of the time resolve in a 404. Only time it wouldn't would be a multi user environment where another user added media and the other user has not done a full refresh therefore item is not in their cache.

A content item could be referencing a 'deleted image'  indefinitely causing this network to call to go off every single time the table is viewed

So the trade off is potential indefinite N image 404 api calls vs a niche multi user case that can be resolved with a refresh

finnar-bin: @agalin920 I see your point. Personally, I'm leaning more towards choosing the latter and having the user manually refresh to get a media item uploaded by a different person since it does remove the problem with having indefinite 404 image api calls. As discussed  with Andres, we don't normally guard on a different user making changes, so I believe removing the need to do a separate API call for each image isn't really necessary.

cc @shrunyan

finnar-bin: @agalin920 @shrunyan Updated to remove separate api calls to fetch image data

shrunyan (approved):

PR #3226, opened 2025-02-17
[3222] :Schema Sidebar - Scroll to Model List Item on Model Creation

### LINKED ISSUE #3222: Schema: Sidebar - Scroll to Model List Item on Model Creation
Watch Loom: https://www.loom.com/share/ce5a61d8ad834de4912572c2f535e54d

geodem127:


src/shell/components/NavTree/components/NavTreeItem.tsx
```diff
@@ -32,13 +33,38 @@ export const NavTreeItem: FC<Props> = React.memo(
...
 
+    useEffect(() => {
+      if (selected === nodeId && !!itemTreeRef.current) {
+        setTimeout(() => {
+          const targetRect = itemTreeRef?.current?.getBoundingClientRect();
+          const parentRect = itemTreeRef?.current
+            ?.closest('[data-nav="app-sidebar"]')
+            ?.parentElement?.getBoundingClientRect();
+
+          const isVisible =
```
agalin920: do we need to do this check? Any issues always scrolling into view?

src/shell/components/NavTree/components/NavTreeItem.tsx
```diff
@@ -32,13 +33,38 @@ export const NavTreeItem: FC<Props> = React.memo(
...
     const depthPadding = currentDepth * 1;
     const itemId = `${nodeId}-${currentDepth}`;
 
+    useEffect(() => {
+      if (selected === nodeId && !!itemTreeRef.current) {
+        setTimeout(() => {
+          const targetRect = itemTreeRef?.current?.getBoundingClientRect();
+          const parentRect = itemTreeRef?.current
+            ?.closest('[data-nav="app-sidebar"]')
+            ?.parentElement?.getBoundingClientRect();
```
agalin920: This is a code smell. prefer not using dom selectors. We should pass references around.

agalin920 (changes requested):

src/shell/components/NavTree/components/NavTreeItem.tsx
geodem127: clicking an item would trigger the scrollIntoView. But i would replace this with an intersection observer.

src/shell/components/NavTree/components/NavTreeItem.tsx
```diff
@@ -28,12 +31,38 @@ export const NavTreeItem: FC<Props> = React.memo(
     nodeData,
     onItemDrop,
     dragAndDrop = false,
+    selected = "",
   }) => {
+    const itemTreeRef = useRef(null);
     const currentDepth = depth + 1;
     const depthPadding = currentDepth * 1;
+    const [scrolled, setScrolled] = useState(false);
+    const { isVisible } = useIntersectionObserver(itemTreeRef);
```
agalin920: This will add so much overhead if we are making an intersection observer for every item. Very concerned about performance issues. I think i mentioned before but why not always scroll into view why the need to explicitly compute if needed?

agalin920 (dismissed):

src/shell/components/NavTree/components/NavTreeItem.tsx
geodem127: intersection observer has been removed.

finnar-bin (approved):

PR #3234, opened 2025-02-20
[Content] Handle empty url string

### LINKED ISSUE #3228: TypeError: Failed to construct 'URL': Invalid URL
Sentry Issue: [MANAGER-UI-2BE](https://zestyio.sentry.io/issues/6293316831/?referrer=github_integration)

```
React ErrorBoundary Error: Failed to construct 'URL': Invalid URL
  at AnalyticsPropertySelector (./src/apps/content-editor/src/app/views/Analytics/components/AnalyticsPropertySelector.tsx:25:45)
  at qZ (./src/apps/content-editor/src/app/views/Analytics/views/AnalyticsDashboard/index.tsx:46:31)
  at QZ (./src/apps/content-editor/src/app/views/Analytics/index.tsx:18:22)
  at wI (./src/apps/content-editor/src/app/ContentEditor.js:102:22)
  at FU (./src/shell/views/Shell/Shell.tsx:33:20)
...
(40 additional frame(s) were not displayed)

TypeError: Failed to construct 'URL': Invalid URL
  at removeProtocolAndSubdomain (./src/apps/content-editor/src/app/views/Analytics/components/AnalyticsPropertySelector.tsx:14:18)
  at AnalyticsPropertySelector (./src/apps/content-editor/src/app/views/Analytics/components/AnalyticsPropertySelector.tsx:63:15)
  at <anonymous> (./src/apps/content-editor/src/app/ContentEditor.js:113:20)
...
(11 additional frame(s) were not displayed)
```

finnar-bin:
Fixes #3228

### DIFF
--- a/src/apps/content-editor/src/app/views/Analytics/components/AnalyticsPropertySelector.tsx
+++ b/src/apps/content-editor/src/app/views/Analytics/components/AnalyticsPropertySelector.tsx
@@ -11,6 +11,8 @@ type Props = {
 };
 
 const removeProtocolAndSubdomain = (url: string): string => {
+  if (!url) return "";
+
   const urlObj = new URL(url);
   const parts = urlObj.hostname.split(".");
 


src/apps/content-editor/src/app/views/Analytics/components/AnalyticsPropertySelector.tsx
```diff
@@ -11,6 +11,8 @@ type Props = {
 };
 
 const removeProtocolAndSubdomain = (url: string): string => {
+  if (!url) return "";
```
shrunyan, line 14: How is this function used? Will it expect an empty string? Should this rather throw an error? i.e. what should the return contract on this function be?

shrunyan: Also what value was being feed into this function causing the error? Is that value an acceptable value to be provided?

finnar-bin: @shrunyan Actually I wasn't able to replicate this even when I logged in to the client's instance where this issue happened at as indicated on Sentry. The function is used to format a url that's from an api call to the analytics api. I assumed that the function received an undefined value due to it missing from the response that's why I decided to just return an empty string.

The function renders this text
image

agalin920 (approved):

PR #3239, opened 2025-02-21
[Content] Preserve selections from being cleared when filtering items in relational fields

### LINKED ISSUE #3235: Content: One to Many Field - Add Item Modal - Preserve Selected Items from Being Cleared During Search or Filtering
**Describe the bug**
The user provided client feedback regarding an issue in one-to-many fields, where selecting multiple items is not functioning correctly (when search function is used). 
When a user searches for and selects an item, then searches and selects for another item, it appears that only one item remains selected. 
See below video demonstrating the problem was shared. The customer has requested that this issue be prioritized for resolution.

https://github.com/user-attachments/assets/dca604a2-7d7a-4277-846d-135b0e86a432

**To Reproduce**
1. Go to Content with one-to-many field
2. Click on 'Add Existing [field name]'
3. Search an item in the 'Search' field (this is specifically for long list of items)
4. Then select the item being searched. 
5. Search again another item, and when selecting another item - it appears only 1 item is selected

**Expected behavior**
Searching, then selecting items in one-to-many field should work as intended.

finnar-bin:
Preserve all active selections when filtering for items in the add related field dialog
Issue was happening because the DataGrid's selectionModel is getting cleared whenever the number of rows change. Solution was to move all filtering logic to DataGrid via filterModel.

[Content---nar-test-2---Zesty-io---zesty-pw---Manager.webm](https://github.com/user-attachments/assets/6529b385-28cb-4810-ac35-8a1560f6833d)


agalin920 (approved):
Seems like a lot more work than simply storing the selected items in a different dataset

finnar-bin:
That was actually what I tried at first but I had issues when attempting to unselect an item. That's why I thought of just handing over the filtering logic to DataGrid since it keeps the selectionModel even when filtering. But yeah, a lot of code movement had be done :weary:

PR #3240, opened 2025-02-21
Ensure promise rejection is always of an Error instance

### LINKED ISSUE #3229: UnhandledRejection: Non-Error promise rejection captured with value: Object Not Found Matching Id:1, MethodName:update, ParamCount:4
Sentry Issue: [MANAGER-UI-2BJ](https://zestyio.sentry.io/issues/6294002260/?referrer=github_integration)

```
UnhandledRejection: Non-Error promise rejection captured with value: Object Not Found Matching Id:1, MethodName:update, ParamCount:4
```

finnar-bin:
This makes sure that when returning a Promise rejection, it is always of an Error instance

Fixes #3229

### DIFF
--- a/src/utility/request.js
+++ b/src/utility/request.js
@@ -101,7 +101,11 @@ export function request(url, opts = {}) {
       } else {
         // Network errors
         store.dispatch(notify({ message: err.message, kind: "warn" }));
-        return Promise.reject(err);
+        return Promise.reject(
+          err instanceof Error
+            ? err
+            : new Error(err.message || JSON.stringify(err))
+        );
       }
     });
 }


src/utility/request.js
```diff
@@ -101,7 +101,11 @@ export function request(url, opts = {}) {
       } else {
         // Network errors
         store.dispatch(notify({ message: err.message, kind: "warn" }));
-        return Promise.reject(err);
+        return Promise.reject(
+          err instanceof Error
```
agalin920, line 105: In what scenarios is it not an instance of Error?

finnar-bin: Based on my research about the error, it's Outlook's safe link scanning: https://github.com/getsentry/sentry-javascript/issues/3440#issuecomment-865857552

I tried doing a whois for the ip address logged in sentry and it's pointing to Microsoft

finnar-bin: I thought about doing what the others did on that thread to add it to ignoreErrors, but might as well just also convert it to an Error instead of ignoring it.

agalin920 (approved):

PR #3251, opened 2025-02-25
Multi item page performance improvements.

### LINKED ISSUE #3245: Content: Multi Page Table - Slow Initial Load Time
# Problem
- Models with many content items cause long initial load times.  
- Likely due to excessive front-end processing and multiple API calls.  
- Interactive elements like toggle buttons in the table might also be adding to the lag.  
- A lot of data is being processed on the Front-end. Front end processing is particularly inefficient on low-powered computers that also leads to further lag. 

# Expected Solution 
Users are expecting a faster Initial Load Time. 
@agalin920 please provide insights on back-end optimizations that can be done for the same.

### LINKED ISSUE #3247: Content: Multi Page Table - Slow Vertical Scrolling & Sorting
# Problem 
- Sorting operations feel sluggish.

agalin920:
This holistic improvement speeds up
 - Sorting and filtering
 - Initial load
 
 closes #3245

### DIFF
--- a/src/apps/content-editor/src/app/views/ItemList/index.tsx
+++ b/src/apps/content-editor/src/app/views/ItemList/index.tsx
@@ -21,10 +21,6 @@ import { SearchRounded, RestartAltRounded } from "@mui/icons-material";
 import noSearchResults from "../../../../../../../public/images/noSearchResults.svg";
 import { ItemListFilters } from "./ItemListFilters";
 import { useParams } from "../../../../../../shell/hooks/useParams";
-import {
-  useGetAllBinFilesQuery,
-  useGetBinsQuery,
-} from "../../../../../../shell/services/mediaManager";
 import { useDispatch, useSelector } from "react-redux";
 import { AppState } from "../../../../../../shell/store/types";
 import { useStagedChanges } from "./StagedChangesContext";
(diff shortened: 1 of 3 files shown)


src/apps/content-editor/src/app/views/ItemList/index.tsx
```diff
@@ -304,6 +284,29 @@ export const ItemList = () => {
     });
   }, [items, allItems, fields, users, isFieldsFetching, isUsersFetching]);
 
+  /* 
+    We debounce the processed items compute since certain fields can call for items to be fetched if they are not in memory 
+    and we don't want to trigger a heavy computation on every item fetched in parallel. This reduces initial load time.
+  */
+  const debouncedCompute = useMemo(() => {
```
finnar-bin, line 291: Why are we using useMemo here instead of useCallback?

agalin920: Yea both hooks help avoid recreating a function on every render, but here we're generating a new function by calling debounce. Using useMemo means we compute and save that debounced function once (or only when its dependencies change). While useCallback is just shorthand for useMemo(() => fn, deps), it’s more suited to wrapping an existing function rather than computing a new value from a function call. In this case, useMemo is clearer and more appropriate for creating and caching the debounced function.

finnar-bin (approved):

zcolah:
also closes https://github.com/zesty-io/manager-ui/issues/3247

PR #3259, opened 2025-02-26
[Content] Handle error when fields fail to load

### LINKED ISSUE #3230: TypeError: Cannot read properties of undefined (reading 'length')
Sentry Issue: [MANAGER-UI-2BD](https://zestyio.sentry.io/issues/6286408319/?referrer=github_integration)

```
React ErrorBoundary Error: Cannot read properties of undefined (reading 'length')
  at Meta (./src/apps/content-editor/src/app/views/ItemEdit/Meta/index.tsx:110:6)
  at AIGeneratorProvider (./src/shell/components/withAi/AIGeneratorProvider.tsx:39:39)
  at ItemEdit (./src/apps/content-editor/src/app/views/ItemEdit/ItemEdit.js:80:20)
  at wI (./src/apps/content-editor/src/app/ContentEditor.js:102:22)
  at FU (./src/shell/views/Shell/Shell.tsx:33:20)
...
(31 additional frame(s) were not displayed)

TypeError: Cannot read properties of undefined (reading 'length')
  at metaFields (./src/apps/content-editor/src/app/views/ItemEdit/Meta/index.tsx:135:18)
  at Meta (./src/apps/content-editor/src/app/views/ItemEdit/Meta/index.tsx:134:24)
...
(14 additional frame(s) were not displayed)
```

finnar-bin:
Handles error when the api call to fetch fields fail
Fixes #3230 

Screenshot from 2025-02-26 11-53-13

### DIFF
--- a/src/apps/content-editor/src/app/views/ItemEdit/ItemEdit.js
+++ b/src/apps/content-editor/src/app/views/ItemEdit/ItemEdit.js
@@ -105,7 +105,7 @@ export default function ItemEdit() {
   const [fieldErrors, setFieldErrors] = useState({});
   const [SEOErrors, setSEOErrors] = useState({});
   // const [hasSEOErrors, setHasSEOErrors] = useState(false);
-  const { data: fields, isLoading: isLoadingFields } =
+  const { data: fields, isError: fieldsLoadingError } =
     useGetContentModelFieldsQuery(modelZUID);
   const [showDuoModeLS, setShowDuoModeLS] = useLocalStorage(
     "zesty:content:duoModeOpen",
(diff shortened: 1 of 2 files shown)


src/apps/content-editor/src/app/views/ItemEdit/ItemEdit.js
```diff
@@ -170,6 +170,17 @@ export default function ItemEdit() {
     setSEOErrors({});
   }, [location.pathname]);
 
+  useEffect(() => {
+    if (fieldsLoadingError) {
```
agalin920, line 174: Is this something we need to account for? If so I think we should lift this at a higher level because then it begs the question why do we show a toast here but no where else where we fetch fields

agalin920 (changes requested):

src/apps/content-editor/src/app/views/ItemEdit/ItemEdit.js
finnar-bin: You mean like show a toast error when **this** api call fails? 

Actually, this was something I personally decided on adding here since the editor loads and shows the no fields component. I thought it might confuse users thinking that someone deleted all their model's fields therefore showing that error message would immediately let them know that it was just a network issue.

I was actually thinking if there's like a way to make a notification dispatch like this within the rtk query itself so we don't have make the notif dispatch individually on each component but I'm kind of torn since I feel like it's not something the rtk query should be responsible.

agalin920 (approved):

PR #3274, opened 2025-02-27
[#3241] : One to one relationship field selection issue

### LINKED ISSUE #3241: Content App: One to One and One to Many Field - Replace Item Button
Customer finds it hard to update the selection of the one to one relationship field. 
Before, it's easier for them to change/update the value just by clicking the field then a dropdown of selection appears.

Now with the new design, they find it unintuitive to change a value - they would have to remove an item and then add a new one. 
It's confusing for the customer as they say, "_confusion on the pencil icon and having to make multiple (unintuitive) clicks to remove a selection_"

_There is no obvious UI to CHANGE a selection - the pencil icon links out out of your current item and into the related item to edit it._ 
Video: https://www.loom.com/share/aa6ab06fd99043e1bff94a155e8a5b27?sid=06ca2524-c0ce-40f7-83cf-55b7294d4b72

geodem127:
Added a button that opens the FieldSelector dialog, allowing the user to replace the currently selected content item.


src/shell/components/RelationalFieldBase/ActiveItem/index.tsx
```diff
@@ -42,6 +44,23 @@ import { fetchItemPublishing } from "../../../store/content";
 import { SchedulePublish } from "../../SchedulePublish";
 import { useDomain } from "../../../hooks/use-domain";
 
+const ReplaceContentIcon = (props: SvgIconProps) => {
```
agalin920: This should be added to our zesty material repo if it does not exist in @mui/material-icons

agalin920 (dismissed):

finnar-bin (changes requested):
@geodem127 I checked out the branch and upon testing, it doesn't seem like selecting a replacement item is limited to only 1 choice. Is this something that @zcolah specifically requested? I'm under the assumption that a user should **only** be able to select 1 item as a replacement since they're supposed to be replacing **only** that item.

Also, I noticed that upon selecting an item and clicking Done, it **replaces** everything that were already previously selected and not just the item you clicked the "Replace Item" button on.

src/shell/components/RelationalFieldBase/ActiveItem/index.tsx
geodem127: Created a PR to add the icon: https://github.com/zesty-io/material/pull/112

finnar-bin: @geodem127 you should remove this here and update the import to use the icon you added on the material repo

src/shell/components/RelationalFieldBase/FieldSelectorDialog/DialogHeader.tsx
```diff
@@ -89,3 +93,50 @@ export const DialogHeader = ({
     </DialogTitle>
   );
 };
+
+type ReplaceDialogHeaderProps = Omit<
+  DialogHeaderProps,
+  "fieldLabel" | "multiselect" | "replace" | "onDeselectAll"
+>;
+
+export const ReplaceDialogHeader = ({
```
finnar-bin: I would prefer to have this as it's own separate file instead of bundling it here since they have their own distinct functions

finnar-bin (changes requested):

src/shell/components/RelationalFieldBase/FieldSelectorDialog/DialogHeader.tsx
geodem127: updated

finnar-bin (approved):

PR #3294, opened 2025-03-06
[Content] Prevent UI jank when creating or posting reply to a comment

### LINKED ISSUE #3278: Content: Comments - Remove Jitter when creating or adding a new comment or reply
# Bug
View Loom:
https://www.loom.com/share/b6fbffcef67b4e72861ea082a43359de?from_recorder=1&focus_title=1

# Solution 
The experience of adding a comment in should have no jitter and should be smooth similar to Google Docs.

### LINKED ISSUE #3279: Content: Comments - Bugs and Enhancements
# Bugs
- [x] #3277
# Enhancements 
- [x] #3276
- [x] #3278

finnar-bin:
[Content---nar-test-2---Zesty-io---zesty-pw---Manager (2).webm](https://github.com/user-attachments/assets/962dd828-6d9f-4aee-8a5b-0ca75aaf83a1)

### DIFF
--- a/src/shell/components/Comment/CommentItem.tsx
+++ b/src/shell/components/Comment/CommentItem.tsx
@@ -55,6 +55,7 @@ type CommentItemProps = {
   withResolveButton?: boolean;
   withReopenButton?: boolean;
   onParentCommentDeleted: () => void;
+  commentCount: number;
 };
 export const CommentItem = ({
   commentZUID,
@@ -65,6 +66,7 @@ export const CommentItem = ({
   withResolveButton,
   withReopenButton,
   onParentCommentDeleted,
+  commentCount,
 }: CommentItemProps) => {
   const { resourceZUID } = useParams<PathParams>();
   const location = useLocation();
@@ -229,6 +231,7 @@ export const CommentItem = ({
               onCancel={() => setCommentZUIDtoEdit(null)}
               commentResourceZUID={resourceZUID}
               parentCommentZUID={parentCommentZUID}
+              commentCount={commentCount}
             />
           ) : (
             <Typography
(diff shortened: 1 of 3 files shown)


agalin920:
I don't think if optimistic rendering is Truely required for this issue. Why not simply add a loading spinner on the submit?

We dont really do optimistic rendering anywhere else so changing that paradigm would need good reason. It also affects how we should go about our E2E test since we can't rely on UI for accurate result and need to check api call

finnar-bin:
@agalin920 you mean like add a skeleton loader at the end when the comment button is clicked?

agalin920:
I was thinking a spinner on the reply button. That way you don't get that jarring render where the text disappears and then shows up randomly

agalin920:
Actually Looking at the video closely it actually seems the issue is the delayed get call after the submit

agalin920:
Is it easy for us to also pass down the is fetching from the get call so the spinner waits for that as well to clear out the field

finnar-bin:
The current implementation already has the loading spinner on the button on click which I removed on this PR since I opted for optimistic updates. And yeah it's the delay on the get call that's causing it, that's why I thought of just doing optimistic updates while the actual api call happens in the background

finnar-bin:
> Is it easy for us to also pass down the is fetching from the get call so the spinner waits for that as well to clear out the field

Yeah I guess that would work, I think the api call for that is only happening one component above anyway. I'll look into doing that

finnar-bin:
@agalin920 I've updated the PR. Found a simpler way to fix the issue

agalin920 (approved):
Clean 🧼

PR #3307, opened 2025-03-11
[Hotfix] Remove frontend validation limits to one-to-many fields

### LINKED ISSUE #3303: Content: One to Many Field - Too Many Items Selected' Error Appears Incorrectly
**Describe the bug**
Customer reported that they're getting limit error when selecting more items in a one-to-many field (see screenshots below)

Screenshot:
Image
Image

Per checking with Markel, on the backend side field limit of the field is already upto 65k chars so the issue appears to be on the front-end validation.

**Expected behavior**
The frontend check/validation should align with the backend validation.

**Additional context**
Slack thread [here](https://zesty-io.slack.com/archives/C050120G5/p1741377340142769)

finnar-bin:
Removes the predefined frontend validation limit for one-to-many fields and rely on the backend limit validation.

Resolves #3303

### DIFF
--- a/src/apps/content-editor/src/app/views/ItemCreate/ItemCreate.tsx
+++ b/src/apps/content-editor/src/app/views/ItemCreate/ItemCreate.tsx
@@ -298,11 +298,22 @@ export const ItemCreate = () => {
             if (dataLongErrorMatch?.[1]) {
               const fieldName = dataLongErrorMatch[1];
               const errors = cloneDeep(fieldErrors);
+              const oneToManyFieldNames = activeFields?.reduce(
+                (names, currItem) => {
+                  if (currItem?.datatype === "one_to_many") {
+                    return [...names, currItem?.name];
+                  }
+
+                  return names;
+                },
+                []
+              );
 
               errors[fieldName] = {
                 ...(errors[fieldName] ?? {}),
-                CUSTOM_ERROR:
-                  "Cannot save field. Please reduce the total number of items selected.",
+                CUSTOM_ERROR: oneToManyFieldNames?.includes(fieldName)
+                  ? "Cannot save field. Please reduce the total number of items selected."
+                  : "Cannot save field. Value is too long.",
               };
 
               setFieldErrors(errors);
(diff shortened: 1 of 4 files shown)


src/apps/content-editor/src/app/views/ItemEdit/ItemEdit.js
```diff
@@ -390,13 +390,31 @@ export default function ItemEdit() {
...
+          const dataLongErrorMatch = res.error?.match(/'([^']*)'/);
+
+          if (dataLongErrorMatch?.[1]) {
+            const fieldName = dataLongErrorMatch[1];
+            const errors = cloneDeep(fieldErrors);
+
+            errors[fieldName] = {
+              ...(errors[fieldName] ?? {}),
+              CUSTOM_ERROR:
+                "Cannot save field. Please reduce the total number of items selected.",
```
agalin920: Seems like this is too specific to one to many and code branch could occur on any field

src/apps/content-editor/src/app/views/ItemCreate/ItemCreate.tsx
```diff
@@ -291,10 +291,28 @@ export const ItemCreate = () => {
...
+            const dataLongErrorMatch = res.error?.match(/'([^']*)'/);
+
+            if (dataLongErrorMatch?.[1]) {
+              const fieldName = dataLongErrorMatch[1];
+              const errors = cloneDeep(fieldErrors);
+
+              errors[fieldName] = {
+                ...(errors[fieldName] ?? {}),
+                CUSTOM_ERROR:
+                  "Cannot save field. Please reduce the total number of items selected.",
```
agalin920: Same here

src/apps/content-editor/src/app/views/ItemEdit/ItemEdit.js
finnar-bin: I've updated it to only show that error message for one_to_many fields and different error message for the other fields.

src/apps/content-editor/src/app/views/ItemCreate/ItemCreate.tsx
finnar-bin: I've updated it to only show that error message for one_to_many fields and different error message for the other fields.

agalin920 (approved):

PR #3320, opened 2025-03-13
Starter blocks routing to blocks page after creating

geodem127:


agalin920:
If these tests are no longer valid can we remove them instead. Why are we commenting out?

geodem127:
> If these tests are no longer valid can we remove them instead. Why are we commenting out?

I wanted to fix it because those lines were important for validating the correct number of fields created and displayed. But anyway, I've removed them now.

PR #3331, opened 2025-03-18
3244 content multi page table   performance enhancements

### LINKED ISSUE #3266: Content: Multi Page Table - Initial Load Time Needs to Be Reduced to a Maximum of 5 seconds
# Problem 
Currently the Multi Page Table takes more 10 seconds to load extremely large datasets with more than 500 items. This is a major UX frustration for content editors/marketers as it keeps them waiting for too long and slows down their daily workflow in Zesty.

# Solution 
We want to bring this down to a maximum of 5 seconds.

agalin920:
closes [#3266](https://github.com/zesty-io/manager-ui/issues/3266)

### DIFF
--- a/src/apps/content-editor/src/app/views/ItemList/TableCells/OneToManyCell.tsx
+++ b/src/apps/content-editor/src/app/views/ItemList/TableCells/OneToManyCell.tsx
@@ -1,6 +1,6 @@
 import { useEffect, useRef, useState } from "react";
 import { Box, Chip, Tooltip } from "@mui/material";
-import { shallowEqual, useSelector } from "react-redux";
+import { shallowEqual, useDispatch, useSelector } from "react-redux";
 
 import { AppState } from "../../../../../../../shell/store/types";
 import { searchItems } from "../../../../../../../shell/store/content";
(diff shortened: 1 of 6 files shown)


src/apps/content-editor/src/app/views/ItemList/index.tsx
```diff
@@ -182,24 +193,16 @@ export const ItemList = () => {
 
   const computeProcessedItems = useCallback(() => {
     if (!items || isFieldsFetching || isUsersFetching) return [];
-    const fieldMap = fields?.reduce((acc, field) => {
-      // @ts-ignore
-      acc[field.name] = field.datatype;
-      return acc;
-    }, {});
+
+    const userMap = new Map(users?.map((user) => [user.ZUID, user]));
```
finnar-bin, line 191: Is there a significant performance gain to using Map instead of an object/array?

finnar-bin (approved):

shrunyan (approved):

src/apps/content-editor/src/app/views/ItemList/TableCells/OneToManyCell.tsx
```diff
@@ -42,13 +42,22 @@ export const OneToManyCell = ({ items }: OneToManyCellProps) => {
...
   const hiddenItems = items?.length - lastValidIndex - 1;
+  const dispatch = useDispatch();
 
   useEffect(() => {
     setLastValidIndex(
       getNumOfItemsToRender(parentWidth, chipContainerRef.current?.children)
     );
   }, [parentWidth]);
 
+  useEffect(() => {
```
shrunyan, line 53: Won't this exasperate performance? As this does not check the redux stores and instead makes an API request for every item listed every time the view loads.

agalin920: @shrunyan Good observation but this component is very low on the component tree. At the point of this mount the resolution has already been attempted by its parent but has missed and so the prop was maintained as a zuid. 

The reason we pass the zuid and do the single resolution on this component specifically is because this is a virtualized cell in the table so it will only trigger the single resolution api call on view so we dont need to do tons of single resolutions off the bat.

We already do this on internal link/one to one cell I just noticed it was missing on the one to many cell

PR #3361, opened 2025-03-27
[#3341]: 403 forbidden response not showing notification

### LINKED ISSUE #3341: 403 forbidden response not showing notification
**Describe the bug**
When an API request generates a 403 forbidden response code there is no notification shown. This can create confusion on the users end on why the action taken did not occur.

**To Reproduce**
1. Go to https://8-f48cf3a682-7fthvk.manager.stage.zesty.io/media/folder/2-12061579-05gxvc?fileId=3-12ee3f44-rgff58
2. Click on Delete File(Trashcan icon)
4. See lack of error response

**Expected behavior**
Alert toast with error shown. Notification message within the notification drawer added.

**Screenshots**
Image

**Additional context**
Discovered in the lower environment with media deletion limitation but I would guess is not limited to that specific action but the 403 response code as a whole

geodem127:
RCA:
- An error handler was not properly implemented.

FIX:
- Introduced error response handling for the delete file request.
- Added a loading indicator.
- Implemented a notification dispatch to handle errors if the request fails.

https://github.com/user-attachments/assets/9d98c882-15f9-45a2-9acd-f3f0c7725717

### DIFF
--- a/src/apps/media/src/app/components/FileModal/FileModalContent.tsx
+++ b/src/apps/media/src/app/components/FileModal/FileModalContent.tsx
@@ -1,10 +1,9 @@
-import { FC, useEffect, useState, useRef, useCallback } from "react";
+import { FC, useEffect, useState } from "react";
 import {
   Typography,
   Box,
   TextField,
   InputAdornment,
-  Avatar,
   InputLabel,
   IconButton,
   Menu,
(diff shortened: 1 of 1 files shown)


src/apps/media/src/app/components/FileModal/FileModalContent.tsx
```diff
@@ -194,8 +191,16 @@ export const FileModalContent: FC<Props> = ({
...
+      if (!deleteResponse?.error) return handleCloseModal();
+      setShowDeleteFileModal(false);
+      dispatch(
+        notify({
+          kind: "error",
+          message: deleteResponse?.error?.data?.message,
+        })
+      );
     });
-    handleCloseModal();
```
finnar-bin, line 198: Wouldn't this make it so that even if operation was successful the model no longer closes?

geodem127: The handleCloseModal() function closes both the delete confirmation modal and the FileContent modal when the response is not an error.

[
```suggestion
      if (!deleteResponse?.error) return handleCloseModal();
```
]()

agalin920 (approved):

PR #3376, opened 2025-04-03
[#3366]: Redirects: Delete Redirect Confirmation Modal

### LINKED ISSUE #3366: Redirects: Delete Redirect Confirmation Modal
# Problem

Currently, the user experience for deleting redirects in the Redirects App can be confusing. When a user clicks the delete button, the redirect entry disappears immediately from the table without any confirmation prompt or success feedback. This lack of interaction can lead to uncertainty about whether the action was successful and increases the risk of accidental deletions.

**Video of Current Behavior:**
[Loom Recording](https://www.loom.com/share/8303c91fa5ae4be0bf5ff4cba33d0c62)

# Solution
To improve the user experience and prevent accidental deletions, we are implementing two key changes:
1.  **Confirmation Modal:** Before a redirect is deleted, display a modal dialog asking the user to explicitly confirm the action.
2.  **Success Notification:** After a redirect has been successfully deleted, display a notification (e.g., a toast message) to inform the user the action was completed.

* This change also align the deletion process with the rest of the Zesty platform for destructive actions.

Image

Image

# Figma
https://www.figma.com/design/zOI7oSH3fG1XDmmPzsP6PX/Zesty-3.0-Revamp?node-id=5170-66205&t=ZcWwcBJATALjX2Fx-1

geodem127:
1.  **Confirmation Modal:** Before a redirect is deleted, display a modal dialog asking the user to explicitly confirm the action.
2.  **Success Notification:** After a redirect has been successfully deleted, display a notification (e.g., a toast message) to inform the user the action was completed.

## Enhancements
- [x] Make Create Redirect into a Primary Filled Button (instead of Outline). 
- [x] Add ability to resize columns to maintain feature parity.
- [x] Remove pointy arrows from tooltip.
- [x] Add Delete Redirect Confirmation Modal when user clicks on delete button. 

https://github.com/user-attachments/assets/9ddfd1c3-23e7-44c0-a7b1-4e4693b19f9d

### DIFF
--- a/src/apps/seo/src/views/RedirectsManager/RedirectsTable/DeleteDialog.tsx
+++ b/src/apps/seo/src/views/RedirectsManager/RedirectsTable/DeleteDialog.tsx
@@ -1,0 +1,231 @@
+import { memo, useState, useCallback } from "react";
+import { useHistory } from "react-router";
+import LoadingButton from "@mui/lab/LoadingButton";
+import {
+  Button,
+  Dialog,
+  DialogActions,
+  DialogTitle,
+  Typography,
+  Stack,
+  Box,
+} from "@mui/material";
+import { DeleteRounded } from "@mui/icons-material";
+import { useDispatch } from "react-redux";
+import { removeRedirect } from "../../../store/redirects";
+import { notify } from "../../../../../../shell/store/notifications";
+import { DialogContent } from "@mui/material";
+import { RedirectTargetCell } from "./RedirectTargetCell";
+import { CellWrapper } from "./RedirectTable";
+
+interface DeleteDialogProps {
+  open: boolean;
+  onClose: () => void;
+  ZUID: string;
+  path: string;
+  type: string;
+  target: string;
+  code: number;
+}
+
+export const DeleteDialog = memo(function DeleteDialog(
+  props: DeleteDialogProps
+) {
+  const { open, onClose, ZUID, path, type, target, code } = props;
+
+  const [deleting, setDeleting] = useState(false);
+  const history = useHistory();
+  const dispatch = useDispatch();
+
+  const handleDeleteFile = () => {
+    if (!ZUID) return;
+
+    try {
+      setDeleting(true);
+      Promise.resolve(dispatch(removeRedirect(ZUID))).then((res: any) => {
+        setDeleting(false);
+        onClose();
+        if (res.status === 200) {
+          dispatch(
+            notify({
+              kind: "error",
+              message: `Redirect Delete: ${path}`,
+            })
+          );
+        } else {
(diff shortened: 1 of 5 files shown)


src/apps/seo/src/views/RedirectsManager/RedirectsTable/DeleteDialog.tsx
```diff
@@ -0,0 +1,231 @@
...
+
+    try {
+      setDeleting(true);
+      Promise.resolve(dispatch(removeRedirect(ZUID))).then((res: any) => {
+        setDeleting(false);
+        onClose();
+        if (res.status === 200) {
+          dispatch(
+            notify({
+              kind: "error",
```
finnar-bin, line 47: Shouldn't this be "success" considering that the api call was successful?

src/apps/seo/src/views/RedirectsManager/RedirectsTable/DeleteDialog.tsx
```diff
@@ -0,0 +1,231 @@
...
+              message: `Redirect Delete: ${path}`,
+            })
+          );
+        } else {
+          throw new Error(`Redirect Delete failed: ${path}`);
```
finnar-bin: I think showing an error message to let the user know the api call has failed would be useful here.

src/shell/index.js
```diff
@@ -63,12 +63,11 @@ const appTheme = createTheme(theme, {
     },
```
finnar-bin: What is this theme used for again? I was under the impression that we already removed all other themes.

finnar-bin (changes requested):

src/apps/seo/src/views/RedirectsManager/RedirectsTable/DeleteDialog.tsx
geodem127: This is part of Zosh's design

src/apps/seo/src/views/RedirectsManager/RedirectsTable/DeleteDialog.tsx
geodem127: The error is handled in the dispatch(removeRedirect(ZUID)), but I believe this also needs to be updated, as it's no longer necessary, or the logic should be moved here instead. Wil update this

src/shell/index.js
geodem127: This is the root themeProvider. I needed to add a custom background color that can be accessed globally across the code app and tabs. The other changes are small adjustments to fix minor differences in the action palettes.

finnar-bin (approved):

PR #3423, opened 2025-04-17
[#3353]: Content: Multi Page Table - Skeleton Loading Screen

### LINKED ISSUE #3353: Content: Multi Page Table - Skeleton Loading Screen
# Problem 
The multi-page table is blank while loading data, offering no indication of what to expect to a user. This results in a jarring and unclear loading experience, especially when users scroll quickly or are waiting for the table to load.

# Solution
We are adding in skeleton loaders so the user knows what to expect (aka a table). These same skeleton loaders will also load up when the user is scrolling down the table fast and we are still loading the data for the rows being scrolled into view. This solution provides users with a clearer expectation of the final layout and creates a smoother, more cohesive loading experience.

Image

# Figma
https://www.figma.com/design/hFep75jMKpWEHXkq6oK6lP/Loading-Screens?node-id=47-10771&t=gMw6ErumLShyh02C-1

# Developer Notes
The skeleton loader for each kind of cell is provided here: https://www.figma.com/design/hFep75jMKpWEHXkq6oK6lP/Loading-Screens?node-id=49-5267&t=gMw6ErumLShyh02C-1

geodem127:
- [x] created a loading skeleton component that appears while the data is still loading

https://github.com/user-attachments/assets/f4301431-52fb-4374-a17a-9bb5c3192ca4


src/apps/content-editor/src/app/views/ItemList/index.tsx
```diff
@@ -611,11 +622,16 @@ export const ItemList = () => {
...
         ) : (
           <>
-            <ItemListFilters />
             <ItemListTable
               key={modelZUID}
-              loading={isFieldsFetching || isUsersFetching}
+              loading={
+                isFieldsFetching ||
+                isUsersFetching ||
+                isModelItemsFetching ||
```
agalin920: Why are these additional loading conditions being added? Why are they now needed for the table to behave properly?

geodem127: These additional loading conditions are being added to ensure the user sees a clear visual indication of the loading state. Without them, the loading skeleton is often not visible, and in cases where it does appear, it can cause unwanted flickering. These conditions help the table behave more consistently and improve the overall user experience.

agalin920: @geodem127 We shouldnt be waiting on model items fetching since the table should be able to render what is has in cache immediatley

geodem127: Updated.

agalin920 (approved):

finnar-bin (approved):

PR #3449, opened 2025-04-24
3236 mui and mui x dependency updates

agalin920:
depends on https://github.com/zesty-io/material/pull/116


package.json
```diff
@@ -59,7 +53,7 @@
     "@tinymce/tinymce-react": "^4.3.0",
     "@welldone-software/why-did-you-render": "^6.1.1",
     "@zesty-io/core": "1.10.0",
-    "@zesty-io/material": "^0.15.7",
+    "@zesty-io/material": "file:../material/zesty-io-material-0.16.0.tgz",
```
finnar-bin: Pointing to a local package

src/shell/index.js
```diff
@@ -12,10 +12,10 @@ import CssBaseline from "@mui/material/CssBaseline";
 
 import "chart.js/auto";
 
-import { LicenseInfo } from "@mui/x-license-pro";
-LicenseInfo.setLicenseKey(
-  "e9268055e7858ccf7d7bc5d078217f7eTz00ODIyOCxFPTE2OTA3Mzk5NDkxNjgsUz1wcm8sTE09c3Vic2NyaXB0aW9uLEtWPTI="
-);
+// import { LicenseInfo } from "@mui/x-license";
```
finnar-bin: Why is this removed?

finnar-bin (changes requested):

package.json
agalin920: Yep updating till material package gets published

src/shell/index.js
agalin920: This will be commented back in once license is available. For dev purposes this needs to be commented out or else components do not render

finnar-bin (approved):

geodem127 (approved):

src/shell/index.js
shrunyan: @agalin920 has the new license been added in?

PR #3540, opened 2025-05-23
[Content] Notify users when a related item failed to publish due to missing an allow publish status

### LINKED ISSUE #3539: Content - No error message when publishing related one-to-one/many content item without allow publish status
## Problem
When attempting to publish a related one-to-one/many content item that has no Allow publishing status label the api call fails but the user has no way of knowing that it has failed.

## Screenshots

https://github.com/user-attachments/assets/460bed10-cbfd-416f-aa0f-34e10af0e21e

## Steps to reproduce

1. Create a workflow status which has the allow publishing checkbox checked if there is none created yet
2. Go to a content item which has a one-to-one/many field and add a content item that has not been published yet and does not have the allow publishing status
3. Make sure to add the allow publishing status on the current item that's open
4. Click publish
5. Check network tab and see the error, the current open content item is successfully published but the related item with the missing status will fail but no error message is shown to the user

## Expected behavior

The user should be able to know via error message if the related content item failed to publish due to the missing allow publish status

finnar-bin:
Show a notification to let users know when a related content item failed to publish due to missing an allow publish workflow status
Resolves #3539 

[Content---nar-test-2---Zesty-io---zesty-pw---Manager.webm](https://github.com/user-attachments/assets/e479d925-373b-4115-a0ce-f07fa547276d)


finnar-bin:
Temporarily placing on hold as we need to request for a change on the api to provide a [clearer error message instead of assuming that the 403 was caused by a missing workflow status](https://github.com/zesty-io/instances-api/issues/676)

geodem127 (approved):

finnar-bin:
closing in favor of https://github.com/zesty-io/manager-ui/pull/3844

PR #3553, opened 2025-05-29
Update workflows test to make sure clean up is performed even if the test fails

finnar-bin:
Update workflows test to make sure clean up is performed even if the test fails

### DIFF
--- a/cypress/e2e/content/workflows.spec.js
+++ b/cypress/e2e/content/workflows.spec.js
@@ -1,12 +1,13 @@
 import ContentItemPage from "./pages/ContentItemPage";
 import CONFIG from "../../../src/shell/app.config";
 import instanceZUID from "../../../src/utility/instanceZUID";
+const NOW = Date.now();
 
 const INSTANCE_API = `${
   CONFIG?.[process.env.NODE_ENV]?.API_INSTANCE_PROTOCOL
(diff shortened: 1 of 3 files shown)


cypress/e2e/content/workflows.spec.js
```diff
@@ -28,9 +29,40 @@ const LABEL_DATA = {
     removePermissionRoles: ["30-86f8ccec82-swp72s", "30-8ee88afe82-gmx631"],
   },
 };
+const cleanUp = () => {
+  // Delete test content item
+  cy.location("pathname").then((loc) => {
+    const [_, __, modelZUID, itemZUID] = loc?.split("/");
```
geodem127, line 35: Could this potentially fail if the current page is not the intended one? modelZUID and itemZUID might return undefined in that case.

geodem127 (approved):

cypress/e2e/content/workflows.spec.js
finnar-bin: Normally it would that's why I added `failOnStatusCode: false` to handle that since it doesn't really matter if the item we're trying to delete exists or not when this is executed on the before hook, since that just means the content item doesn't exist and the previous test run had a successful clean up. I just executed this on the before hook as well to make sure the labels get cleaned up in case they weren't deleted on the previous test run.

PR #3555, opened 2025-05-29
[#3499] Content: Unable to Create New Item - Required Field Error

### LINKED ISSUE #3499: Content: Unable to Create New Item - Required Field Error
**Description:**
Users are reporting an issue where they are unable to create new items. The UI prompts a " _- Required field_ error" even when all fields appear to be filled out.
Per initial findings:
The `og_image` (which is a special reserved meta field) is set as a *Required field thus an error prompted when no meta image is added, but the error message is not clear as it doesn't specifiy which field must be filled out.

More details of the issue/ steps to replicate found on this [thread](https://zesty-io.slack.com/archives/C050120G5/p1746823523646149)

**Proposed Fix:**
The error message should be updated to be more informative. Instead of a generic " _- Required field_ error," the message should specify which field(s) are missing or invalid. This will help users quickly identify and correct the issue.

geodem127:
**Root Cause:**
- The "og_image" field name is used for the "API/Parsley Code ID," which causes the error because it is a hidden field by default and not required when creating or editing a content item.

**Fix:**
- Update the logic to filter out the og_image field only if it is not marked as required.

**BEFORE:**

https://github.com/user-attachments/assets/1b844a6c-535a-47fa-8633-eb89e73221f7

**AFTER:**

https://github.com/user-attachments/assets/46163888-feab-4af1-bc05-a535647f7a66


finnar-bin (changes requested):
The og_image field has to be hidden from the fields list and should **only** be rendered on the meta tab when editing and be placed under "SEO & Open Graph Settings" right after Meta Description when creating a new content item. Based on the video you attached it's being shown on both the fields list and the meta settings.

The current prod behavior of how the field is being rendered needs to be retained, only the validation needs to be updated.

finnar-bin (changes requested):
@geodem127 tried testing the branch locally. But I found some bugs, click on the field name on the error blurb does not take me to the field where the error occured, in this case the og_image field. This happens to both error blurbs when creating and editing a content.


https://github.com/user-attachments/assets/f0d412cc-ccf7-4735-9d41-70c6f8f05be7

geodem127:
Fixed field error links:

https://github.com/user-attachments/assets/94789222-e62f-4e11-8601-d641c5009cbb

src/apps/content-editor/src/app/views/ItemCreate/ItemCreate.tsx
```diff
@@ -218,14 +257,14 @@ export const ItemCreate = () => {
...
-            res.missingRequired?.reduce(
-              (acc: string[], curr: ContentModelField) => {
-                acc = [curr.name, ...acc];
-                return acc;
-              },
-              []
-            );
+          const missingRequiredFieldNames: string[] = res.missingRequired
+            ?.filter(
+              (field: ContentModelField) => !["og_image"].includes(field?.name)
```
finnar-bin: Instead of filtering it here, why not filter it from the actual redux action?

geodem127: Good catch — filtering from the Redux action is definitely the way to go. I’ll update the code.

finnar-bin (approved):

src/apps/content-editor/src/app/views/ItemCreate/ItemCreate.tsx
```diff
@@ -142,32 +144,69 @@ export const ItemCreate = () => {
...
+        ...prev,
+        og_image: {
+          ...prev["og_image"],
+          MISSING_REQUIRED: !item?.data?.og_image,
+        },
+      }));
+    }
+  }, [item?.data?.og_image, ogImage]);
+
+  const validateRequiredFields = () => {
```
agalin920: If the fix is "Update the logic to filter out the og_image field only if it is not marked as required." 

Why are we adding **another** validation step. Were we not validating anything before?

agalin920 (changes requested):

src/apps/content-editor/src/app/views/ItemCreate/ItemCreate.tsx
geodem127: This was added to validate the required fields before executing the request, as the required field validation and corresponding error messages were not appearing when SEO errors were present. This is because the validation logic is embedded in the Redux function, which does not account for SEO validation.

agalin920: If the redux function doesnt account for it then we were not doing any seo validation? If thats the case then why are we doing seo validation now? 

I would argue that this field should not even be able to be manually made or made required at all to avoid this issue at a more holistic level.

any thoughts here @finnar-bin

src/apps/schema/src/app/utils/index.ts
```diff
@@ -19,6 +19,15 @@ export const modelNameMap = {
   block: "Block",
 };
 
+export const RESERVED_FIELD_NAMES = [
+  "og_image",
+  "og_title",
+  "og_description",
+  "tc_title",
+  "tc_description",
+  "tc_image",
```
finnar-bin: @geodem127 if I remember correctly, the following field names **should** be allowed for manual creation as these fields do not get automatically created like the `og_image` field. We only want `og_image` field to be blocked from manual creation, the others should still be allowed as we have users manually creating these for their models since there's no other way to generate those fields.

finnar-bin (changes requested):
@geodem127 just to verify since I don't see any change related to it, has the check for og_image field been removed even if it was manually created as a required field? We may have existing users who already have this field created as required on their models, so that might still cause an issue during content item creation/update.

geodem127:
> @geodem127 just to verify since I don't see any change related to it, has the check for og_image field been removed even if it was manually created as a required field? We may have existing users who already have this field created as required on their models, so that might still cause an issue during content item creation/update.

Ohh got it. Didn’t consider that case. I’ll add a similar check and error message to cover scenarios where users already have an existing og_image field marked as required

src/apps/content-editor/src/app/views/ItemEdit/Meta/settings/MetaImage.tsx
```diff
@@ -268,68 +279,72 @@ export const MetaImage = ({ onChange }: MetaImageProps) => {
...
+    );
+  }
+
+  // If no image field
+  return (
+    <Box data-cy="og_image" id={field?.ZUID || "og_image"}>
       <FieldShell
         settings={{
-          label: "Meta Image",
+          label: field?.label || "Meta Image",
```
finnar-bin: The meta image field's label should not be customizable

src/apps/content-editor/src/app/views/ItemEdit/Meta/settings/MetaImage.tsx
```diff
@@ -139,7 +145,7 @@ export const MetaImage = ({ onChange }: MetaImageProps) => {
           datatype: "images",
           description:
             "This field allows you to set an open graph image via the SEO tab. An Open Graph (OG) image is an image that appears on a social media post when a web page is shared.",
-          label: "Meta Image",
+          label: field?.label || "Meta Image",
```
finnar-bin: This should not be customizable

finnar-bin (changes requested):
I think there's way too many things going on here, wouldn't it be easier if on `createItem` and `saveItem` redux functions, just check if the field name is `og_image` then skip required validation for it? Since the required fields validation happens on those redux actions. Thoughts @agalin920 ?

src/apps/content-editor/src/app/views/ItemEdit/Meta/settings/MetaImage.tsx
geodem127: updated

src/apps/content-editor/src/app/views/ItemEdit/Meta/settings/MetaImage.tsx
geodem127: updated

geodem127:
> I think there's way too many things going on here, wouldn't it be easier if on `createItem` and `saveItem` redux functions, just check if the field name is `og_image` then skip required validation for it? Since the required fields validation happens on those redux actions. Thoughts @agalin920 ?

Sorry for the confusion—I misunderstood our last discussion. I understand it now and will make the necessary updates.

agalin920 (approved):

finnar-bin (approved):

PR #3563, opened 2025-06-02
Persist global blockAnnouncements call

agalin920:


shrunyan:
@agalin920 looking at the code I'm not sure what this has to do with "persist localStorage across settings workflows tests"? Which if this is indeed doing so I would be concerned about the tests being statefully dependent on the browser.

agalin920:
@shrunyan The way the tests were setup is was not getting preceded by our global blockAnnouncements call. Title of the PR is misleading will update

finnar-bin (approved):

geodem127 (approved):

PR #3565, opened 2025-06-02
AI Content Editor

agalin920:


src/shell/views/Shell/AIDrawer.tsx
```diff
@@ -0,0 +1,493 @@
...
+            type: response.type,
+            payload: {
+              refKey: response.payload.refKey,
+              value: response.payload.value,
+            },
+          });
+        }
+      });
+    } catch (error) {
+      console.error("Error parsing AI response", error);
```
finnar-bin, line 213: Remove comment if unused

finnar-bin (approved):

src/shell/views/Shell/AIDrawer.tsx
agalin920: Which comment?

finnar-bin: Oh I meant the console.error, unless that's actually needed and not just for debugging purposes

PR #3567, opened 2025-06-03
[#3549] - Content: Workflow Status - Switching between versions triggers multiple API requests

### LINKED ISSUE #3549: Content: Workflow Status - Switching between versions triggers multiple API requests.
## Description:

Bug in the add/remove status labels functionality for content items.
When switching between versions after performing these actions, multiple API requests are triggered—one for each version—which can result in labels being unexpectedly added or removed.

## Recording:

https://github.com/user-attachments/assets/a17be97a-ef74-4a5e-9083-d0233bbeb355

https://github.com/user-attachments/assets/8a380191-c797-4752-892b-e5899104fa84

## Steps to Reproduce:
- Create a status label with both add and remove privileges.
- Open any content item.
- Create multiple versions of that content item.
- Select a version and add or remove one or more status labels.
- Quickly switch to another version—either one without labels or one from which labels were just removed in the previous step.
- Observe the label behavior across versions.

## Expected Behavior:
- Only a single API request should be made per label action.
- Switching between versions should not trigger additional label-related API requests or alter label states.

## Actual Behavior:
- Multiple API requests are sent—one for each version during switching.
- Labels can be inadvertently added or removed due to these redundant requests.

## Impact:
- This causes label states to become inconsistent and unpredictable. Users may see labels reappear or disappear without intent, requiring them to reopen items to verify the current state. This behavior disrupts workflows and undermines trust in version management.

## Suggested Fix:
- Ensure that label actions (add/remove) trigger only a single API request.
- Refactor the version-switching logic to prevent redundant API calls that interfere with label states.

geodem127:
**RCA:**
- [x] The check !isEqual(labelsBeforeUpdate, activeLabels) depends on data.labels, which may change due to re-fetches.
- [x] A re-render with new data can trigger a false diff.
- [x] Doesn’t track the last confirmed saved state, so it can’t distinguish between user-initiated changes vs. external data updates.

**FIX:**
- [x] Track Last Saved State with useRef
- [x] Store the last confirmed API payload in a ref (lastSavedLabels) and compare against it instead of data.labels.
- [x] Only update lastSavedLabels.current after updateItemWorkflowStatus succeeds, preventing duplicate calls.

https://github.com/user-attachments/assets/80e75ca3-60a2-4889-836e-52a7a9836a64


src/apps/content-editor/src/app/views/ItemEdit/components/ItemEditHeader/VersionSelector/VersionItem.tsx
```diff
@@ -94,6 +95,7 @@ export const VersionItem = memo(
       const [activeLabels, setActiveLabels] = useState(
         data?.labels?.map((label) => label?.ZUID)?.filter((label) => !!label)
       );
+      const lastSavedLabels = useRef(activeLabels);
```
finnar-bin, line 98: Is there any reason why this is using a `useRef` instead of just `useState`?

geodem127: We use useRef to keep the last saved labels across renders without causing re-renders, which helps avoid unnecessary API calls during component switches.

finnar-bin (approved):

agalin920 (approved):

PR #3572, opened 2025-06-04
Feature/3524 preserve scroll position when editing block variants

### LINKED ISSUE #3524: Blocks: Preserve Scroll Position When Editing Block Variants
## Current behavior
Currently, when editing a block that has multiple variants (especially a large number of variants), users navigate to the bottom-right "Variants" section, scroll to find the desired variant, and click it to edit. 
Upon loading the new page for that specific variant, the page resets its scroll position to the top. This forces the user to manually scroll back down to the "Variants" section and re-locate the variant they just clicked, creating unnecessary friction and a cumbersome workflow.

## Proposed Solution
When editing a specific block variant, the page should scroll to and highlight that variant in the "Variants" section upon page load.

**Additional Info**
See video reference [here](https://zesty-io.slack.com/archives/C051VRB2PD2/p1747857432955079)

geodem127:
**Root Cause:**
- [ ] The` <Route>` was using the render() method, which caused the BlockItem component to remount whenever the /:itemZUID route param changed. This led to a full reset of the component's state.

**Fix:**
- [ ] Updated the route to render BlockItem as a child instead of using render(), preventing unnecessary remounts and preserving local state.
- [ ] Modified the useEffect to include itemZUID as a dependency, ensuring that the model data is correctly refreshed when the route changes.

https://github.com/user-attachments/assets/640f1afc-86dc-4e1b-85c1-c473c2a89240


agalin920 (approved):

src/apps/blocks/views/BlockItem.tsx
```diff
@@ -8,30 +8,15 @@ import { ItemCreate } from "../../content-editor/src/app/views/ItemCreate";
...
-        display="flex"
-        justifyContent={"center"}
-        alignItems={"center"}
-        height="100%"
-        width="100%"
-      >
-        <CircularProgress />
-      </Box>
-    );
-  }
```
finnar-bin, line 34: Why are we removing the loading indicator?

geodem127: It also triggers the component to remount when switching between the loader and the actual component. the preview and editor components already have their own loaders.

finnar-bin (approved):

PR #3586, opened 2025-06-11
[#3585] : Blocks - Newly-created variant doesn't get added to the variants list tab

### LINKED ISSUE #3585: Blocks - Newly-created variant doesn't get added to the variants list tab
## Problem

After creating a new block variant, the new variant doesn't get added to the variants list tab. I had to refresh the page in order to see it on the list. For additional context, this issue doesn't happen in prod yet and is only happening on stage and dev.

## Screenshots

https://github.com/user-attachments/assets/76c45b12-5837-4e3e-8700-7b7dbefa6a58

## Steps to reproduce

1. Go to blocks
2. Click on an existing block
3. Create a new variant

## Expected behavior

Newly-created block variants should be visible on the block variants list tab

### LINKED ISSUE #3596: Blocks - Save button on active preview saves changes for an incorrect variant
## Problem

When opening a block item and then viewing a different variant, once you make changes to the variant and clicking save on the active preview, it makes an api call to save the changes on the previously open variant and not the currently open variant. The issue right now only happens on dev and has not yet made it to prod.

Upon checking the network tab, I also saw that the PUT api call was being performed on the ZUID of the previously opened variant and not the currently opened variant. I also noticed that the screenshots on the variants picker are messed up due to this.

Image

## Screenshots

https://github.com/user-attachments/assets/fb4fa91a-e704-4993-859e-604db4f37900

## Steps to reproduce

1. Go to blocks
2. Open any block item with multiple variants
3. Open a different variants from the variant picker
4. Make some changes on the currently opened variant
5. Notification shows that the saved changes were for the first opened variant, checking the network tab also shows that the PUT action was done for the previously opened variant

## Expected behavior

Clicking save should save changes for the currently opened variant

### LINKED ISSUE #3597: Blocks - Duplicate block modal doesn't go away after successful duplication
## Problem

After successfully duplicating a block item the duplicate variant modal does not go away, instead it stays there and asks me if I want to duplicate again. After manually closing it, the duplicated variant does not get listed on the variants list. The issue right now only happens on dev and has not yet made it to prod.

## Screenshots

https://github.com/user-attachments/assets/fd1e77fe-2a20-4ed0-af94-397efd710b29

## Steps to reproduce

1. Open a block with a variant already created
2. Duplicate the variant
3. See error

## Expected behavior

The duplicate modal should automatically close upon successful duplication. The duplicated variant should also be listed on the variants list.

geodem127:
### RCA:
- A required prop was missing.
- The RTK Query refetch wasn’t triggered after dispatching the createItem Redux action.

### Fix:
- Added the missing prop to ensure the block preview updates correctly.
- Manually triggered a refetch of modelItems immediately after item creation.

https://github.com/user-attachments/assets/a5bc9dde-63af-4c80-bae1-d406257e55ba


src/apps/blocks/components/CreateVariantDialog.tsx
```diff
@@ -50,14 +54,16 @@ export const CreateVariantDialog = ({
...
+      dispatch(
+        createItem({
+          modelZUID: model.ZUID,
+          itemZUID: `new:${model.ZUID}`,
+          skipPathPartValidation: true,
+        })
+      )
     );
-
+    refetchModelItems();
```
finnar-bin: Instead of doing this here and potentially having to manually do this again for other use cases, I suggest you invalidate the tag inside the `createItem` redux action on success, so the rtk query cache gets busted every time that redux action succeeds.

finnar-bin (changes requested):

agalin920 (approved):

src/apps/blocks/components/CreateVariantDialog.tsx
geodem127: updated. suggestion inplemented

geodem127:
NOTE: This change also addresses the issue outlined in these tickets: https://github.com/zesty-io/manager-ui/issues/3597, https://github.com/zesty-io/manager-ui/issues/3596

finnar-bin (approved):

PR #3627, opened 2025-06-30
Resolve non-promise rejections from unhandled rtk query unwrap()

### LINKED ISSUE #3124: Non-Error promise rejection captured with keys: data, status
Sentry Issue: [MANAGER-UI-25B](https://zestyio.sentry.io/issues/4950799071/?referrer=github_integration)

```
UnhandledRejection: Non-Error promise rejection captured with keys: data, status
```

finnar-bin:
Unhandled errors from using `unwrap()` on an rtk query causes the non-error promise rejections since unwrap purposely rejects with plain objects on instead of an Error object. The solution is to make sure that when using `unwrap()` errors should be handled.


agalin920 (approved):

agalin920:
@finnar-bin nice find

PR #3649, opened 2025-07-05
[#3264]: Schema - Integration Field

### LINKED ISSUE #3264: Schema: Integration Field
# Intro
We are introducing a powerful new field in Schema called Integration Fields. With Integration Fields, content editors can now select and store data in their content teams from external data sources through a remote API. This enables seamless third-party integrations without the need for custom development.

## How it works

**In the Schema App:**
Users can now add an Integration Field to any model. When setting up the field, they will:
1. Specify the remote JSON API endpoint they want to pull data from. And configure any necessary authentication headers for secure access.
Image
2. Choose how the fetched API items should be displayed in the Content App (by selecting which JSON keys to use for display labels and values).
Image

**In the Content App:**
When a user edits a content item that includes an Integration Field, they will:
1. See a field that allows them to select items retrieved live from the remote API.

Image

2. Be able to select one or more items from the external data source and store those selections within their content item.

Image

This feature greatly enhances Zesty.io’s ability to integrate with third-party systems, external databases, or even other Zesty.io instances — all while keeping workflows simple for content editors.

# Figma
https://www.figma.com/design/ZlpuseFmUTEdZFVfaQuIAe/Integration-Field?node-id=69-9604&t=W5OT8ZMxVuBv9WFX-1

# Dev Notes
## Display Types
We will have a huge library of display types. These display types are. 

### Generic Display Types

1. Simple Card
2. Image Card
3. Video Card
4. Detail Card
5. Text Card

### Special Display Types
These are display types the user will only be able to select if the user enters the following JSON endpoints. 

1. MUX ([MUX API](https://www.mux.com/docs/api-reference/video/assets/list-assets))
2. Youtube (Youtube [videos API](https://developers.google.com/youtube/v3/docs/videos/list).)
3. Shopify ([Shopify Products API](https://shopify.dev/docs/api/ajax/reference/product))
4. Classy ([Classy API](https://docs.classy.org/#tag/Activity/operation/listCampaignActivities))

# Back End Notes
## Integration Field Definition

| Component                 | Details                                                                                                    |
| :------------------------ | :--------------------------------------------------------------------------------------------------------- |
| **Field Type** | Integration Field                                                                                          |
| **Field Key** | User-defined. Determines which key from the JSON object is displayed in relationship tags (e.g., `title`).   |
| **Authentication Headers**| Optional, user-defined key-value pairs. Supports Bearer tokens, Basic Auth, and API key authentication.    |
| **Remote JSON Endpoint** | URL defining the external data source. Must return a JSON array of flat objects with a consistent shape.   |
| **Validation Checks** | **Description** |
| Field Key Validation      | Ensure uniqueness within the content model and adherence to Zesty.io specifications.                       |
| Field Value Validation    | Verify that the specified Field Key exists in the JSON response.                                           |
| Authentication Header Valid.| Confirm headers are correctly formatted and functional.                                                    |
| Remote JSON Endpoint Valid. | Ensure the provided URL is valid and returns a structured JSON array.                                      |

## Database Requirements
- The referenced JSON object will be stored as an object in the content item table.
- API responses should return the full JSON object.
- Integration with Parsley must be documented for usage within Zesty.io templates.

# Implementation Steps
1. Extend the existing content model schema to include the "JSON Field" field type.
2. Develop backend logic for fetching JSON data, including authentication headers.
3. Update the Golang instances API to support the new field type and store JSON objects.
5. Integrate necessary validation checks for field keys, values, authentication, and endpoints.
6. Implement error handling with clear messaging for debugging.
7. Conduct extensive testing, including unit, integration, and end-to-end tests.
8. Update Zesty.io developer documentation with usage guidelines and examples.

geodem127:
- [x] Schema: Create Block Type Card - Field list item
- [x] Schema: Create Option Cards
- [x] Schema: Add "Integration Field" to the field type selection options, tooltip and form
- [x] Schema:  Create Display Type Options Selection Modal
- [x] Schema: Form Dialog (Connect to  API)
- [x] Schema: Form Dialog (Display Type Selection)
- [x] Schema: Form Dialog (Configure Display Options)
- [x] Schema: JSON view - Edit Dialog
- [x] Swap mock data with real data

---

- [x] Content:Item Selection Dialog
- [x] Content: Field View Components
- [x] Content: JSON view - Edit Dialog

---

Technical Specs: https://docs.google.com/document/d/1xynOhoDtj6TJB_tUrvx_rPncL4sIT156TdPd3qlDaek/edit?usp=sharing

### ADD FIELD:
---
https://github.com/user-attachments/assets/ff4fe888-c77e-42eb-9ee0-63c9468c99b7

### SELECT ITEMS:
---
https://github.com/user-attachments/assets/19427df8-5dca-4d1f-8240-c46c5334b222


cypress/fixtures/integration-field/fields.json
```diff
@@ -0,0 +1,68 @@
+{
```
finnar-bin: I'm assuming this is just dummy data for the tests since BE is not yet done?

src/apps/content-editor/src/app/views/ItemEdit/Meta/ContentInsights/index.tsx
```diff
@@ -218,7 +218,9 @@ export const ContentInsights = ({}) => {
         if (!!value) {
           value = cleanContent(String(value));
 
-          words = [...words, ...value.split(" ")];
+          if (typeof value === "string") {
```
finnar-bin, line 224: How is this change relevant to the integration field?

src/apps/schema/src/app/components/AddFieldModal/InputRange.tsx
```diff
@@ -54,15 +57,20 @@ export const InputRange = ({
         label={
           <Box>
             <Typography variant="body2" fontWeight="600">
-              Limit Input Range
+              {type === "integration"
```
finnar-bin: I suggest adding a new prop that accepts custom primary and secondary text instead of determining the text depending on the type. This makes it more flexible in case we need to add more custom description text for this component in the future.

src/apps/schema/src/app/components/AddFieldModal/InputRange.tsx
```diff
@@ -54,15 +57,20 @@ export const InputRange = ({
...
+                : "Limit Input Range"}
             </Typography>
             <Typography
               variant="body3"
               color="text.secondary"
               fontWeight="600"
               display="block"
             >
-              Set a minimum and/or maximum allowed value
+              {/* Set a minimum and/or maximum allowed value */}
```
finnar-bin: Remove if unused

src/apps/schema/src/app/components/AddFieldModal/views/FieldForm.tsx
```diff
@@ -823,7 +855,7 @@ export const FieldForm = ({
           borderTop: 0,
         }}
       >
-        {activeTab === "details" && (
+        <div role="tabpanel" hidden={activeTab !== "details"}>
```
finnar-bin: Wouldn't this make the all 3 tabs be mounted the whole time? This was purposely added so that only the active tab will be mounted, we only want to mount the relevant active tab to improve performance. Same comments as the other 2 below.

src/shell/components/FieldTypeIntegration/ConfigureIntegration.tsx
```diff
@@ -0,0 +1,207 @@
...
+            <Typography
+              width={170}
+              variant="body2"
+              fontWeight={600}
+              flexGrow={0}
+              flexShrink={0}
+            >
+              Display Items as
+            </Typography>
+            <InputBase
```
finnar-bin: Why not just use a TextField here?

src/shell/components/FieldTypeIntegration/DisplayCard/Details.tsx
```diff
@@ -0,0 +1,143 @@
...
+type DetailsProps = {
+  data: any[];
+  listItems: string[];
+  loading?: boolean;
+};
+
+const DetailsSkeleton = () => {
+  return (
+    <>
+      <Stack
```
finnar-bin: I suggest just creating an array and looping via map to render the skeleton components as they all seem to be the same repeated code.

src/shell/components/FieldTypeIntegration/DisplayCard/Details.tsx
```diff
@@ -0,0 +1,143 @@
...
+  );
+};
+
+const Details: FC<DetailsProps> = ({ data, listItems, loading = false }) => {
+  if (!listItems?.length)
+    return (
+      <Typography variant="body2" color="text.secondary">
+        + Add Detail
+      </Typography>
+    );
```
finnar-bin: Wrap return in { }

src/shell/components/FieldTypeIntegration/DisplayCard/Details.tsx
```diff
@@ -0,0 +1,143 @@
...
+          animation="wave"
+          height="10px"
+          width="50px"
+        />
+      </Stack>
+    </>
+  );
+};
+
+const Details: FC<DetailsProps> = ({ data, listItems, loading = false }) => {
```
finnar-bin: Minor nitpick but prefer to use types instead of passing as a generic prop in FC for consistency

src/shell/components/FieldTypeIntegration/DisplayCard/DraggableCard.tsx
```diff
@@ -0,0 +1,250 @@
...
+      className="draggableCard"
+      id={id}
+      elevation={0}
+      ref={!draggable ? null : (node) => drop(preview(node as HTMLElement))}
+      sx={{
+        zIndex: isDragging ? 1 : 0,
+        opacity: isDragging ? 0.1 : 1,
+        py: 0,
+        pl: "28px",
+        pr: "40px",
```
finnar-bin: Use theme values for paddings & margins

src/shell/components/FieldTypeIntegration/DisplayCard/MediaThumbnail.tsx
```diff
@@ -0,0 +1,78 @@
...
+      <AddPhotoAlternateRoundedIcon
+        className="media-thumbnail-icon"
+        sx={{ color: "grey.400" }}
+      />
+    ) : (
+      <VideoCallRoundedIcon
+        className="media-thumbnail-icon"
+        sx={{ color: "grey.400" }}
+      />
+    );
```
finnar-bin: I would rather we change this to a proper if-else statement for better readability, nested ternary operations can get out of hand real quick.

src/shell/components/FieldTypeIntegration/DisplayCard/MediaThumbnail.tsx
```diff
@@ -0,0 +1,78 @@
...
+        position: "relative",
+        display: "flex",
+        placeContent: "center",
+        alignItems: "center",
+        "& .media-thumbnail-icon": {
+          position: "absolute",
+          top: "50%",
+          left: "50%",
+          transform: "translate(-50%, -50%)",
+        },
```
finnar-bin: alignItems: "center" and justifyContent: "center" should also be able to center the icon

src/shell/components/FieldTypeIntegration/DisplayCard/SelectCard.tsx
```diff
@@ -0,0 +1,164 @@
...
+        alignItems: "center",
+        borderBottom: "1px solid",
+        borderColor: "border",
+        boxShadow: "none",
+        backgroundColor: "background.paper",
+
+        "&.select-card": {
+          backgroundColor: (theme) => alpha(theme.palette.primary.light, 0.04),
+          boxShadow: (theme) =>
+            `0px -2px 0px 0px ${theme.palette.primary.light} inset`,
```
finnar-bin: I would advise avoiding theme overrides like this for consistency, can't this be achieved by changing the elevation if selected?

src/shell/components/FieldTypeIntegration/DisplayCard/SelectCard.tsx
```diff
@@ -0,0 +1,164 @@
...
+          placeContent: "center",
+        }}
+      >
+        {loading ? (
+          <Skeleton
+            animation="wave"
+            variant="rounded"
+            height="20px"
+            width="20px"
+            sx={{ m: "8px" }}
```
finnar-bin: Use theme values for margins and paddings

src/shell/components/FieldTypeIntegration/DisplayCard/SelectCard.tsx
```diff
@@ -0,0 +1,164 @@
...
+
+      <Box
+        width="100%"
+        height="100%"
+        position="relative"
+        boxSizing="border-box"
+        overflow="hidden"
+        sx={{
+          "& .media-thumbnail": {
+            borderRadius: "8px",
```
finnar-bin: Same here, use theme values

src/shell/components/FieldTypeIntegration/DisplayCard/index.tsx
```diff
@@ -0,0 +1,179 @@
...
+              variant="body2"
+              fontWeight={400}
+              color="text.secondary"
+              noWrap
+              textOverflow="ellipsis"
+              width="100%"
+            >
+              {subHeading || "Add Sub-heading"}
+            </Typography>
+          )}
```
finnar-bin: Try to avoid nested ternary operations as much as possible for code readability

src/shell/components/FieldTypeIntegration/SelectItems.tsx
```diff
@@ -0,0 +1,219 @@
...
+  return (
+    <>
+      <Box
+        sx={{
+          display: "flex",
+          flexDirection: "column",
+          justifyContent: "flex-start",
+          alignItems: "flex-start",
+          width: "100%",
+          borderRadius: "8px",
```
finnar-bin: Use theme values

src/shell/components/FieldTypeIntegration/configs.ts
```diff
@@ -0,0 +1,227 @@
...
+//   },
+//   {
+//     key: "",
+//     value: "",
+//   },
+//   {
+//     key: "",
+//     value: "",
+//   },
+// ];
```
finnar-bin: Clean up if unused

src/shell/components/FieldTypeIntegration/forms/IntegrationForm/ConfigureDisplayOptions.tsx
```diff
@@ -0,0 +1,471 @@
+import { useEffect, useState, useRef } from "react";
+import Button from "@mui/material/Button";
+import DialogActions from "@mui/material/DialogActions";
+import DialogContent from "@mui/material/DialogContent";
+import DialogTitle from "@mui/material/DialogTitle";
+import { Box, Stack, Typography } from "@mui/material";
+import CloseRoundedIcon from "@mui/icons-material/CloseRounded";
+import IconButton from "@mui/material/IconButton";
```
finnar-bin: I suggest picking either a barrel or direct import to keep things consistent

src/shell/components/FieldTypeIntegration/forms/IntegrationForm/ConfigureDisplayOptions.tsx
```diff
@@ -0,0 +1,471 @@
...
+          alignItems: "stretch",
+          rowGap: 2,
+          overflowY: "auto",
+          overflowX: "hidden",
+          flexGrow: 1,
+          position: "relative",
+          "&::-webkit-scrollbar-thumb, & *::-webkit-scrollbar-thumb": {
+            backgroundColor: "grey.300",
+            borderRadius: "4px",
+          },
```
finnar-bin: Why are we overriding the scrollbar styling here?

src/shell/components/FieldTypeIntegration/forms/IntegrationForm/SelectDisplayOptions.tsx
```diff
@@ -0,0 +1,443 @@
...
+      data-cy={`integrationDisplayOption-${type}`}
+      elevation={0}
+      variant="outlined"
+      role="button"
+      className={isSelected ? "active" : "free-card"}
+      sx={{
+        position: "relative",
+        width: "100%",
+        bgcolor: "grey.100",
+        borderRadius: "4px",
```
finnar-bin: Use theme values

src/shell/components/FieldTypeIntegration/forms/SelectionForm/JsonViewer.tsx
```diff
@@ -0,0 +1,156 @@
...
+              wordWrap: "on",
+              minimap: {
+                enabled: false,
+              },
+              padding: {
+                top: 20,
+                bottom: 20,
+              },
+            }}
+            value={!data ? "" : JSON.stringify(data, null, 2)}
```
finnar-bin: Might be worth wrapping this in a try-catch in case data is malformed to prevent app from crashing

src/shell/components/FieldTypeIntegration/forms/SelectionForm/index.tsx
```diff
@@ -0,0 +1,338 @@
...
+      open={formSelectorOpen}
+      onClose={() => setFormSelectorOpen(false)}
+      slotProps={{
+        paper: {
+          style: {
+            maxWidth: "800px",
+            minHeight: "860px",
+            height: "calc(100vh - 40px)",
+            maxHeight: "1240px",
+            margin: "20px",
```
finnar-bin: Use theme values

src/shell/components/FieldTypeIntegration/forms/SelectionForm/index.tsx
```diff
@@ -0,0 +1,338 @@
...
+          />
+        </Box>
+        <Paper
+          data-cy="integrationSelectionFormListContainer"
+          elevation={0}
+          variant="outlined"
+          sx={{
+            width: "100%",
+            minHeight: `calc(100% - 70px)`,
+            borderRadius: "8px",
```
finnar-bin: Use theme values

src/shell/components/FieldTypeIntegration/forms/SelectionForm/index.tsx
```diff
@@ -0,0 +1,338 @@
...
+            overflow: "hidden",
+            boxSizing: "border-box",
+            position: "relative",
+            borderColor: "border",
+            p: 0,
+            "& .list-container > div": {
+              padding: "0 32px 16px 32px",
```
finnar-bin: Theme values

src/shell/components/FieldTypeIntegration/forms/SelectionForm/index.tsx
```diff
@@ -0,0 +1,338 @@
...
+                      }
+                    });
+                  }}
+                  onViewJson={() => openViewer(item)}
+                  loading={isFetching}
+                  disabled={maxItemsSelected}
+                />
+              ))}
+            </>
+          )}
```
finnar-bin: Suggest extracting a function that uses an if-else statement for readability instead of nested ternary operations

finnar-bin (changes requested):

src/shell/components/FieldTypeIntegration/IntegrationFieldProvider.tsx
agalin920, line 1: Big request change but I am not a fan large complex providers to handle slices of app state and there is a reason we don't have any in the codebase (other than the global library based level ones)

Context is best used for sharing global, stable state like theming, authentication status, or user locale. Frequent or complex state updates (like those in this provider) cause every consumer to re-render, often unnecessarily this can lead to performance bottlenecks and harder-to-track re-renders.

Placing lots of local/ephemeral state into context hides architectural problems rather than solving them. Better alternatives are to colocate state in the components that need it most an using custom hooks for reusable logic. If truely needed we should leverage RTK that employs patterns that are more scalable

A massive provider like this becomes a god object too many unrelated responsibilities, hard to test in isolation, and hard to reason about. Changes or bugs in one part of the context can unintentionally affect unrelated consumers.

Let me know if there are any other thoughts on this @finnar-bin @shrunyan

agalin920 (changes requested):

src/shell/components/FieldTypeIntegration/IntegrationFieldProvider.tsx
geodem127: @agalin920 
The context's scope is strictly limited to the Integration Field component, with all effects and side effects contained within the component itself. Its sole external interaction is through the setValue function, which updates the Integration Field's formData value.

src/apps/content-editor/src/app/views/ItemEdit/Meta/ContentInsights/index.tsx
geodem127: Resolved - deleted the lingering 'unknown' type that was meant as a temporary placeholder.

src/apps/schema/src/app/components/AddFieldModal/InputRange.tsx
geodem127: updated. implemented suggested

src/shell/components/FieldTypeIntegration/ConfigureIntegration.tsx
geodem127: TextField brings unnecessary default styles we'd have to override

src/shell/components/FieldTypeIntegration/DisplayCard/Details.tsx
geodem127: they have different widths

src/shell/components/FieldTypeIntegration/IntegrationFieldProvider.tsx
finnar-bin: I agree with @agalin920 point regarding avoiding using large context providers like this to avoid bugs and make it easier to debug, hovever I also do see that @geodem127 's context is solely being used for the integration field. I do however think it does hold a lot of functionality though, is everything on the context being used on different components within the integration field? Kinda curious since maybe we could find a way to move some of them locally to the component using it.

geodem127: We could probably move some of the state to local components, but since the form depends on input values from previous steps, we’d still need to pass that state down as props eventually. I find it easier to scope the state within the parent form component and its children so that all steps have access to the shared values, regardless of where the user is in the flow. This ensures a single source of truth. The context is only used to store the values—calculations are handled within the individual form components.

In terms of performance, only the currently mounted form step is affected. Changes made within the current form are not committed to the context until the user clicks “Next” (unless I missed any cases where this might not be true).

I’ve also reduced the amount of state held in the context by removing some unnecessary values.

src/shell/components/FieldTypeIntegration/DisplayCard/SelectCard.tsx
geodem127: I see your point about theme consistency.Just to clarify, we're not overriding theme settings —we're using box-shadow to create a 2px bottom-border effect. basically similar with "borderBottom: 2px solid color" but without altering the box-models' dimension and position and it works well with MUI's ripple effect (since elevation uses box-shadow implementation).

src/shell/components/FieldTypeIntegration/IntegrationFieldProvider.tsx
agalin920: @geodem127 While there are pros/cons it all comes down to we don't use react context as a pattern in our codebase. We rely on component structure to logically separate and manage state along with RTK query to manage server state and redux to manage state that has a very large footprint across the app

src/shell/components/FieldTypeIntegration/utils.ts
```diff
@@ -0,0 +1,160 @@
...
+      } else if (typeof value === "object" && value !== null) {
+        result.push(...getAllArrayKeyPaths(value, currentKey));
+      }
+    }
+  }
+
+  return result;
+}
+
+function getKeyValue<T, K extends string>(obj: T, path: K): any {
```
agalin920: A couple of utils seem like core utils we should move this higher or better yet rely on a core util package ideally one we already have.

Some of these other functions seem like they can live within specific components as I know our team is not a big fan of util barrel files inside feature sets unless they are truly shared and feature specific

src/shell/components/FieldTypeIntegration/forms/JsonViewer.tsx
```diff
@@ -0,0 +1,165 @@
...
+      component={isSlider ? Slide : "div"}
+      {...(isSlider
+        ? { in: open, ...SLIDER_PROPS }
+        : { width: "100%", height: "100%" })}
+    >
+      <Paper
+        elevation={0}
+        variant="outlined"
+        sx={{
+          zIndex: 5,
```
agalin920: Why 5? this seems arbitrary. Mui provides some semantic zIndex values

src/shell/components/FieldTypeIntegration/useIntegrationField.ts
```diff
@@ -0,0 +1,116 @@
...
+const useIntegrationField = (initialConfig: IntegrationFieldConfig | null) => {
+  const [config, setConfig] = useState(initialConfig);
+  const [apiData, setApiData] = useState<any>(null);
+  const [selectedItems, setSelectedItems] = useState<any[]>([]);
+  const [isLoading, setIsLoading] = useState(false);
+  const [error, setError] = useState<string | null>(null);
+
+  const { endpoint, headers, type, keyPaths } = config;
+
+  const generateItemId = useCallback(
```
agalin920: What is this for?

src/shell/components/FieldTypeIntegration/useIntegrationField.ts
```diff
@@ -0,0 +1,116 @@
...
+          throw new Error("Failed to fetch data from API");
+        }
+      } catch (err) {
+        setError(err.message || "Failed to fetch data");
+        setApiData(null);
+      }
+
+      setTimeout(() => {
+        setIsLoading(false);
+      }, 2000);
```
agalin920: This forces loading to be two seconds? what happens if the request takes longer or takes less time? Any reason we dont do it based on the actual call?

src/shell/components/FieldTypeIntegration/forms/ConnectToApi.tsx
```diff
@@ -0,0 +1,387 @@
...
+  const [headersLocal, setHeadersLocal] = useState<
+    { key: string; value: string }[] | null
+  >(keyValuePairsToArray(headers));
+
+  const handleApiConnect = useCallback(async () => {
+    setReqAborted(false);
+    setStatus("connecting");
+
+    const headersWithValues = headersLocal.filter(
+      (i) => i.value !== "" && !i.value
```
finnar-bin: This seems redundant since `!a.value` also checks empty strings

src/apps/schema/src/app/components/AddFieldModal/views/FieldForm.tsx
finnar-bin: @geodem127 is there any reason why this is being hidden instead of unmounted when not active?

finnar-bin (changes requested):

src/shell/components/FieldTypeIntegration/useIntegrationField.ts
geodem127: It is use to generate unique ID for keeping track of the selected/reordered items since we dont have a fixed API data structure when rendering the list items.

src/shell/components/FieldTypeIntegration/useIntegrationField.ts
geodem127: This just for testing purposes. Sorry forgot to remove it

finnar-bin (approved):

src/shell/components/FieldTypeIntegration/utils.ts
```diff
@@ -0,0 +1,50 @@
...
+    return [part];
+  });
+
+  return keys.reduce((acc: any, key) => {
+    if (acc === null || acc === undefined) return undefined;
+    return acc[key];
+  }, obj);
+}
+
+export const fetchApi = async <T = unknown>({
```
agalin920: I question the value of this abstraction, it’s mildly useful at best, but it introduces some questionable design choices that might hurt more than help.

Its a redundant abstraction and you lose flexibility without gaining much.

agalin920 (changes requested):

src/shell/components/FieldTypeIntegration/useIntegrationField.ts
agalin920: If this is really needed you could just use uuid library that we have instead of this complex logic. 

Although I would ask you to question this further and see if this is really needed or if it can all be index based

agalin920 (changes requested):

src/shell/components/FieldTypeIntegration/forms/Wrappers.tsx
```diff
@@ -0,0 +1,111 @@
...
+export const FieldWrapper = ({
+  name,
+  label,
+  description,
+  toolTip,
+  isRequired,
+  error,
+  children,
+}: {
+  name?: string;
```
agalin920: Use your standard prop typing pattern. FieldWrapperProps

src/shell/components/FieldTypeIntegration/forms/SelectDisplayOptions.tsx
```diff
@@ -0,0 +1,493 @@
...
+
+const SelectDisplayOptions = ({
+  activeStep,
+  setActiveStep,
+  endpoint,
+  type,
+  setType,
+  closeForm,
+}: {
+  activeStep: number;
```
agalin920: Same here regarding prop typing

agalin920 (changes requested):

agalin920 (approved):

src/shell/components/FieldTypeIntegration/IntegrationFieldConfigure/SelectDisplayOptions.tsx
```diff
@@ -0,0 +1,309 @@
...
+          display: "flex",
+          flexDirection: "column",
+          justifyContent: "space-between",
+          alignItems: "stretch",
+          rowGap: 2,
+          overflowY: "auto",
+          overflowX: "hidden",
+          flexGrow: 1,
+          position: "relative",
+          "&::-webkit-scrollbar-thumb, & *::-webkit-scrollbar-thumb": {
```
agalin920: Don't think we want specific scroll bar styles just for this specific dialog

agalin920 (changes requested):

src/shell/components/FieldTypeIntegration/IntegrationFieldConfigure/SelectDisplayOptions.tsx
geodem127: Got it - I've removed the scrollbar styling.

.nyc_output/out.json
finnar-bin, line 1: Not sure if we'd want to commit this file

src/apps/content-editor/src/app/views/ItemList/ItemListTable.tsx
```diff
@@ -256,6 +256,25 @@ const fieldTypeColumnConfigMap = {
...
+        <Box
+          height="100%"
+          width="100%"
+          display="flex"
+          flexDirection="row"
+          justifyContent="flex-start"
+          alignItems="center"
+        >
+          <Typography variant="body2" noWrap>
+            {params.value?.toUpperCase()}
```
finnar-bin: Is this rendering the entire JSON value as a string in the multipage table?

src/shell/components/FieldTypeIntegration/components/DisplayCard.tsx
```diff
@@ -0,0 +1,345 @@
...
+    }
+
+    return (
+      <>
+        {withCardMedia && (
+          <Box
+            sx={{
+              borderRadius: mediaVariant === "rounded" ? 2 : 0,
+              height: "80px",
+              width: isVideoType ? "142px" : "80px",
```
finnar-bin, line 194: 142px seems like a very specific width value. Is this what was asked for in the design? Is this supposed to stay the same even if the window width has increased decreased? Just wanted to make sure the layout isn't breaking.

src/shell/components/FieldTypeIntegration/components/JsonViewer.tsx
```diff
@@ -0,0 +1,198 @@
...
+      <Box
+        borderRadius={2}
+        sx={{
+          height: "calc(100% - 84px)",
+          width: "100%",
+          overflow: "auto",
+          bgcolor: "grey.50",
+          py: 2,
+          px: 4,
+          "& .react-monaco-editor-container": {
```
finnar-bin, line 145: Isn't it possible to pass these monaco editor styles into the monaco editor component itself instead of declaring these on the parent?

src/shell/components/FieldTypeIntegration/IntegrationFieldConfigure/ConfigureDisplayOptions.tsx
```diff
@@ -0,0 +1,450 @@
...
+  }, [displayConfig, rootPathData, detailsPathData, rootPath]);
+
+  return (
+    <FormWrapper height="calc(100vh - 40px)" width="1200px">
+      <DialogTitle component="div" flexGrow={0}>
+        <Settings
+          color="primary"
+          sx={{
+            p: 1,
+            borderRadius: "20px",
```
finnar-bin: Use theme values

src/shell/components/FieldTypeIntegration/IntegrationFieldConfigure/ConfigureDisplayOptions.tsx
```diff
@@ -0,0 +1,450 @@
...
+                }))}
+                showPlayIcon={false}
+              />
+              <MoreHoriz color="action" sx={{ m: 1.75 }} />
+            </Paper>
+          </Box>
+        </Box>
+      </DialogContent>
+
+      <DialogActions sx={{ p: 2, height: 76, minHeight: 76 }}>
```
finnar-bin: Why are we doing a one-time override for this?

src/shell/components/FieldTypeIntegration/IntegrationFieldConfigure/ConnectToApi.tsx
```diff
@@ -0,0 +1,429 @@
...
+  const [headersLocal, setHeadersLocal] = useState<
+    Record<string, { key: string; value: string }>
+  >(() => {
+    if (!headers || Object.keys(headers).length === 0) {
+      const id = Date.now().toString();
+      return { [id]: { key: "", value: "" } };
+    }
+
+    return Object.entries(headers).reduce((acc, [key, value]) => {
+      const id = Date.now() + Math.random().toString(36).substr(2, 9);
```
finnar-bin: Why not just use the browser's built in [randomUUID](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID) method from the Crypto interface?

src/shell/components/FieldTypeIntegration/IntegrationFieldConfigure/ConnectToApi.tsx
```diff
@@ -0,0 +1,429 @@
...
+  }, [endpointLocal, headersLocal]);
+
+  return (
+    <FormWrapper width="480px" height="fit-content">
+      <DialogTitle component="div">
+        <DataObjectRoundedIcon
+          color="primary"
+          sx={{
+            padding: 1,
+            borderRadius: "20px",
```
finnar-bin: Use theme values

src/shell/components/FieldTypeIntegration/IntegrationFieldConfigure/ConnectToApi.tsx
```diff
@@ -0,0 +1,429 @@
...
+          Establish a connection to an endpoint for users to select items from
+        </Typography>
+        <Box
+          display="flex"
+          flexDirection="row"
+          columnGap={1}
+          justifyContent="flex-start"
+          alignItems="center"
+          component={Link}
+          href="#"
```
finnar-bin: Is this supposed to just be an empty link?

src/shell/components/FieldTypeIntegration/IntegrationFieldConfigure/DisplayOption.tsx
```diff
@@ -0,0 +1,133 @@
...
+                height="100%"
+                direction="column"
+                spacing={0.5}
+                p={2}
+                sx={{
+                  justifyContent: "center",
+                  alignItems: "flex-start",
+                }}
+              >
+                {["shopify", "youtube", "mux", "classy"]?.includes(type) && (
```
finnar-bin: Don't need to add the optional chaining here since the array is always going to present

src/shell/components/FieldTypeIntegration/IntegrationFieldConfigure/KeyPathSelector.tsx
```diff
@@ -0,0 +1,130 @@
...
+                  borderRadius: 1,
+                  fontSize: "12px",
+                  bgcolor: `${typeColor}.50`,
+                  color: `${typeColor}.600`,
+                  py: 0,
+                  px: 1,
+                }}
+              >
+                {valueType}
+              </Box>
```
finnar-bin: Why not wrap {valueType} in Typography?

src/shell/components/FieldTypeIntegration/IntegrationFieldConfigure/SelectDisplayOptions.tsx
```diff
@@ -0,0 +1,305 @@
...
+      </DialogContent>
+      <DialogActions
+        sx={{
+          p: "20px",
+          flexGrow: 0,
+          height: "76px",
+          minHeight: "76px",
+          maxHeight: "76px",
+          borderTop: "1px solid",
+          borderColor: "border",
```
finnar-bin: Any reason why we're making one-time overrides for this? Also, if it's really needed prefer to use theme values instead

finnar-bin (changes requested):

src/apps/content-editor/src/app/views/ItemList/ItemListTable.tsx
geodem127: No, it's not rendering the entire JSON value as a string in the multipage table. It is displayed as a string (as required), but the value is truncated based on the cell width.

.nyc_output/out.json
geodem127: yup. this has been removed

src/shell/components/FieldTypeIntegration/components/DisplayCard.tsx
geodem127: Yes, it should stay the same — only the heading and details adjust when the screen width changes.

src/shell/components/FieldTypeIntegration/components/JsonViewer.tsx
geodem127: Yes, but I haven't seen any simpler way than this. Targeting the editor's internal classes from a parent container is actually one of the recommended approaches. We could use custom classes, but that would be more complicated, especially for a read-only viewer.

src/shell/components/FieldTypeIntegration/IntegrationFieldConfigure/ConfigureDisplayOptions.tsx
geodem127: updated

src/shell/components/FieldTypeIntegration/IntegrationFieldConfigure/ConnectToApi.tsx
geodem127: updated. used the uuid lib instead

src/shell/components/FieldTypeIntegration/IntegrationFieldConfigure/DisplayOption.tsx
geodem127: agree. this has been removed

src/shell/components/FieldTypeIntegration/IntegrationFieldConfigure/KeyPathSelector.tsx
geodem127: agree. this has been updated

src/shell/components/FieldTypeIntegration/IntegrationFieldConfigure/SelectDisplayOptions.tsx
geodem127: I wanted to make sure the height was set properly, but after double-checking, I realized there's no need to set the height or padding, so I removed them. The borders, however, remain as per the design.

src/shell/components/FieldTypeIntegration/IntegrationFieldConfigure/DisplayOption.tsx
finnar-bin: I can still see the optional chaining though

src/shell/components/FieldTypeIntegration/IntegrationFieldConfigure/ConfigureDisplayOptions.tsx
finnar-bin: I don't see the changes

src/shell/components/FieldTypeIntegration/IntegrationFieldConfigure/KeyPathSelector.tsx
finnar-bin: Don't see the changes

src/shell/components/FieldTypeIntegration/IntegrationFieldConfigure/SelectDisplayOptions.tsx
finnar-bin: Same here, don't see the changes as well

src/shell/components/FieldTypeIntegration/IntegrationFieldConfigure/ConfigureDisplayOptions.tsx
geodem127: this has been updated. turns out there is no need to set the hight

src/shell/components/FieldTypeIntegration/IntegrationFieldConfigure/ConnectToApi.tsx
geodem127: updated

src/shell/components/FieldTypeIntegration/IntegrationFieldConfigure/ConnectToApi.tsx
geodem127: link updated

src/utility/validateUrl.ts
agalin920, line 1: We've been moving away from a top-level src/utility/ folder. our orgs's view is that "utility" folders tend to become dumping grounds and obscure where logic actually lives. Could we colocate validateUrl next to the feature that consumes it? If it's genuinely cross-cutting, something with a clear scope tends to age better than a generic utility/ bucket.

agalin920 (changes requested):

src/utility/validateUrl.ts
geodem127: updated

finnar-bin (approved):

agalin920 (approved):

PR #3655, opened 2025-07-10
[Global] Use RichTreeView for in-app sidebar and several UI bug fixes

### LINKED ISSUE #3583: Bug: Sidebar List Items – Incorrect Padding, Icon Gaps, and Hover/Selected States Not Full Width
## Problem
**There are UI inconsistencies in the app sidebar component:**
- List items have incorrect padding, resulting in awkward spacing.
- Icon gaps are inconsistent or misaligned with the text.
- Hover and selected states do not extend from edge to edge, breaking visual continuity.

## Screenshots

Image
Image

## Expected behavior
- Sidebar items should have consistent and visually balanced padding.
- Hover/selected states should extend from edge to edge, covering the entire sidebar item area.

Image

## Suggested Fix:
Adjust sidebar list item styles:
- Use consistent padding/margin across items (hover/selected state).
- Fix icon alignment/gap.

### LINKED ISSUE #3628: [Media | Content] Hovering over in-app sidebar items flickers button bg color
## Problem

Hovering over items on the in-app sidebar on the Media and Content app causes some buttons to flicker from an orange background to a transparent background.

## Screenshots

https://github.com/user-attachments/assets/e1fb0a22-5b4d-4659-9f3e-82cde2dd6001

https://github.com/user-attachments/assets/36e6ccf7-80fe-48e1-a08e-225d9b73aff1

## Steps to reproduce

1. Go to media or content
2. Hover over some items on the in-app sidebar
3. See error

## Expected behavior

The buttons should not flicker from an orange to a transparent background,  it should stay orange

### LINKED ISSUE #3631: Content - Minimize in-app sidebar re-renders when opening/closing modals
## Problem

The in-app sidebar are re-rendering multiple times whenever some modals are being closed/opened even if the modal has nothing to do with the component

## Solution

- [x] Optimize component so it only re-renders if it's actually necessary
- [x] Migrate from SimpleTreeView to RichTreeView as this is more performant for large trees akin to a file explorer which is kind of what we have
- [ ] Add tests for all sidebars

finnar-bin:
- Migrate to `RichTreeView` in the in-app sidebar component from `SimpleTreeView`
  - Resolves #3631 
- Several UI bug fixes for the tree items
  - Fixes #3628 
  - Fixes #3583 

[Launchpad---Zesty-io---zesty-pw---Manager.webm](https://github.com/user-attachments/assets/0b48b2d3-f63b-409f-904f-039b19cf7c91)


src/shell/components/NavTree/index.tsx
```diff
@@ -120,29 +140,9 @@ export const NavTree: FC<Readonly<Props>> = ({
               onToggleCollapse(nodeIds);
```
geodem127: We can probably drop the if check — `expansionTrigger` is already set.

finnar-bin: Removed on latest commit

src/shell/components/NavTree/components/RichTreeItem.tsx
```diff
@@ -0,0 +1,120 @@
...
+
+    return (
+      <TreeItem2
+        {...props}
+        children={props.children}
+        ref={ref}
+        slots={{ label: NavTreeLabel }}
+        slotProps={
+          {
+            content: {
```
geodem127, line 42: The content component within the TreeViewItem must have its id set to modelZUID, as the menu popup uses it as its container. This helps maintain the hover state while the menu is open.

agalin920 (approved):

src/shell/components/NavTree/components/RichTreeItem.tsx
finnar-bin: @geodem127 I've added the same functionality to retain the hover status when opening the more menu for the content, blocks and media app. I went with a different approach from your original fix as setting the hover id on state causes the entire tree to re-render every time an item is hovered which severely affected performance.

geodem127: I’m not sure how the re-render on hover is related to the ID, especially since the ID used is modelZUID, which stays constant throughout the component’s lifecycle. It’s solely used to target the container for the menu popup. That said, if it improves performance, then that’s definitely a good outcome.

geodem127 (approved):

src/shell/components/NavTree/components/RichTreeItem.tsx
finnar-bin: Oh what I meant was that the hovered ID was being saved to state. With the rich tree view, this caused the entire tree to re-render whenever setHoverId was updated when hovering through items.

PR #3682, opened 2025-07-28
Dispose monaco model on unmount

agalin920:
Here we can see TextModel is being retained by monaco even after code app unmount. We have x38 retained models in this example
Screenshot 2025-07-28 at 12 47 22 PM
I also exposed monaco object to the window to more easily visualize the models that are retained
Screenshot 2025-07-28 at 12 42 50 PM

After the disposal code update we can observe the TextModels are no longer retained
Screenshot 2025-07-28 at 12 52 21 PM

Requesting the models from console also now shows no retained models
Screenshot 2025-07-28 at 12 48 46 PM

### DIFF
--- a/src/apps/code-editor/src/app/components/Editor/components/MemoizedEditor/MemoizedEditor.js
+++ b/src/apps/code-editor/src/app/components/Editor/components/MemoizedEditor/MemoizedEditor.js
@@ -29,9 +29,12 @@ export const MemoizedEditor = memo(
           // attach contentModelZUID to monaco model for lookup within completion provider
           query: `contentModelZUID=${props.contentModelZUID}&fileZUID=${props.fileZUID}`,
         });
-        const model =
-          monaco.editor.getModel(filenameURI) ||
-          monaco.editor.createModel(props.code, language, filenameURI);
+        let model = monaco.editor.getModel(filenameURI);
+        const didCreate = !model;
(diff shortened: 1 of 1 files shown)


src/apps/code-editor/src/app/components/Editor/components/MemoizedEditor/MemoizedEditor.js
```diff
@@ -29,9 +29,12 @@ export const MemoizedEditor = memo(
           // attach contentModelZUID to monaco model for lookup within completion provider
           query: `contentModelZUID=${props.contentModelZUID}&fileZUID=${props.fileZUID}`,
         });
-        const model =
-          monaco.editor.getModel(filenameURI) ||
-          monaco.editor.createModel(props.code, language, filenameURI);
+        let model = monaco.editor.getModel(filenameURI);
+        const didCreate = !model;
```
shrunyan: Any purpose to setting the variable vs just re-checking `!model` down on line 59?

agalin920: @shrunyan  No purpose other than some semantic reasoning for its purpose, but it is redundant and unnecessary so I have removed

finnar-bin (approved):

PR #3689, opened 2025-07-30
[#3679]: Blocks - Can't duplicate any variant of a Block 

### LINKED ISSUE #3679: Blocks - Can't duplicate any variant of a Block
## Problem

Instance: 8-928ba7ffc1-445x05
Block: 7-f8dcf7a6df-4hq7sc

When trying to duplicate any variant of that block, we get an error "Failed to duplicate block variant" error. Upon checking the network tab, there's a failed request to /items showing with this error:

```
{
    "_meta": {
        "timestamp": "2025-07-25T15:03:21.265421066Z",
        "totalResults": 1,
        "start": 0,
        "offset": 0,
        "limit": 1
    },
    "data": {},
    "error": "Bad Request: validation error: Unable to create item: Field doesn't exist: title"
} 
```

The block had a deactivated field which had that required rule but even after removing that required rule, the error still persist. 

## Screenshots

Image

geodem127:
RCA:
- [ ] missing filter for deactivated fields

Fix:
- [ ] Implement logic to filter out deactivated fields.

### DIFF
--- a/src/apps/content-editor/src/app/views/ItemEdit/components/ItemEditHeader/DuplicateItemDialog.tsx
+++ b/src/apps/content-editor/src/app/views/ItemEdit/components/ItemEditHeader/DuplicateItemDialog.tsx
@@ -32,7 +32,10 @@ export const DuplicateItemDialog = ({ onClose }: DuplicateItemProps) => {
     modelZUID: string;
     itemZUID: string;
   }>();
-  const { data: modelFields } = useGetContentModelFieldsQuery(modelZUID);
+  const { data: modelFields } = useGetContentModelFieldsQuery({
+    modelZUID,
+    showDeleted: false,
+  });
   const history = useHistory();
   const item = useSelector(
     (state: AppState) => state.content[itemZUID] as ContentItem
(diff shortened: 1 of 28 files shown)


src/apps/content-editor/src/app/views/ItemEdit/components/ItemEditHeader/DuplicateItemDialog.tsx
```diff
@@ -39,16 +42,27 @@ export const DuplicateItemDialog = ({ onClose }: DuplicateItemProps) => {
   const [createContentItem, { isLoading }] = useCreateContentItemMutation();
 
   const duplicateItem = () => {
+    const activeFields =
+      modelFields?.filter((field) => !field?.deletedAt) || [];
```
agalin920: Can't we fetch fields without the show deleted parameter?

agalin920 (changes requested):

src/shell/services/instance.ts
```diff
@@ -336,14 +336,23 @@ export const instanceApi = createApi({
       invalidatesTags: ["ContentModels", "ContentNav"],
     }),
     // https://www.zesty.io/docs/instances/api-reference/content/models/#Get-Fields
-    getContentModelFields: builder.query<ContentModelField[], string>({
-      query: (modelZUID) =>
-        `content/models/${modelZUID}/fields?showDeleted=true`,
+    getContentModelFields: builder.query<
+      ContentModelField[],
+      string | { modelZUID: string; showDeleted?: boolean }
```
finnar-bin: Is there any specific reason why the param is open to both a string and object being passed in instead of just 1 type?

geodem127: It is just to allow the query with just the modelZUID so we no longer need to update other components that uses that query function.

finnar-bin (approved):

src/shell/services/instance.ts
```diff
@@ -336,14 +336,23 @@ export const instanceApi = createApi({
...
+    >({
+      query: (args) => {
+        const modelZUID = typeof args === "string" ? args : args.modelZUID;
+        const showDeleted =
+          typeof args === "string" ? true : args.showDeleted ?? true;
+
+        return `content/models/${modelZUID}/fields?showDeleted=${showDeleted}`;
```
agalin920: please use the rtk query pattern for applying params

agalin920 (changes requested):

src/shell/services/instance.ts
agalin920: I would say we should update past uses if we do plan on doing this. Supporting two types incurs tech debt. cc: @finnar-bin

agalin920 (changes requested):

finnar-bin (approved):

agalin920 (approved):

PR #3690, opened 2025-07-31
[Global] Support .content.one domain for manager-ui

### LINKED ISSUE #3676: AppShell - New domain support
## Current behavior

Currently the app shell is programmed to work with the *.manager.zesty.io domain

## Proposed change

We need the same logic to support *.cms.content.one and all of the lower environment variations.

## Additional notes

DNS, Infra and API configurations are already in place. The code base needs to be updated to support relieving the session token from the auth.api.zesty.io service, writing to the *.cms.content.one domain(and it's environment variants), and properly sending the session token on subsequent API requests to zesty.io services.

e.g. https://8-f48cf3a682-7fthvk.cms.content.one/

finnar-bin:
Makes sure that all subsequent api calls are using the appropriate auth token when using the `.content.one` domain with manager-ui.

[Content---Zesty-io---Manager.webm](https://github.com/user-attachments/assets/dcf70ea8-c657-485a-a46a-32bfeea676b7)

### DIFF
--- a/src/utility/getRuntimeEnv.ts
+++ b/src/utility/getRuntimeEnv.ts
@@ -31,7 +31,7 @@ export default () => {
         };
 
       default:
-        break;
+        throw new Error(`Invalid NODE_ENV value: ${process.env.NODE_ENV}`);
     }
   } else {
     const COOKIE_DOMAIN = ".zesty.io";
--- a/src/utility/getRuntimeEnv.ts
+++ b/src/utility/getRuntimeEnv.ts
@@ -1,6 +1,5 @@
 export default () => {
-  const isContentOneDomain =
-    window.location.hostname.includes(".cms.content.one");
+  const isContentOneDomain = window.location.hostname.includes(".content.one");
 
   if (isContentOneDomain) {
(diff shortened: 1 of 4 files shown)


src/utility/getRuntimeEnv.ts
```diff
@@ -0,0 +1,68 @@
...
+    switch (process.env.NODE_ENV) {
+      case "production":
+        return {
+          URL_MANAGER: ".cms.content.one",
+          COOKIE_DOMAIN,
+        };
+
+      case "stage":
+        return {
+          URL_MANAGER: ".stage.cms.content.one",
```
shrunyan: The domain structure should be `cms.ENV.content.one`. Wanting that to remain consistent with the zesty.io domain

shrunyan (changes requested):

src/utility/getRuntimeEnv.ts
finnar-bin: @shrunyan updated

shrunyan (approved):

PR #3698, opened 2025-08-05
[#3673] : Blocks - Duplicating a variant and renaming it does not update the seo title

### LINKED ISSUE #3673: Blocks - Duplicating a variant and renaming it does not update the seo title
## Problem

When a user duplicates a variant and renames it, the SEO title is not automatically updated to reflect the new variant name. Instead, it retains the original SEO title from the duplicated variant and just adds "(copy)". This causes the pill at the top (which displays the SEO title) to show outdated information. Since the interface does not provide a way for users to manually edit the SEO title, there is no way to correct this mismatch, leading to confusion.

[- Slack conversation link](https://zesty-io.slack.com/archives/C050120G5/p1753197876880369)
[- Direct link of the instance where this happened](https://8-928ba7ffc1-445x05.manager.zesty.io/launchpad)

## Screenshots

[- Video demo](https://www.loom.com/share/59c0acd513d84abe814d19bf1a7a7542?sid=255b1bf8-9993-4d8d-907d-81b25d14f833)

## Steps to reproduce

1. Go to a block
2. Duplicate a variant
3. Rename the variant
4. The Title in the app tab is not updated

## Expected behavior

The title in the app tab (shown in the image below) should be updated with the new name of the variant

Image

geodem127:
RCA:
- [ ] Undefined metaLinkText: The value was being incorrectly assigned during variant duplication.
- [ ] Stale Data: No cache invalidation after duplicating/renaming variants caused UI-state mismatches.

Fix:
- [ ] Excluded the field during duplication to avoid undefined assignments.
- [ ] Added automatic data refetching and cache invalidation post-duplication/rename to ensure real-time UI updates.

https://github.com/user-attachments/assets/35a3d350-e2e8-4ea6-aec9-f8bdd6887fcd

### DIFF
--- a/src/apps/content-editor/src/app/views/ItemEdit/components/ItemEditHeader/DuplicateItemDialog.tsx
+++ b/src/apps/content-editor/src/app/views/ItemEdit/components/ItemEditHeader/DuplicateItemDialog.tsx
@@ -13,11 +13,13 @@ import { useSelector, useDispatch } from "react-redux";
 import { AppState } from "../../../../../../../../shell/store/types";
 import { ContentItem } from "../../../../../../../../shell/services/types";
 import {
+  instanceApi,
   useCreateContentItemMutation,
   useGetContentModelFieldsQuery,
   useGetContentModelsQuery,
 } from "../../../../../../../../shell/services/instance";
 import { notify } from "../../../../../../../../shell/store/notifications";
+import { fetchItems } from "../../../../../../../../shell/store/content";
 
 type DuplicateItemProps = {
   onClose: () => void;
@@ -34,7 +36,8 @@ export const DuplicateItemDialog = ({ onClose }: DuplicateItemProps) => {
   const item = useSelector(
     (state: AppState) => state.content[itemZUID] as ContentItem
   );
-  const { data: models } = useGetContentModelsQuery();
+  const { data: models, refetch: refetchContentModels } =
+    useGetContentModelsQuery();
 
   const [createContentItem, { isLoading }] = useCreateContentItemMutation();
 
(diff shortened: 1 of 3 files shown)


src/apps/content-editor/src/app/views/ItemEdit/components/ItemEditHeader/DuplicateItemDialog.tsx
```diff
@@ -66,7 +66,6 @@ export const DuplicateItemDialog = ({ onClose }: DuplicateItemProps) => {
         web: {
           canonicalTagMode: item.web.canonicalTagMode,
           parentZUID: item.web.parentZUID,
-          metaLinkText: item.web.metaLinkText?.slice(0, 143) + " (copy)",
```
finnar-bin, line 69: Shouldn't this only be removed when duplicating Block items? As far as I know this component is also being used for regular content items which I assume we would want `metaLinkText` to be included during item duplication

agalin920: Agree we should simply have logic where we don't append copy if its empty

src/shell/store/content.js
```diff
@@ -283,10 +283,16 @@ export function fetchItem(modelZUID, itemZUID) {
...
         return res;
       },
-    }).catch((err) => {
-      console.error("Failed to fetch item:", err);
-      return err;
-    });
+    })
+      .then(() => {
+        dispatch(
+          instanceApi.util.invalidateTags(["ContentItems", "ContentModels"])
```
agalin920: Why are we invalidating on a FETCH_RESOURCE call aka GET? invalidations are for mutations so confused why we need this here

agalin920 (changes requested):

src/shell/store/content.js
geodem127: Sorry, my bad — it was part of my troubleshooting, and I forgot to apply my final stash before pushing.

agalin920 (approved):

PR #3714, opened 2025-08-14
Amplitude integration and workflows tagging

agalin920:


src/shell/views/Shell/Shell.tsx
```diff
@@ -118,7 +131,6 @@ export default memo(function Shell() {
           <Sentry.ErrorBoundary
             fallback={() => <AppError />}
             beforeCapture={(scope) => {
-              scope.setLevel(Severity.Fatal);
```
finnar-bin: This is used to tag a sentry error that actually crashed the app. Any reason this was removed?

geodem127 (approved):

finnar-bin (approved):

src/apps/content-editor/src/app/views/ItemEdit/components/ItemEditHeader/VersionSelector/VersionItem.tsx
```diff
@@ -219,6 +219,7 @@ export const VersionItem = memo(
 
               return (
                 <Chip
+                  id="content-active-status-abel-item"
```
shrunyan: I'm a little concerned about using `id` attributes. As these need to be document unique. If using these in components the likelihood of rendering a component multiple times on the same screen seems high. Meaning we will be triggering browser warnings on duplicate `id`. 

- Would it make sense to use `data-` attributes? Similar to our cypress `data-cy`. 
- Should we just use the existing `data-cy` attributes, instead of adding new ones?

agalin920: @shrunyan immediate reason for this was due to  Amplitude having some accelerated UX when things are tagged with id attribute but you bring up good points. Let me noodle on pros/cons

agalin920: @shrunyan Opting to go with data- attributes but amplitude only captures data attributes with data-amp-track- prefix. 

However by keeping it separate from data-cy it allows analytics and testing to evolve separately

src/apps/settings/src/app/views/User/Workflows/authorized/forms-dialogs/DeactivationDialog.tsx
```diff
@@ -55,6 +55,7 @@ const DeactivationDialog: FC<DeactivationDialogProps> = ({
 
   return (
     <Dialog
+      id="workflows-status-label-deactivation-dialog"
```
agalin920: Whats the purpose of tracking the dialog?

PR #3725, opened 2025-08-19
[Media] Use bearer tokens when uploading/replacing media files

### LINKED ISSUE #3715: Media -  Unable to upload media files on cms.content.one domain
## Problem

Uploading media files on the cms.content.one domain fails due to auth issues

## Screenshots

Image

## Steps to reproduce

1. Go to an instance on the cms.content.one domain
2. Go to media
3. Upload any file

## Expected behavior

Media upload should go through

finnar-bin:
Change the auth method from using cookies to bearer tokens when uploading or replacing media files.
Resolves #3715 

[All-Media---Nar-Folder---Zesty-io---zesty-pw---Manager.webm](https://github.com/user-attachments/assets/584b2ab3-cc80-4322-9db5-edebc61f1298)

### DIFF
--- a/src/shell/store/media-revamp.ts
+++ b/src/shell/store/media-revamp.ts
@@ -5,6 +5,7 @@ import { notify } from "../../shell/store/notifications";
 import { v4 as uuidv4 } from "uuid";
 import { request } from "../../utility/request";
 import { mediaManagerApi } from "../services/mediaManager";
+import Cookies from "js-cookie";
 
 export type StoreFile = {
   uploadID: string;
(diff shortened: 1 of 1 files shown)


geodem127 (approved):

shrunyan:
@finnar-bin how are we formulating the API requests elsewhere? Why isn't this using the same logic? We shouldn't need a one-off cookie access as described in this change.

finnar-bin:
@shrunyan every other request goes through a function named `request` on request.js which is a wrapper for js fetch, the file upload uses XMLHttpRequest to track a file's upload progress that's why the bearer token needs to be manually added here.

agalin920 (approved):

PR #3728, opened 2025-08-20
[Global Sidebar] Refetch headtags on login to get favicon data

### LINKED ISSUE #3720: Global Sidebar - Instance icon doesn't automatically get fetched after login
## Problem

After logging in, the instance favicon remains in the placeholder state unless the user either clicks on it or refreshes the page.

## Screenshots

https://github.com/user-attachments/assets/a7856050-a1c2-426a-a442-c2e9e53d70e1

## Steps to reproduce

1. Visit an instance with a favicon already setup then make sure you're logged out
2. Log in
3. See error

## Expected behavior

After successfully logging in, the instance favicon should immediately be fetched/shown.

finnar-bin:
Manually refetch headtags on successful login so that favicon data can be loaded
Resolves #3720 

[All-Media---media-folder-2-10100b97-fpzv05---Zesty-io---Manager.webm](https://github.com/user-attachments/assets/95b348dd-b3e4-4638-805e-b41de2a393eb)

### DIFF
--- a/src/shell/components/global-sidebar/components/InstanceAvatar.tsx
+++ b/src/shell/components/global-sidebar/components/InstanceAvatar.tsx
@@ -1,4 +1,4 @@
-import { FC, useEffect, useMemo } from "react";
+import { FC, useMemo } from "react";
 import { useSelector, useDispatch } from "react-redux";
 import { Avatar, Skeleton, Box, SxProps, Theme } from "@mui/material";
 import ImageRoundedIcon from "@mui/icons-material/ImageRounded";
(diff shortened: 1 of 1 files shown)


geodem127 (approved):

agalin920 (approved):

src/shell/components/global-sidebar/components/InstanceAvatar.tsx
```diff
@@ -18,10 +18,13 @@ export const InstanceAvatar: FC<InstanceAvatar> = ({
...
-  const ui = useSelector((state: AppState) => state.ui);
+  const { valid } = useSelector((state: AppState) => state.auth);
   const dispatch = useDispatch();
-  const { data: headTags, isLoading: isLoadingHeadTags } =
-    useGetHeadTagsQuery();
+  const {
+    data: headTags,
+    isLoading: isLoadingHeadTags,
+    refetch: refetchHeadTags,
+  } = useGetHeadTagsQuery();
```
agalin920: Will it makes sense to use skip instead so it fetches until valid is true?

agalin920 (changes requested):

src/shell/components/global-sidebar/components/InstanceAvatar.tsx
finnar-bin: Yeah, that actually makes more sense. I've updated the code to do so

agalin920 (approved):

PR #3741, opened 2025-08-26
[Maintenance] Ugrade Cypress to v15

### LINKED ISSUE #3237: [Maintenance] Update Cypress to the latest version
Update Cypress to the latest version (v14)
Latest version supports TS, might need to convert existing tests to TS

finnar-bin:
Upgrade the cypress package from v12 to v15 (latest)
Resolves #3237


agalin920 (approved):

finnar-bin:
Placing on blocked as the node version in the github action needs to be v22 which is what cypress v15 needs. Otherwise, cypress just fails to run.

finnar-bin:
~~**DO NOT MERGE:** A slew of dependency conflicts need to be resolved due to npm v8 that ships with node v22 being stricter needs to be resolved. The following packages need to be updated or replaced with a similar package that supports react 18, these 2 packages are also dependencies on `@zesty-io/core`:~~

~~- `@aeaton/react-prosemirror`~~
~~- `@aeaton/react-prosemirror-config-default`~~

Already updated the ci.yaml to support running cypress with node 22

PR #3746, opened 2025-08-27
[Blocks] - Preview would sometimes show the home page

### LINKED ISSUE #3674: Blocks - Preview would sometimes show the home page
When selecting a block and checking the preview window it would show not show the block preview instead it would show a page 
I have to reload in order to preview the correct one.

Image

It should be showing this.
Image

geodem127:
## Root Cause Analysis:
- [x] Missing update dependencies - The route function was only called on iframe load/readyState, but not when props changed, causing stale content displays.

## Fix Implementation:
- [x] Added useEffect hook that triggers route updates when relevant props change
- [x] Consolidated iframe load logic with a single event listener and immediate readyState check
- [x] Memoized the route function with useCallback to prevent unnecessary recreations

## Additional Enhancements:
- [x] Implemented iframe loading state indicator for better user feedback
- [x] Centralized application configuration retrieval and passed it to child iframe for synchronization
- [x] Optimized data retrieval by using existing store data instead of redundant API calls

https://github.com/user-attachments/assets/c18bfd0c-8b8a-457f-847e-479dc7e208d9


src/apps/active-preview/components/Frame/Frame.js
```diff
@@ -12,21 +12,41 @@ import pixel7cam from "../../../../../public/images/pixel-7-camera.png";
 import styles from "./Frame.less";
 export function Frame(props) {
   const [frameLoading, setFrameLoading] = useState(true);
+  const [routePath, setRoutePath] = useState("");
 
   useEffect(() => {
     setFrameLoading(true);
-  }, [props.device]);
+  }, [props.route]);
```
finnar-bin: Have you verified if this doesn't affect the functionality when changing device previews?

agalin920: plus 1

finnar-bin (approved):

agalin920 (approved):

src/apps/active-preview/Preview.js
```diff
@@ -222,26 +206,21 @@ export function Preview(props) {
...
-      >
-        <CircularProgress />
-        <Typography variant="h5" fontWeight={600} mt={1.5}>
-          Saving
-        </Typography>
-      </Box>
+      <div className={styles.NoDomain}>
+        <h1 className={styles.headline}>
+          {!domain
+            ? "Your session is not active. Please login to Zesty.io"
```
agalin920, line 243: This is an odd addition. Can you describe in what cases this will be rendered? Seems unrelated to the ticket that this aims to close and might need Product input

agalin920 (changes requested):

src/apps/active-preview/Preview.js
geodem127: updated

agalin920: Has the update been pushed?

geodem127: Yes.

agalin920 (approved):

PR #3759, opened 2025-09-01
[Blocks] - Creating a block with a duplicate name fails silently

### LINKED ISSUE #3749: Blocks - Creating a block with a duplicate name fails silently
## Problem

When a user attempts to create a new block with a name that already exists in the system, the operation fails silently. There is no visual feedback, warning message, or error prompt to inform the user of the conflict.

## Screenshots

https://github.com/user-attachments/assets/a03422ec-13b6-4424-8160-d87396971035

## Steps to reproduce

1. Go to "https://8-f48cf3a682-7fthvk.manager.zesty.io/blocks"
2. Click on "Create Block"
3. Select "Side by Side Hero Image"
4. Click on "Create Block Model"
5. See that the page reloads and goes back to Block Type Selector popup.

## Expected behavior

- A clear, non-disruptive error message is displayed within the modal
- The modal remains open, allowing the user to correct the input without losing their work.

geodem127:
### Root Cause:
The form was losing its state because the conditional rendering logic used the isFetching flag. This flag triggers on every data fetch (including background refetches), causing the entire form to re-render and reset.

### Solution:
Changed the condition from isFetching to isLoading to control the loading state. Since isLoading is only true during the initial data load, this prevents unnecessary re-renders during background updates and preserves the form state.

https://github.com/user-attachments/assets/426f61c0-09aa-492a-9e74-65a8b600fba5


src/shell/services/instance.ts
```diff
@@ -847,7 +847,11 @@ export const instanceApi = createApi({
           return error;
         }
       },
-      invalidatesTags: ["ContentModels", "WebViews", "ContentModelFields"],
+      invalidatesTags: (result, error, arg) => {
```
finnar-bin: Doesn't RTK query already by default not invalidate the cache when the api call responded with a non-200 status? Also, wouldn't it be better to handle the re-render on a component-level instead of here?

geodem127: After further investigation, I've corrected my initial assessment as well as the fix.

finnar-bin (approved):

PR #3775, opened 2025-09-08
[Maintenance]: Stabilize Flaky Cypress Tests

geodem127:
### There is an opportunity to enhance the stability of the Cypress tests by looking into the following:

1. Shared Live Data: Tests use a shared database that changes, causing inconsistent results.
2. Unreliable Selectors: Tests use non-resilient selectors that break with UI changes.
3. State Dependency: Tests are not self-contained and rely on the state from previous tests.

## Proposed change

1. Implement Test Data Management: Create and tear down unique, dynamic data for each test.
2. Use Data-Cypress Attributes: Adopt a `data-cy` strategy for robust, purpose-built element targeting.
3. Ensure Test Independence: Restructure specs so each test can run in isolation and support retries.


shrunyan:
Close in favor of https://github.com/zesty-io/manager-ui/pull/3802

PR #3800, opened 2025-09-25
[Maintenance]: Stabilize flaky cypress tests - DB manager Implementation

geodem127:
- [ ] create a plugin that seeds database
- [ ] create json templates to be used as a template when seeding databases
- [ ] relocated network requests when seeding db to the node instance to prevent UI blocking


shrunyan:
Close in favor of https://github.com/zesty-io/manager-ui/pull/3802

PR #3802, opened 2025-09-26
Cypress: Setup Content Fixture Seeding

### LINKED ISSUE #3776: Cypress: Setup Content Fixture Seeding
## Problem

- Tests fail when executed in parallel with other workflows due to reliance on shared live data.
- Some test cases depend on the state established by previous tests, leading to cascading failures if earlier conditions are not met.
- Tests fail if associated test data is removed from the database.
- Tests are sensitive to specific status labels being set, which can cause failures under certain conditions.

## Root Cause Analysis (RCA)
- Shared Live Data: Tests rely on a shared and mutable database, leading to inconsistent and unreliable test outcomes.
- Fragile Selectors: Tests utilize non-robust UI selectors, making them prone to failure when the interface changes.
- State Dependency: Test cases are interdependent and not isolated, relying on the state set by preceding tests, which introduces fragility and flakiness.

## Proposed Changes
- Implement Test Data Management: Generate and clean up unique, dynamic test data for each test to eliminate dependencies on shared live data.
- Adopt Resilient Selectors: Use Cypress-recommended data-cy attributes to enable stable and maintainable element targeting.
- Ensure Test Isolation: Refactor test specifications to ensure each test is self-contained, can run independently, and supports reliable retries.

## Action Plan

1. **Implement Test Data Management:**
This will be implemented incrementally through multiple pull requests, categorized by application areas: content, schema, blocks, media, code, leads, redirects, reports, and settings.

2. **Adopt Cypress-Recommended Selectors:**
Each pull request will include updates to element selectors using data-cy attributes to improve test stability and maintainability.

3. **Ensure Test Independence:**
Test isolation will be addressed progressively, aligned with the same application-area breakdown, ensuring that each test can run independently without relying on shared state.

##  PR
seeder/content: https://github.com/zesty-io/manager-ui/pull/3802
schema:
blocks:
media:
code-editor:
leads:
redirects:
reports:
settings:

shrunyan:
This PR introduces per-suite data seeding to significantly enhance the stability of our test runs.

Currently, concurrent CI pipeline runs can lead to data clashing when tests modify or rely on the same shared database records. This change seeds a unique, isolated dataset for each suite execution, effectively eliminating the risk of data conflicts and reducing intermittent test failures.

**Note**: We rely on the nightly sync to overwrite the downstream environment changes. Effectively acting as our **cleanup process**. Otherwise the data generation while seeding would build up overtime eventually becoming problematic.

Key changes:
- Data isolation: introduce a new pattern for seeding data before a test spec is ran.
- plugin: introduce the usage of cypress plugins
- - by using plugins we can allow for work to happen within the nodejs process when cypress initial starts, before code execution in the headless browser
- add: auth function during seeding 
- add: content fixture data
- add: node-sdk
- update: content.spec.js element selection and view navigation to work with dynamic data generation from seeding.


finnar-bin (approved):

agalin920:
@geodem127 what will occur if an api call fails on the seed step?

cypress.config.js
```diff
@@ -1,4 +1,7 @@
 const { defineConfig } = require("cypress");
+const { execSync } = require("child_process");
+
+const instanceZUID = "8-f48cf3a682-7fthvk";
```
shrunyan: Should this be in the config?

cypress/plugins/seeds/content.ts
```diff
@@ -0,0 +1,124 @@
...
+  items: [],
+};
+
+module.exports = function content(config) {
+  const { formatPathPart } = require("../../../src/utility/formatPathPart");
+
+  async function setUp(path: string, context: Record<string, any> = {}) {
+    const [filePath, target = null] = path.split("#");
+    const fullFilePath = join(__dirname, "../../fixtures/", filePath);
+    if (!existsSync(fullFilePath)) throw new Error("Invalid file path");
```
shrunyan: Always use block statement syntax

cypress/plugins/seeds/content.ts
```diff
@@ -0,0 +1,124 @@
...
+    const jsonString = readFileSync(fullFilePath, "utf8");
+    const json = JSON.parse(jsonString);
+    const jsonData = !target ? json : json?.[target];
+
+    if (!jsonData)
```
shrunyan: Always use block statement syntax

cypress/plugins/seeds/content.ts
```diff
@@ -0,0 +1,124 @@
...
+      throw new Error(`Target '#${target}' not found in ${fullFilePath}`);
+    if (!jsonData?.model)
```
shrunyan: Always use block statement syntax

cypress/plugins/seeds/content.ts
```diff
@@ -0,0 +1,124 @@
...
+      throw new Error("Seed data must contain a model definition");
+
+    const jsonModel = jsonData?.model;
```
shrunyan: Why assign the `jsonData.model` reference to a variable? Now I have to switch what I'm remembering when I think about the following code. 
Why use a optional chaining? You already check for the presence of the `model` key in the statement before.

cypress/plugins/seeds/content.ts
```diff
@@ -0,0 +1,124 @@
...
+                  ...fieldPayload,
+                };
+              });
+          })
+        )
+      : [];
+    LOOKUP_CONTEXT.fields = fields;
+
+    // 3) Create Items
+    const items = !!jsonItems?.length
```
shrunyan: What is the purpose of the `!!` operator?

shrunyan: Why use a ternary statement? Lookup the `map` function and understand its return value

cypress/plugins/seeds/content.ts
```diff
@@ -0,0 +1,124 @@
...
+  fields: [],
+  items: [],
+};
+
+module.exports = function content(config) {
+  const { formatPathPart } = require("../../../src/utility/formatPathPart");
+
+  async function setUp(path: string, context: Record<string, any> = {}) {
+    const [filePath, target = null] = path.split("#");
+    const fullFilePath = join(__dirname, "../../fixtures/", filePath);
```
shrunyan: Why is `path` a variable? This fixture is 1:1 related to this setup.

cypress/fixtures/content.json
```diff
@@ -0,0 +1,745 @@
...
+          "text": "text",
+          "wysiwyg_basic": "wysiwyg_basic",
+          "yes_no": 0,
+          "link": "https://www.zesty.io/",
+          "images": "{{baseUrl}}images/defaultImg.png,{{baseUrl}}images/onboardingIcon.svg,{{baseUrl}}images/zestyLogo.svg,{{baseUrl}}images/blocksOnboarding3.png"
+        }
+      }
+    ]
+  },
+  "block": {
```
shrunyan: why is block in the content entity fixture?

cypress/fixtures/content.json
```diff
@@ -0,0 +1,745 @@
...
+          "link": "https://accounts-api.zesty.org/",
+          "markdown": "New Model - Markdown",
+          "og_image": "https://storage.googleapis.com/zesty-dev-instance-screenshots/375tcfqp-dev.preview.dev.zesty.io!-!block!block_model.html!variant=7-baebe39995-g8lq7h&version=3&_bypassError=true&preview=global_1280_720.png",
+          "text": "New Model - Text",
+          "textarea": "New Model - Textarea"
+        }
+      }
+    ]
+  },
+  "common": {
```
shrunyan: why is common in the content entity fixture?

cypress/plugins/seeds/utils.ts
```diff
@@ -0,0 +1,91 @@
...
+      mediaAPIURL: config.env.MEDIA_MANAGER_URL,
+    });
+
+    sdkCache.set(cacheKey, sdkInstance);
+  }
+
+  return sdkCache.get(cacheKey);
+}
+
+export function lookupValue(
```
shrunyan: What is this? Why aren't we feeding direct payloads into the SDK call when seeding?

geodem127:
> @geodem127 what will occur if an api call fails on the seed step?

test will not be interrupted but it'll log an error in cypress then proceeds.

cypress/plugins/seeds/utils.ts
```diff
@@ -0,0 +1,91 @@
...
+  context: Record<string, any> = {}
+): any {
+  if (!jsonData) {
+    return jsonData;
+  }
+  const jsonString =
+    typeof jsonData !== "string" ? JSON.stringify(jsonData) : jsonData;
+  const safeContext = context || {};
+
+  const replaced = jsonString.replace(/{{(.*?)}}/g, (_, expr) => {
```
agalin920: Related to @shrunyan question. Can you explain the purpose of this function as a whole. This regex is concerning

cypress/plugins/seeds/content.ts
geodem127: This lets us seed data on demand by running the task at any point in the test and pointing it to the seed file.

cypress/fixtures/content.json
geodem127: I considered placing all the required seed data for a given spec into a single file (e.g., content.spec.test) and targeting it via the URL fragment passed to the path variable, like /path.json#model|block.

cypress/plugins/seeds/utils.ts
geodem127: Most of the time, we need to inject dynamic data from another source — for example, the relatedModelZUID of a created model, or data from the response of a previously run task.

cypress/plugins/seeds/utils.ts
geodem127: This function looks up the context passed to the task. Any value wrapped in {{}} is evaluated against that context and replaced with the corresponding value

cypress/plugins/seeds/content.ts
geodem127: !! is indeed not needed here since jsonItems?.length already works fine in the ternary. I used the ternary operation because the return value of the promise has already been transformed, so this way I no longer need an if block statement. Would you consider this a bad approach, or is it acceptable in this case?

geodem127: I just realized this would be very hard to troubleshoot when seeding fails, so I've added error logging to both fields and items validation processes

cypress/fixtures/content.json
geodem127: updated seeding data into individual files for better maintainability. 
Added seed data for block and additional model to meet the content spec requirements.

cypress/plugins/seeds/content.ts
shrunyan: That would still be possible with a static file path. As well as does not speak towards the 1:1 nature of the seed to test suite.

cypress/fixtures/content.json
shrunyan: Do we have block tests as part of the content tests? If so I would imagine they should be separate as they have different apps.

cypress/plugins/seeds/utils.ts
shrunyan: > data from the response of a previously run task.

That does not make sense. Seeding should be a deterministic and only occur once. I do not see why we would have dynamic data.

cypress/plugins/seeds/content.ts
shrunyan: I'm just not sure it is necessary. The `jsonItems` comes from a fixture we provide so there should be a guarantee that the expected data shape and values are present. Meaning checking data shape of Array and length indicating presence of values doesn't seem necessary. Along with `map` always returns an array so it will ensure Promise.all receives it's expected type

cypress/plugins/seeds/utils.ts
shrunyan: Wouldn't the related data need to already be created and statically defined in the fixture? Are you saying we are creating the related data as part of the seeding process and then setting dynamically into the fixture? I am not seeing creating related data in this PR code.

cypress/fixtures/content.json
geodem127: There is 1 block selector field

cypress/plugins/seeds/utils.ts
geodem127: I’m not entirely sure how we’ll populate the dropdowns and checkboxes for the relational and block selector fields. What I did was first create the “common” setup for the one_to_one and one_to_many fields, and then built the data (“block fixture”) for the block selector field.

cypress/plugins/seeds/utils.ts
agalin920: Im still not following can you elaborate further please and why this is needed?

cypress.config.js
geodem127: refactored: instanceZUID is acquired using the baseUrl

cypress/plugins/seeds/content.ts
geodem127: updated to block statement syntax. removed unnecessary checks

cypress/plugins/seeds/content.ts
geodem127: updated to block statement syntax. removed unnecessary checks

cypress/plugins/seeds/content.ts
geodem127: updated to block statement syntax. removed unnecessary checks

cypress/plugins/seeds/content.ts
geodem127: update: remove unnecessary variable assignments

cypress/plugins/seeds/content.ts
geodem127: updates: removed unnecessary checks

cypress/plugins/seeds/content.ts
geodem127: updates: 
- Create specialized module and fixture exclusively for content.spec tests

cypress/fixtures/content.json
geodem127: updates:
- Replace dynamic data seeding with pre-configured test data
- Remove dynamic seeding in favor of static test fixtures

cypress/fixtures/content.json
geodem127: updates:
- Replace dynamic data seeding with pre-configured test data
- Remove dynamic seeding in favor of static test fixtures

cypress/plugins/seeds/utils.ts
geodem127: updates:
- Replace dynamic data seeding with pre-configured test data
- Remove dynamic seeding in favor of static test fixtures

cypress/plugins/seeds/utils.ts
geodem127: updates:
- Replace dynamic data seeding with pre-configured test data
- Remove dynamic seeding in favor of static test fixtures

cypress/e2e/content/content.spec.js
```diff
@@ -206,43 +229,54 @@ describe("Content Specs", () => {
...
+      cy.get("[data-cy='field:sort'] input[type='text']").should(
+        "have.value",
+        "12"
+      );
     });
 
     // Skipping relationship tests due to current fetching flow limitation
     it.skip("One to many Field", () => {
-      // cy.get("#12-269a28-1bkm34 input").clear();
+      // cy.get("[data-cy='field:one_to_many'] input").clear();
```
finnar-bin: Remove if unused

cypress/e2e/content/content.spec.js
```diff
@@ -314,26 +361,28 @@ describe("Content Specs", () => {
...
-        cy.visit("/content/6-556370-8sh47g/7-b939a4-457q19");
+        cy.visit(
+          `/content/${Cypress.env("modelZUID")}/${Cypress.env("itemZUID")}`
+        );
       });
     });
 
     it("should be able to clear date entries", () => {
-      cy.get("#12-63ab04-0nkwcc", { timeout: 10000 })
+      cy.get("[data-cy='field:date']", { timeout: 30000 })
```
finnar-bin: Wasn't it agreed on to remove long timeouts like this?

cypress/e2e/content/content.spec.js
```diff
@@ -343,68 +392,70 @@ describe("Content Specs", () => {
...
-        cy.visit("/content/6-556370-8sh47g/7-b939a4-457q19");
+        cy.visit(
+          `/content/${Cypress.env("modelZUID")}/${Cypress.env("itemZUID")}`
+        );
       });
     });
 
     it("should be able to clear date and time entries", () => {
-      cy.get("#12-f3db44-c8kt0q", { timeout: 10000 })
+      cy.get("[data-cy='field:datetime']", { timeout: 30000 })
```
finnar-bin: Same comment here regarding the timeout

cypress/e2e/content/content.spec.js
```diff
@@ -510,50 +569,55 @@ describe("Content Specs", () => {
...
-        force: true,
-      });
+        "[data-cy='field:one_to_many'] [data-cy='create-new-relational-item-button']"
+      )
+        .should("be.enabled")
+        .click();
 
-      cy.get("#12-d6e4c1d797-sjv628", { retries: 1 })
+      cy.get(
+        "#createNewItemDialog [data-cy='field:node-sdk_updateItem_1733876716599']"
```
finnar-bin, line 598: Is there any reason why this field is named this way?

cypress/e2e/content/content.spec.js
```diff
@@ -9,23 +9,36 @@ const formatDate = (ts) =>
 describe("Content Specs", () => {
   const TIMESTAMP = Date.now();
 
+  before(() => {
+    // Seed content
+    cy.task("seed:content").then(({ model, items }) => {
```
finnar-bin, line 14: What happens if the seed:content task fails, like for example an api call  to create the schema or fields fail?

package-lock.json
```diff
@@ -9966,6 +9994,16 @@
...
+    "node_modules/moment": {
+      "version": "2.30.1",
+      "resolved": "https://registry.npmjs.org/moment/-/moment-2.30.1.tgz",
+      "integrity": "sha512-uEmtNhbDOrWPFS+hdjFCBfy9f2YoyzRpwcl+DqpC6taX21FzsTLQVbMV/W7PzNSX6x/bhC1zA3c2UQ5NzH6how==",
+      "dev": true,
+      "license": "MIT",
+      "engines": {
+        "node": "*"
+      }
+    },
```
finnar-bin: Kind of concering that moment is being brought back in since it's going to inflate the bundle size again, but I do see it's a dependency on @zesty-io/sdk

finnar-bin (changes requested):

cypress/e2e/content/content.spec.js
geodem127: This is a predefined item field (Articles)

cypress/e2e/content/content.spec.js
geodem127: - A CypressError on model creation fails the test immediately.
- A successful request with an error response allows the test to continue, but it will fail on navigation (due to missing modelZUID/itemZUID) or field assertions.

cypress/e2e/content/content.spec.js
geodem127: updated

cypress/e2e/content/content.spec.js
geodem127: update: removed timeout

cypress/e2e/content/content.spec.js
geodem127: update: removed timeout

package-lock.json
geodem127: Agreed on the bundle size concern. @shrunyan @agalin920

geodem127: From my analysis of the compiled output, it appears that Moment.js isn't making it into the final build, and the SDK dependency seems to be properly excluded from the frontend bundle. However, I'd appreciate a second set of eyes to verify this finding to be certain.

shrunyan: As the SDK is only being used in the CI process, included as a devDependency I would assume, it should not be built into the client bundle. So non-issue with regards to bundle size.

cypress/plugins/seeds/content.ts
```diff
@@ -0,0 +1,100 @@
...
+      ...json.model,
+      label: modelLabel,
+      metaTitle: modelLabel,
+      name: formatName(modelLabel),
+    };
+    const modelResponse = await sdk.instance.createModel(modelPayload);
+    const model = modelResponse?.data;
+
+    // 2) Create Fields
+    let fields = await Promise.all(
```
finnar-bin: Should be declared as a const as I don't see this value being changed anywhere

cypress/plugins/seeds/content.ts
```diff
@@ -0,0 +1,100 @@
...
+      json.fields.map((field) => {
+        return sdk.instance.createField(model?.ZUID, field).then((res) => ({
+          ...field,
+          ZUID: res?.data?.ZUID || null,
+        }));
+      })
+    );
+
+    // 3) Create Items
+    let items = await Promise.all(
```
finnar-bin: Same here, should be declared as a const as well since value doesn't change

finnar-bin (changes requested):

cypress/e2e/content/content.spec.js
finnar-bin: Can you point me towards where this cypress error is being emitted? I don't seem to see where this happens on the seed:content cypress task.

geodem127: Any failure inside the task (database, file system, request) causes Cypress to fail the test.
I didn't add custom error handling because it's unnecessary—Cypress captures all errors automatically, regardless of their source. Letting them bubble up is the correct behavior.

cypress/plugins/seeds/content.ts
geodem127: Updated

cypress/plugins/seeds/content.ts
geodem127: Updated

finnar-bin (approved):
Looks good to me, let's wait for @agalin920 to review it as well.

agalin920 (approved):
lgtm

PR #3824, opened 2025-10-07
[Global] Add list of whitelisted domains to the CSP connect-src directive

### LINKED ISSUE #3819: Global - Prevent browser extensions from collecting data
## Problem

If a user has a compromised browser extension that's stealing and sending out data to an external domain, we currently do not have a way of knowing if this is happening behind the scenes.

## Solution

Create a service worker that will listen to all network requests and block network requests to non-zesty domains

## Design

To follow

## Steps

- [ ] Find prior art on how to block and prevent browser extensions from stealing data
- [ ] Check if a serviceworker can monitor a child frame network requests
- [ ] Create a POC of a serviceworker that monitors all network requests and block non-zesty domains. Upon blocking, send a response with an appropriate status and with an error body regarding why we blocked this network request

finnar-bin:
Add a list of whitelisted domains to the app's connect-src directive
Resolves #3819

### DIFF
--- a/etc/nginx.conf
+++ b/etc/nginx.conf
@@ -33,7 +33,7 @@ http {
       add_header X-Permitted-Cross-Domain-Policies "";
       add_header X-Frame-Options "deny";
       add_header Referrer-Policy "no-referrer-when-downgrade";
-      add_header Content-Security-Policy "";
+      add_header Content-Security-Policy "connect-src 'self' *.api.dev.zesty.io;";
       add_header Feature-Policy "";
     }
 
--- a/etc/nginx.conf
+++ b/etc/nginx.conf
@@ -33,7 +33,7 @@ http {
       add_header X-Permitted-Cross-Domain-Policies "";
       add_header X-Frame-Options "deny";
       add_header Referrer-Policy "no-referrer-when-downgrade";
-      add_header Content-Security-Policy "connect-src 'self' *.api.dev.zesty.io;";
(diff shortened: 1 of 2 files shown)


etc/nginx.conf
```diff
@@ -33,7 +33,7 @@ http {
       add_header X-Permitted-Cross-Domain-Policies "";
       add_header X-Frame-Options "deny";
       add_header Referrer-Policy "no-referrer-when-downgrade";
-      add_header Content-Security-Policy "";
+      add_header Content-Security-Policy "connect-src 'self' *.api.dev.zesty.io *.amplitude.com *.zesty.io *.a.run.app *.tiny.cloud *.getbynder.com *.bynder.com d8ejoa1fys2rk.cloudfront.net *.sentry.io www.googleapis.com us-central1-zesty-dev.cloudfunctions.net;";
```
agalin920: doesnt *.zesty.io already include *.api.dev.zesty.io?

finnar-bin: @agalin920 Updated, I forgot to remove that when I added *.zesty.io

etc/nginx.conf
```diff
@@ -33,7 +33,7 @@ http {
       add_header X-Permitted-Cross-Domain-Policies "";
       add_header X-Frame-Options "deny";
       add_header Referrer-Policy "no-referrer-when-downgrade";
-      add_header Content-Security-Policy "";
+      add_header Content-Security-Policy "connect-src 'self' *.amplitude.com *.zesty.io *.a.run.app *.tiny.cloud *.getbynder.com *.bynder.com d8ejoa1fys2rk.cloudfront.net *.sentry.io www.googleapis.com us-central1-zesty-dev.cloudfunctions.net;";
```
shrunyan, line 36: What is the AWS d8ejoa1fys2rk.cloudfront.net URL? DM if this is customer information.

finnar-bin: Oh I forgot to add some comments to outline the non-obvious domains, but that one's for the bynder modules. I decided to not use a wildcard to make sure we're specifically only allowing that domain.

agalin920 (approved):

PR #3828, opened 2025-10-08
[#3818]: Content - Publish status chip in data table not displaying correctly

### LINKED ISSUE #3818: Content - Publish status chip in data table not displaying correctly
## Problem

There is an issue with the display of the published status in the data table (content grid) where items that are already published are initially shown in the unpublished state (blue chip).

This behavior makes it difficult for users to accurately isolate unpublished items when reviewing content in the grid, leading to confusion and unnecessary validation steps.

- see affected content item table [here](https://8-eae68eeef3-7q6bpc.manager.zesty.io/content/6-c4998c9d93-wtqfwv?lang=en-AU)

## Screenshots/video reference

https://github.com/user-attachments/assets/0839b848-7222-4113-952f-296410e63bc5

## Steps to reproduce

1. Navigate to a content data table/grid, see example [content item table](https://8-eae68eeef3-7q6bpc.manager.zesty.io/content/6-c4998c9d93-wtqfwv?lang=en-AU) 
2. Observe the content items publish status. Several items that are known to be published will display the blue unpublished status indicator.
3. Click on one of these items to open the content editor and confirm it is published.
4. Navigate back to the data grid.

**Note:** I haven't replicated this in my test instance, issue could be related to long-list of items in table or using multiple locales

## Expected behavior

The content item's status indicator in the data table/grid should accurately reflect its true published status (i.e., display the green published status if the item is published) immediately upon initial page load.

geodem127:
RCA:
- Item publishings were not loading upon model initialization or language change.

FIX:
- Implemented a call to fetchItemPublishings during model load and after any language switch.

https://github.com/user-attachments/assets/b0dad085-5483-4257-a868-d3c0eaa03985

### DIFF
--- a/src/apps/content-editor/src/app/views/ItemList/index.tsx
+++ b/src/apps/content-editor/src/app/views/ItemList/index.tsx
@@ -34,7 +34,10 @@ import {
   ContentItemWithDirtyAndPublishing,
   ContentModelFieldDataType,
 } from "../../../../../../shell/services/types";
-import { fetchItems } from "../../../../../../shell/store/content";
+import {
+  fetchItems,
+  fetchItemPublishings,
+} from "../../../../../../shell/store/content";
 import { TableSortContext } from "./TableSortProvider";
 import { fetchFields } from "../../../../../../shell/store/fields";
 import { debounce } from "lodash";
@@ -180,6 +183,7 @@ export const ItemList = () => {
         })
         // @ts-ignore
       ).then(() => {
+        dispatch(fetchItemPublishings());
         setIsModelItemsFetching(false);
       });
     }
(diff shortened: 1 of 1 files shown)


src/apps/content-editor/src/app/views/ItemList/index.tsx
```diff
@@ -180,6 +183,7 @@ export const ItemList = () => {
         })
         // @ts-ignore
       ).then(() => {
+        dispatch(fetchItemPublishings());
```
agalin920: Why are we doing this after a fetchItems call. Why not in parallel? Is this call really dependent on the first?

geodem127: updated

agalin920 (approved):

finnar-bin (approved):

PR #3869, opened 2025-10-21
[#3820]: Content - Item duplication issue affecting localization of relational fields

### LINKED ISSUE #3820: Content - Item duplication issue affecting localization of relational fields
## Problem

When an item is duplicated, eg. EN-NZ version of the new copy incorrectly pulls the field value from the EN-AU version of the original item for its **One-to-One relational** field. It should be pointing to the EN-NZ value from the original item.

This issue also affects items created via API and seems to have started occurring around September 29th (AEST).

## Screenshots/video demo

https://github.com/user-attachments/assets/cb83219c-cf96-428d-a2fd-742ddbd02ae7

## Steps to reproduce

1. Take an existing item with multiple locales eg. EN-AU, EN-NZ, and with a relational field (eg. One-to-One "Form" field) 
2. Duplicate the item.
3. Examine the language/locale of the relational field when switch to the other locale eg. EN-NZ localization of the newly duplicated item.
4. Observe that the One-to-One field (switch to NZ version) is incorrectly set to the original item's EN-AU value.

Here's a [sample copied item](https://8-82d0f8aa87-9p0kv5.manager.zesty.io/content/6-cafbdf99b4-3sgscn/7-a6e386eeb6-zsws6z) for review.

## Expected behavior

When an item (having multiple locales) with relational field is duplicated, the relational field's language/locale of the duplicated item (switch to other locale version) should pick up the correct locale

geodem127:
## Root Cause:
When an item is duplicated, the system incorrectly loads the default language version instead of the version corresponding to the user's currently active language. This causes a mismatch between the UI display and the intended language of the new item.

## Fix:
Updated the duplication logic to correctly identify and use the language-specific ZUID of the item being copied, ensuring the UI displays the proper language version upon duplication.

https://github.com/user-attachments/assets/1534422e-7851-4533-9a15-79232ae9d4d3

### DIFF
--- a/src/apps/content-editor/src/app/views/ItemEdit/components/ItemEditHeader/DuplicateItemDialog.tsx
+++ b/src/apps/content-editor/src/app/views/ItemEdit/components/ItemEditHeader/DuplicateItemDialog.tsx
@@ -98,8 +98,9 @@ export const DuplicateItemDialog = ({ onClose }: DuplicateItemProps) => {
       .unwrap()
       .then((res) => {
         refetchContentModels();
-        Promise.resolve(dispatch(fetchItems(modelZUID))).then(
-          (fetchItemsResponse: any) => {
+        dispatch(fetchItems(modelZUID))
+          //@ts-ignore
+          .then((fetchItemsResponse: any) => {
             const createdItem = fetchItemsResponse?.data?.find(
               (item: ContentItem) => item?.meta?.ZUID === res.data.ZUID
             );
@@ -113,8 +114,7 @@ export const DuplicateItemDialog = ({ onClose }: DuplicateItemProps) => {
                   : "content"
               }/${modelZUID}/${itemLangZUID}`
             );
-          }
-        );
+          });
         onClose();
       })
       .catch((error) => {
--- a/src/apps/content-editor/src/app/views/ItemEdit/components/ItemEditHeader/DuplicateItemDialog.tsx
+++ b/src/apps/content-editor/src/app/views/ItemEdit/components/ItemEditHeader/DuplicateItemDialog.tsx
@@ -41,10 +41,8 @@ export const DuplicateItemDialog = ({ onClose }: DuplicateItemProps) => {
   const item = useSelector(
     (state: AppState) => state.content[itemZUID] as ContentItem
   );
-  const languages = useSelector((state: AppState) => state.languages);
-  const langCode =
-    languages?.find((lang: Language) => lang?.ID === item?.meta?.langID)
-      ?.code || languages?.find((lang: Language) => lang?.default)?.code;
+  const langCode = useSelector((state: AppState) => state?.user?.selected_lang);
+
   const { data: models, refetch: refetchContentModels } =
     useGetContentModelsQuery();
 
--- a/src/apps/content-editor/src/app/views/ItemEdit/components/ItemEditHeader/DuplicateItemDialog.tsx
+++ b/src/apps/content-editor/src/app/views/ItemEdit/components/ItemEditHeader/DuplicateItemDialog.tsx
@@ -13,10 +13,9 @@ import { useSelector, useDispatch } from "react-redux";
 import { AppState } from "../../../../../../../../shell/store/types";
 import {
   ContentItem,
-  Data,
+  Language,
 } from "../../../../../../../../shell/services/types";
 import {
-  instanceApi,
   useCreateContentItemMutation,
   useGetContentModelFieldsQuery,
   useGetContentModelsQuery,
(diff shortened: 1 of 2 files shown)


src/apps/content-editor/src/app/views/ItemEdit/components/ItemEditHeader/DuplicateItemDialog.tsx
```diff
@@ -94,16 +97,25 @@ export const DuplicateItemDialog = ({ onClose }: DuplicateItemProps) => {
...
-        dispatch(fetchItems(modelZUID));
         refetchContentModels();
-        onClose();
-        history.push(
-          `/${
-            models?.find((model) => model?.ZUID === modelZUID)?.type === "block"
-              ? "blocks"
-              : "content"
-          }/${modelZUID}/${res.data.ZUID}`
+        Promise.resolve(dispatch(fetchItems(modelZUID))).then(
```
agalin920: Promise.resolve() isn’t doing anything useful here is it? because if dispatch(fetchItems(modelZUID)) already returns a promise. Wrapping a promise with Promise.resolve() simply returns the same promise which is just redundant. 

If dispatch were synchronous, the .then() would still run immediately on the next microtask, giving the false impression of waiting for async work.

agalin920 (changes requested):

src/apps/content-editor/src/app/views/ItemEdit/components/ItemEditHeader/DuplicateItemDialog.tsx
geodem127: Agreed. The Promise.resolve() is only needed because TypeScript incorrectly infers the return type of the dispatched thunk. 
 I'll replace it with `// @ts-ignore `to directly suppress the error.

src/apps/content-editor/src/app/views/ItemEdit/components/ItemEditHeader/LanguageSelector.tsx
```diff
@@ -68,11 +69,12 @@ export const LanguageSelector = () => {
...
 
-  const activeLanguage = languages?.find(
-    (lang) => lang.ID === item?.meta?.langID
-  ) || {
-    code: "en-US",
-  };
+  useEffect(() => {
+    if (!isLoadingLanguages) {
+      const lang = languages?.find((lang) => lang.ID === item?.meta?.langID);
+      if (lang) setActiveLanguage(lang);
```
finnar-bin: Use block scopes instead of one-liners

finnar-bin (changes requested):

src/apps/content-editor/src/app/views/ItemEdit/components/ItemEditHeader/LanguageSelector.tsx
geodem127: Updated

src/apps/content-editor/src/app/views/ItemEdit/components/ItemEditHeader/LanguageSelector.tsx
```diff
@@ -52,6 +52,7 @@ export const LanguageSelector = () => {
   const item = useSelector(
     (state: AppState) => state.content[itemZUID] as ContentItem
   );
+  const [activeLanguage, setActiveLanguage] = useState({ code: "en-US" });
 
   const onSelect = (langId: string) => {
     dispatch(selectLang(langId));
```
agalin920, line 57: we have this selectLang on select. should we not be using this to set the active lang? is there not a redux store that already supports this that we can select from?

agalin920:
So i see this PR aims to redirect to the corresponding locale when duplicating an item. But the ticket in question states

"When an item is duplicated, eg. EN-NZ version of the new copy incorrectly pulls the field value from the EN-AU version of the original item for its One-to-One relational field"

What that sounds more like is that the item duplication is not pulling the correct values per locale. Is this not the case and in reality the reporter is mistaken?

agalin920 (changes requested):

geodem127:
> So i see this PR aims to redirect to the corresponding locale when duplicating an item. But the ticket in question states
> 
> "When an item is duplicated, eg. EN-NZ version of the new copy incorrectly pulls the field value from the EN-AU version of the original item for its One-to-One relational field"
> 
> What that sounds more like is that the item duplication is not pulling the correct values per locale. Is this not the case and in reality the reporter is mistaken?

The item duplication is actually working correctly—the new EN-NZ item properly gets its values from the original EN-NZ item. The problem is that after the duplication succeeds, the user is incorrectly redirected to view the new item in the default locale (e.g., EN-AU) instead of the locale they duplicated from (EN-NZ).

src/apps/content-editor/src/app/views/ItemEdit/components/ItemEditHeader/LanguageSelector.tsx
geodem127: Yes, you're correct. The current implementation now uses the `selectLang` selector. The more complex approach was a remnant of an initial attempt to address a UI flicker, which fell outside the scope of this PR.

Since that issue wasn't part of the original report, I've removed that logic to keep the fix focused.

agalin920 (approved):

finnar-bin (approved):

PR #3870, opened 2025-10-22
[Content] Ensure hooks are added before react renders

### LINKED ISSUE #3854: Error: Rendered fewer hooks than expected. This may be caused by an accidental early return statement.
Sentry Issue: [MANAGER-UI-2FX](https://zestyio.sentry.io/issues/6942609388/?referrer=github_integration)

```
React ErrorBoundary Error: Rendered fewer hooks than expected. This may be caused by an accidental early return statement.
  at ActiveItem (./src/shell/components/RelationalFieldBase/ActiveItem/index.tsx:58:5)
  at RelationalFieldBase (./src/shell/components/RelationalFieldBase/index.tsx:36:3)
  at FieldShell (./src/apps/content-editor/src/app/components/Editor/Field/FieldShell.tsx:59:3)
  at jA (./src/apps/content-editor/src/app/components/Editor/Field/Field.tsx:96:5)
  at av (./src/apps/content-editor/src/app/components/Editor/Editor.js:25:3)
...
(45 additional frame(s) were not displayed)

Error: Rendered fewer hooks than expected. This may be caused by an accidental early return statement.
  at Nh (./node_modules/react-dom/cjs/react-dom.production.min.js:160:337)
  at cj (./node_modules/react-dom/cjs/react-dom.production.min.js:196:258)
  at bj (./node_modules/react-dom/cjs/react-dom.production.min.js:194:204)
  at Vk (./node_modules/react-dom/cjs/react-dom.production.min.js:296:100)
  at Uk (./node_modules/react-dom/cjs/react-dom.production.min.js:279:389)
...
(6 additional frame(s) were not displayed)
```

finnar-bin:
Resolves an issue where an early render is sometimes causing the component to render with fewer hooks
Fixes #3854

### DIFF
--- a/src/shell/components/RelationalFieldBase/ActiveItem/index.tsx
+++ b/src/shell/components/RelationalFieldBase/ActiveItem/index.tsx
@@ -162,7 +162,10 @@ export const ActiveItem = memo(
 
     // Ensure item is fetched if it wasn't included in the initial 100-item fetch
     useEffect(() => {
-      if (contentItem || !relatedModelData?.ZUID || !itemZUID) return;
+      if (contentItem || !relatedModelData?.ZUID || !itemZUID) {
+        return;
+      }
+
       dispatch(fetchItem(relatedModelData.ZUID, itemZUID));
     }, [contentItem]);
 
--- a/src/shell/components/RelationalFieldBase/ActiveItem/index.tsx
+++ b/src/shell/components/RelationalFieldBase/ActiveItem/index.tsx
@@ -160,6 +160,12 @@ export const ActiveItem = memo(
       };
     }, [contentItems, itemZUID, relatedModelData, users]);
 
+    // Ensure item is fetched if it wasn't included in the initial 100-item fetch
+    useEffect(() => {
+      if (contentItem || !relatedModelData?.ZUID || !itemZUID) return;
+      dispatch(fetchItem(relatedModelData.ZUID, itemZUID));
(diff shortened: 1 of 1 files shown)


geodem127 (approved):

shrunyan:
@finnar-bin I think this should include 1 more change. The fact that there is an "early" return is problematic. I think it should be an `if/else` statement. This way the "return"s are connected in a logic block. This issue being a good example of why most often whenever there is a logic case for an if/else we should use that syntax versus what currently exists in this file. An "if" statement with an implicit else.

agalin920:
> @finnar-bin I think this should include 1 more change. The fact that there is an "early" return is problematic. I think it should be an `if/else` statement. This way the "return"s are connected in a logic block. This issue being a good example of why most often whenever there is a logic case for an if/else we should use that syntax versus what currently exists in this file. An "if" statement with an implicit else.

@shrunyan  early returns are quite idiomatic in react https://react.dev/learn/conditional-rendering.

The main argument for favoring early returns as a best practice for conditional rendering and handling component logic is that it means you handle edge cases and negative conditions at the top of your component function and return early, instead of nesting your entire component logic within a large if...else block. 

The docs do state that hooks must be called at the top level of your component, before any early returns. Since as this error proved calling a Hook conditionally, after an early return, will cause an error because it breaks the consistent order in which React calls hooks on every render.

agalin920 (approved):

PR #3871, opened 2025-10-22
[#3833]: Content - Item with relational field not reflecting correct locale when switch to other language version

### LINKED ISSUE #3833: Content - Item with relational field not reflecting correct locale when switch to other language version
## Problem

When a user creates a new item (with multiple locales configured) and sets a value for a relational field (a field linking to another item), that relational value is not correctly fetching the appropriate localized data when the user switches to a different locale/language version.

## See below test video

https://github.com/user-attachments/assets/8150c9e8-9f99-46b0-aaa0-5866982b8c84

## Steps to reproduce

1. Ensure you have at least two locales configured (eg. en-US, en-PH)
2. Create a new item with a relational field (eg. One-to-One field), and set the relational field value then save.
4. Examine the language/locale of the relational field when switch to the other locale (eg. en-PH)
5. Observe that the One-to-One field (eg. switch to Ph version) is incorrectly set to the original item's locale value (en-US).

## Expected behavior

When an item with relational field is created having multiple locales, and when switch to other language version the relational field's value should pick up the correct localized value

geodem127:
## Root Cause:
The relational field inherits the language selection from when the source item was originally created.

## Fix:
Implemented logic to retrieve content items localized to the currently selected language.

https://github.com/user-attachments/assets/5fbc8118-1215-4c99-88a0-4ca4e2aa1e70


src/apps/content-editor/src/app/components/Editor/Editor.js
```diff
@@ -40,6 +40,12 @@ export default memo(function Editor({
   const [isLoaded, setIsLoaded] = useState(false);
   const [prevFirstContentFieldValue, setPrevFirstContentFieldValue] =
     useState(null);
+  const languages = useSelector((state) => state.languages);
+  const contentItems = useSelector((state) => state.content);
```
agalin920: This is a pretty significant selector that will cause the whole Editor to re-render on content change. We should move it as low as possible in the component tree

src/apps/content-editor/src/app/components/Editor/Editor.js
```diff
@@ -393,6 +399,13 @@ export default memo(function Editor({
   return (
     <div className={styles.Fields}>
       {activeFields?.map((field) => {
+        const fieldItem = contentItems?.[item?.data?.[field.name]];
+        const fieldValue = ["one_to_one", "one_to_many"].includes(
```
agalin920: Related to previous comment. Would it not be a better idea to add this business logic in the relationship field component itself specifically in these types case blocks? since it only relates to specific field types and is much lower in the component tree

agalin920 (changes requested):
.

src/apps/content-editor/src/app/components/Editor/Editor.js
geodem127: removed `contentItems ` selector

src/apps/content-editor/src/app/components/Editor/Editor.js
geodem127: Agreed. That's a much better approach. I've moved the locale-update logic into the relationship field component as suggested.

src/apps/content-editor/src/app/components/Editor/Editor.js
```diff
@@ -1,5 +1,5 @@
 import { memo, useCallback, useEffect, useMemo, useState } from "react";
-import { useDispatch } from "react-redux";
+import { useDispatch, useSelector } from "react-redux";
```
agalin920, line 2: unused import

src/shell/components/RelationalFieldBase/index.tsx
```diff
@@ -67,6 +78,50 @@ export const RelationalFieldBase = ({
...
+    value,
+    contentItemZUID,
+    languages,
+    langID,
+    relatedModelZUID,
+    name,
+  ]);
+
+  useEffect(() => {
+    updateItemLocale();
```
agalin920, line 122: This seems like a very hacky solution to this. If im reading this correctly you are changing the content's state and then manually unmarking it as dirty on mount? Are we also changing values? If so why? We should not be doing any kind of content change that is not explicit by the user

A more holistic solution is when switching locale the value of the initial state should reflect the locale.

agalin920 (changes requested):

src/shell/components/RelationalFieldBase/index.tsx
geodem127: The content of all locale items will be identical, as they inherit the same content upon creation. 

I’m not sure if we have control over that behavior on the frontend.

agalin920:
Based on discussions @geodem127 we should close or mark the pr as draft. Don't want this accidentally merged in

Issue #3873, opened 2025-10-24
Global search api call causes instances api service to crash


finnar-bin:
## Problem

When loading manager-ui on specific instances with huge amounts of content items the instances api service crashes and restarts. 

Upon investigation, on load there is an api call to `/search/items?q=&limit=10000`, the 10k limit is quite huge, there are times that the api eventually responds and the raw resource size is 100mb+. Refreshing this instance several times would sometimes cause the instances api service to crash and restart.

## Screenshots

<img width="983" height="200" alt="Image" src="https://github.com/user-attachments/assets/a7644102-aa64-4a88-b268-7f3a8296259b" />

## Steps to reproduce

1. Go to a customer instance with a lot of content items
2. Check network tab, on manager-ui load, there is an api call to the search endpoint with 10k limit
3. Leave it to try and resolve the response for a minute or so
4. Refresh the page then leave it again for a minute or so
5. Doing this over and over would eventually take down the service and the api calls to the instances api end up in CORS errors

## Expected behavior

The limit should be lowered down as 10k is too much

agalin920:
@finnar-bin  this only occurs on dev

PR #3874, opened 2025-10-24
Create redirects lag

finnar-bin:
Issues this PR is trying to resolve:

- When the create redirect modal is opened, it hits the search endpoint with a 10k limit which is pretty huge. This sometimes causes the instances api service to crash and restart
- Once the dropdown opens the UI lags so bad
- Typing to filter is very laggy

Things done so far:

- Changed the search query limit to 5
- Instead of doing a client side filter, I changed it to hit the search endpoint when a keyword is typed in, then the results are mapped as the options
- Memoized some components in the autocomplete

Issues that need to be resolved:

- When editing a redirect, there are instances where it fails to load the data of the previous value since it currently does a client side filter to find the value from the options, but since the options has been limited to 5, it sometimes fails to display the selected item

Things to note:

- Regarding the issue with the 10k limit that's sometimes causing the instances api service to crash, there's also [another ticket](https://github.com/zesty-io/manager-ui/issues/3873) where the same query is being performed but on a different component which also causes the same problem.

### DIFF
--- a/src/apps/content-editor/src/app/views/Redirects/ContentRedirectModal.tsx
+++ b/src/apps/content-editor/src/app/views/Redirects/ContentRedirectModal.tsx
@@ -32,6 +32,7 @@ export type ContentRedirectModalProps = {
   loading: boolean;
   currentItem: ContentItemProps | null;
   options: ContentItemProps[];
+  onSearch: (value: string) => void;
 };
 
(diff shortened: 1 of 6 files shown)


shrunyan:
closed in favor of https://github.com/zesty-io/manager-ui/pull/3875

PR #3882, opened 2025-10-30
Add SEO disable preview setting

### LINKED ISSUE #3866: Settings - Custom SEO preview setting
## Current behavior

Slack conversation for reference:
https://zesty-io.slack.com/archives/C03TKNN7XRN/p1759772472152959

Issue:
In the manager UI, the first uploaded image on a model is used to fill the SEO / Meta Image field. However, if a user has a custom header, they do not see the correct OG image in the preview on the right side of the SEO tab.

## Proposed change

Create an toggle in settings where developers can turn off the SEO preview, and we can notify customers of this setting if/when they ask about previews being inaccurate (or during onboarding)

## Design

Meant to match the current toggles in settings, and only shown when we enable it for a specific instance.

agalin920:


geodem127 (approved):

src/apps/content-editor/src/app/views/ItemEdit/Meta/index.tsx
```diff
@@ -146,6 +149,9 @@ export const Meta = forwardRef(
       useState<(typeof FlowType)[keyof typeof FlowType]>(null);
     const metaDescriptionButtonRef = useRef(null);
     const metaTitleButtonRef = useRef(null);
+    const showSEOPreview =
+      instanceSettings?.find((setting) => setting.key === "disable_seo_preview")
```
shrunyan, line 153: Is this logic going to work when no setting is present? Will the JSX evaluation see `undefined` as falsey?

agalin920: @shrunyan  yes tested this scenario. Reason is because this is **always** showing seo preview unless this setting is set to 1

finnar-bin (approved):

PR #3883, opened 2025-10-31
[Content]: Publish status chip in data table not displaying correctly

### LINKED ISSUE #3818: Content - Publish status chip in data table not displaying correctly

geodem127:
## RCA: 
The initial implementation encountered API limitations when retrieving publishing records

## FIX:
- [x] Created a new API endpoint to fetch publishing records per model.
- [x] Replaced the front-end Redux action fetchItemPublishings with fetchModelItemsPublishings to use the new endpoint.

## Prerequisite:
This change is dependent on the instances-api PR being merged first.
PR: https://github.com/zesty-io/instances-api/pull/689

### DIFF
--- a/src/shell/store/content.js
+++ b/src/shell/store/content.js
@@ -977,7 +977,7 @@ export function fetchItemPublishings() {
   return (dispatch) => {
     return dispatch({
       type: "FETCH_RESOURCE",
-      uri: `${CONFIG.API_INSTANCE}/content/items/publishings?limit=20000`,
+      uri: `${CONFIG.API_INSTANCE}/content/items/publishings?limit=80000`,
       handler: (res) => {
         if (res.status === 200) {
           dispatch({
@@ -1016,7 +1016,6 @@ export function fetchModelItemsPublishings({
   showActiveOnly = false,
 }) {
   return (dispatch) => {
-    console.debug("dispatch", { dispatch });
     return dispatch({
       type: "FETCH_RESOURCE",
       uri: `${CONFIG.API_INSTANCE}/content/models/${modelZUID}/items/publishings?limit=${limit}&showDeleted=${showDeleted}&showDeletedItems=${showDeletedItems}&showActiveOnly=${showActiveOnly}`,
--- a/src/shell/store/content.js
+++ b/src/shell/store/content.js
@@ -1010,67 +1010,61 @@ export function fetchItemPublishings() {
   };
 }
 
-export function fetchModelItemsPublishings({
+export function fetchAllModelPublishings({
   modelZUID,
-  limit = 10000, //set 10k limit as optimal threshold for 2-second response times
-  showDeleted = false,
-  showDeletedItems = false,
-  showActiveOnly = false,
+  limit = 10000, // 10k limit for optimal 2-second response times
+  showDeleted = null,
+  showDeletedItems = null,
+  showActiveOnly = null,
 }) {
-  const params = new URLSearchParams({
-    limit: limit.toString(),
-    showDeleted: showDeleted.toString(),
-    showDeletedItems: showDeletedItems.toString(),
-    showActiveOnly: showActiveOnly.toString(),
-  });
-
   return async (dispatch) => {
-    const allData = [];
-
     try {
-      //paginate requests to prevent large dataset performance issues
-      for (let page = 1; ; page++) {
-        params.set("page", page.toString());
+      const allData = [];
+      let currentPage = 1;
+      while (true) {
+        const params = new URLSearchParams({
+          limit,
+          page: currentPage,
+          ...(showDeleted && { showDeleted }),
+          ...(showDeletedItems && { showDeletedItems }),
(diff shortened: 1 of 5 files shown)


src/shell/store/content.js
```diff
@@ -977,7 +977,7 @@ export function fetchItemPublishings() {
   return (dispatch) => {
     return dispatch({
       type: "FETCH_RESOURCE",
-      uri: `${CONFIG.API_INSTANCE}/content/items/publishings?limit=100000`,
+      uri: `${CONFIG.API_INSTANCE}/content/items/publishings?limit=20000`,
```
agalin920: Can we add a comment to clearly delineate this is the upper bound.

How as this upper bound determined?

src/shell/store/content.js
```diff
@@ -1008,6 +1008,51 @@ export function fetchItemPublishings() {
     });
   };
 }
+export function fetchModelItemsPublishings({
+  modelZUID,
+  limit = 10000,
```
agalin920: Same here.

src/shell/store/content.js
```diff
@@ -1008,6 +1008,51 @@ export function fetchItemPublishings() {
...
+  showDeleted = false,
+  showDeletedItems = false,
+  showActiveOnly = false,
+}) {
+  return (dispatch) => {
+    console.debug("dispatch", { dispatch });
```
agalin920: debug log

agalin920 (changes requested):

src/shell/store/content.js
finnar-bin: Has this limit been tested on a big instance? Just wanted to make sure it isn't causing any performance issues

src/shell/store/content.js
```diff
@@ -1008,6 +1008,51 @@ export function fetchItemPublishings() {
...
+    return dispatch({
+      type: "FETCH_RESOURCE",
+      uri: `${CONFIG.API_INSTANCE}/content/models/${modelZUID}/items/publishings?limit=${limit}&showDeleted=${showDeleted}&showDeletedItems=${showDeletedItems}&showActiveOnly=${showActiveOnly}`,
```
finnar-bin: I would suggest creating a new URLSearchParams to store all the search params as an object then using the `toString()` to make it more readable and easier to maintain as there's several params involved

finnar-bin (changes requested):

src/shell/store/content.js
geodem127: Reduce the initial load to 10k since model-scoped fetching on navigation has already been implemented.

src/shell/store/content.js
geodem127: updated. added comment

src/shell/store/content.js
geodem127: removed

src/shell/store/content.js
geodem127: updated

src/shell/store/content.js
```diff
@@ -1009,6 +1010,71 @@ export function fetchItemPublishings() {
...
+  limit = 10000, //set 10k limit as optimal threshold for 2-second response times
+  showDeleted = false,
+  showDeletedItems = false,
+  showActiveOnly = false,
+}) {
+  const params = new URLSearchParams({
+    limit: limit.toString(),
+    showDeleted: showDeleted.toString(),
+    showDeletedItems: showDeletedItems.toString(),
+    showActiveOnly: showActiveOnly.toString(),
```
finnar-bin: Is adding `.toString()` on each value necessary?

geodem127: No, .toString() is not necessary. The PR has been updated to remove these

agalin920 (approved):

finnar-bin (approved):

PR #3891, opened 2025-11-07
[Global]: search api call causes instances api service to crash

### LINKED ISSUE #3873: Global search api call causes instances api service to crash

geodem127:
## RCA:
The search API was configured with an excessively high result limit (10,000) and no search term. This caused the request to fetch an unmanageably large dataset on page load, leading to frequent timeouts and failed responses.

## FIX:
- [ ] Removed the automatic API call on page load.
- [ ] Implemented a debounced search that only triggers on valid user input.
- [ ] Added sensible pagination/default limits to all search requests.
- [ ] Refactored the logic to leverage the Redux store for efficient state management.

https://github.com/user-attachments/assets/9f2551a1-12fc-4678-aa33-878e09dc9886


src/shell/components/GlobalSearch/index.tsx
```diff
@@ -128,35 +131,57 @@ export const GlobalSearch = () => {
...
-    isFetchingBlocksResults;
+    isFetchingContentSearchResults;
+
+  useEffect(() => {
+    //Update models in store to provide access to other search hooks
+    dispatch(fetchModels());
+  }, []);
+
+  useEffect(() => {
+    if (isEmpty(apiQueryTerm)) return setSearchedContents([]);
```
finnar-bin: Prefer block scope

src/shell/components/GlobalSearch/index.tsx
```diff
@@ -128,35 +131,57 @@ export const GlobalSearch = () => {
...
+    //Update contents in store to provide access to other search hooks
+    dispatch(searchItems(apiQueryTerm))
+      //@ts-ignore
+      .then((res) => {
+        if (!!res?.data?.length) {
+          setSearchedContents(res?.data);
+        } else {
+          setSearchedContents([]);
+        }
+        setIsFetchingContentSearchResults(false);
```
finnar-bin, line 156: Shouldn't this be placed on the `.finally()`?

src/shell/components/GlobalSearch/index.tsx
```diff
@@ -356,7 +380,7 @@ export const GlobalSearch = () => {
     });
     const url = `/search?${searchParams.toString()}`;
 
-    // Do not save term as a recent keyword if it's empty
+    // Do not save term as a recent keyword if it's empty1
```
finnar-bin: Typo at the end

src/shell/hooks/useSearchBlocksByKeyword.ts
```diff
@@ -64,89 +73,145 @@ export const useSearchBlocksByKeyword = (): SearchResult => {
...
-      Object.fromEntries(modelsRaw?.map((model) => [model?.ZUID, model])),
-    [modelsRaw]
-  );
+  const normalizedSearchTerm = searchTerm.toLowerCase()?.trim();
+  const showAll = blockKeywords.includes(normalizedSearchTerm);
 
-  const normalizedSearch = searchTerm.toLowerCase()?.trim();
+  const blocks = useMemo<BlockModel[]>(() => {
+    const modelsArray = models ? Object.values(models) : [];
+    if (!modelsArray?.length) return [];
```
finnar-bin: Prefer block scope

src/shell/hooks/useSearchBlocksByKeyword.ts
```diff
@@ -64,89 +73,145 @@ export const useSearchBlocksByKeyword = (): SearchResult => {
...
+
+  useEffect(() => {
+    // Fetch all block models and variants if user searches for "blockKeywords"
+    if (showAll && !!blocks?.length) {
+      setIsFetchingVariants(true);
+      Promise.all(
+        blocks?.map((block) =>
+          dispatch(fetchItems(block?.ZUID, { limit: 1000 }))
+        )
+      ).then(() => {
```
finnar-bin: I would place this inside the `finally()` to make sure the flag still gets set to false if this fails

src/shell/views/SearchPage/SearchPage.tsx
```diff
@@ -48,17 +48,14 @@ export const SearchPage: FC = () => {
   const keyword = params.get("q") || "";
   const instanceId = useSelector((state: any) => state.instance.ID);
   const ecoId = useSelector((state: any) => state.instance.ecoID);
+  const allModels = useSelector((state: any) => state.models);
```
finnar-bin: Use the `AppState` type instead of any

finnar-bin (changes requested):

src/shell/components/GlobalSearch/index.tsx
geodem127: Updated

src/shell/components/GlobalSearch/index.tsx
geodem127: Updated

src/shell/components/GlobalSearch/index.tsx
geodem127: Updated

src/shell/hooks/useSearchBlocksByKeyword.ts
geodem127: Updated

src/shell/hooks/useSearchBlocksByKeyword.ts
geodem127: Updated

src/shell/views/SearchPage/SearchPage.tsx
geodem127: Updated

finnar-bin:
@geodem127 is there any reason why the base branch has been changed? There's currently 300+ changed files

geodem127:
> @geodem127 is there any reason why the base branch has been changed? There's currently 300+ changed files

I'm not sure what happened, but I’ve already switched it back to dev.

finnar-bin (approved):

agalin920 (approved):

shrunyan:
Why is there so much code changed in the PR when the core of it, as descibed, is going from `10000` to `1000`? i.e. I'm expecting a single line change PR.

> no default search term. 

Also not sure what that means. Why would search have a default term? How would we know what a baseline default term is that meets a users expectations.

geodem127:
> Why is there so much code changed in the PR when the core of it, as descibed, is going from `10000` to `1000`? i.e. I'm expecting a single line change PR.
> 
> > no default search term.
> 
> Also not sure what that means. Why would search have a default term? How would we know what a baseline default term is that meets a users expectations.

The other modifications (debounce, filtering block items, and the UI icon) were initially part of the PR but have since been removed. The diff now primarily shows the core architectural change from RTK Query to Redux dispatch.

Just  to clarify the RCA description: "no default search term" was a typo. It should read "no search term specified"

PR #3899, opened 2025-11-17
Cypress: Setup Fixture Seeding - Content Actions

geodem127:
- [x] Created fixtures for content actions.
- [x] Updated the content seeder function to take a path parameter.
- [x] Updated element selector tagging.
- [x] Re-enabled and executed previously skipped tests.

NOTE: updated `InternalLink` component to fetch item if `internalLinkRelatedItem` does not have meta data since it causes the test to fail if it is not yet fetched


finnar-bin (approved):

cypress.config.js
```diff
@@ -7,6 +7,7 @@ module.exports = defineConfig({
   viewportHeight: 1080,
   video: false,
   defaultCommandTimeout: 15000,
+  experimentalInteractiveRunEvents: true,
```
shrunyan: what is this and why is it being added?

cypress.config.js
```diff
@@ -26,6 +27,14 @@ module.exports = defineConfig({
 
         return launchOptions;
       });
+      // Get spec info to be used for the model's label/name
+      on("before:spec", (spec) => {
```
shrunyan: This seems very complicated for what amounts to labeling something

cypress.config.js
geodem127: this has been removed

cypress.config.js
geodem127: this has been removed

cypress/plugins/seeds/content.ts
```diff
@@ -55,12 +53,13 @@ module.exports = function content(config) {
       json.items.map((item, index) => {
         // Append commit id to item labels for spec tracking
         // append timestamp to prevent naming conflicts
-        const itemLabel = `${item.web.metaTitle}-${config.env.COMMIT_ID}-${timeStamp}`;
+        const itemLabel = `E2E: ${item.web.metaTitle} | ${config.env.COMMIT_ID} | ${timeStamp}`;
```
shrunyan, line 56: Why is this label structure changing?

geodem127: The label structure was adjusted solely for better readability.

PR #3905, opened 2025-11-19
[Content] Lazy load comments api call

### LINKED ISSUE #3897: N+1 API Call
Sentry Issue: [MANAGER-UI-2G9](https://zestyio.sentry.io/issues/7015799667/?referrer=github_integration)
|  |  |
| ------------- | --------------- |
| **Offending Spans** | http.client - GET https://accounts.api.zesty.io...

finnar-bin:
Only perform the api call to fetch comment by resource item when the component is in view to prevent firing off multiple api requests once a content item is loaded
Resolves #3897 

[Screencast_20251119_115912.webm](https://github.com/user-attachments/assets/36d4fd84-893a-48c1-b519-63b081da7311)


agalin920 (approved):

shrunyan:
I like the general idea. Would this be better implemented as a custom hook? this way we could apply the technique elsewhere?

finnar-bin:
@shrunyan yeah, that makes sense. I'll do just that.

finnar-bin:
@shrunyan converted to a hook as suggested

geodem127 (approved):

cypress/e2e/schema/field.spec.js
```diff
@@ -829,6 +829,8 @@ describe("Schema: Fields", () => {
 
     cy.wait("@updateField");
     cy.wait("@getFields");
+    cy.reload();
+    cy.wait(3000);
```
shrunyan: why are we adding a wait?

finnar-bin: Removed, that was me testing if adding a reload and wait would fix the issue with the test that suddenly kept failing which we discussed on last week's FE sync.

agalin920:
@finnar-bin https://github.com/streamich/react-use/blob/master/src/useIntersection.ts

finnar-bin:
> @finnar-bin https://github.com/streamich/react-use/blob/master/src/useIntersection.ts

@agalin920 I've updated the code to make use of the already existing intersection hook from react-use. I tried making eratic and fast scrolls and wasn't able to see any major slow downs on the app despite the button being re-rendered everytime due to the intersection observer.

agalin920 (approved):

PR #3915, opened 2025-11-26
Cypress: Setup Fixture Seeding - Content List

geodem127:
- [x] created content list fixture
- [x] enabled all skipped tests
- [x] updated element selectors

NOTE: This work is based on and requires [#3899](https://github.com/zesty-io/manager-ui/pull/3899) to be merged first.


cypress/e2e/content/list.spec.js
```diff
@@ -1,21 +1,34 @@
...
-    cy.get(".MuiDataGrid-cell").contains("Turkey Run");
-    cy.getBySelector("MultiPageTableSearchField").type("{selectAll}{del}");
+  it("Filters list items based on search term", function () {
+    cy.getBySelector("MultiPageTableSearchField").type(
+      contentItems[1].data.text
+    );
+    cy.get(".MuiDataGrid-cell").contains(contentItems[1].data.text);
+    cy.getBySelector("MultiPageTableSearchField")
+      .type("{selectAll}{del}")
+      .wait(500);
```
agalin920: Why was this "wait(500)" introduced? Is it required?

cypress/e2e/content/list.spec.js
```diff
@@ -35,49 +48,50 @@ describe("Content List Filters", () => {
     cy.getBySelector("NoResults").should("not.exist");
   });
 
-  it("Sorts list items", () => {
+  it("Sorts list items", function () {
```
agalin920: The 'function' keyword is not consistent across our test suites

cypress/e2e/content/list.spec.js
```diff
@@ -1,21 +1,34 @@
...
-      cy.visit("/content/6-0c960c-d1n0kx");
+      cy.visit(`/content/${Cypress.env("modelZUID")}`);
     });
   });
 
-  it("Filters list items based on search term", () => {
-    cy.getBySelector("MultiPageTableSearchField").type("turkey");
-    cy.get(".MuiDataGrid-cell").contains("Turkey Run");
-    cy.getBySelector("MultiPageTableSearchField").type("{selectAll}{del}");
+  it("Filters list items based on search term", function () {
```
finnar-bin: Why was this changed from an arrow function to a traditional function declaration? I would prefer we keep things consistent unless this change is really necessary.

cypress/e2e/content/list.spec.js
finnar-bin: Same comment here regarding the function declaration.

cypress/e2e/content/list.spec.js
```diff
@@ -35,49 +48,50 @@ describe("Content List Filters", () => {
...
     cy.getBySelector("sortByFilter_default").click();
-    cy.getBySelector("dateCreatedFilterOption").click();
-    cy.get(".MuiDataGrid-cell[data-colindex='3']").contains(
-      "Parent pre selection with fast typing"
-    );
+    cy.getBySelector("createdOnFilterOption").click();
+
+    cy.get(".MuiDataGrid-cell[data-colindex='2']")
+      .should("exist")
+      .contains(contentItems[4].data.text, { matchCase: false });
```
finnar-bin: Why is matchCase being set to false here? I would assume we would want to make an exact match to make sure the UI is behaving correctly

cypress/plugins/seeds/content.ts
```diff
@@ -17,7 +17,6 @@ module.exports = function content(config) {
   const { formatPathPart } = require("../../../src/utility/formatPathPart");
   const { formatName } = require("../../../src/utility/formatName");
   const { getSDK } = require("./utils");
-
```
finnar-bin, line 20: Seems unnecessary, plus I prefer we keep the spacing here to keep things more organized

finnar-bin (changes requested):

cypress/e2e/content/list.spec.js
geodem127: In Cypress, variables stored using cy.wrap need to be accessed via this, which requires using a function() declaration. Since I used a different approach, I’ll be reverting it back to the original declaration.

cypress/e2e/content/list.spec.js
geodem127: updated

cypress/e2e/content/list.spec.js
geodem127: updated

cypress/plugins/seeds/content.ts
geodem127: updated

agalin920 (changes requested):

src/apps/content-editor/src/app/views/ItemList/ItemListTable.tsx
```diff
@@ -450,6 +451,9 @@ export const ItemListTable = memo(
                 variant: "skeleton",
                 noRowsVariant: "skeleton",
               },
+              row: {
+                className: "listItemTableRow",
```
finnar-bin: Is there any reason for introducing a new `className` for the rows?

cypress/e2e/content/list.spec.js
geodem127: this has been removed

cypress/e2e/content/list.spec.js
geodem127: updated

src/apps/content-editor/src/app/views/ItemList/ItemListTable.tsx
geodem127: The className was added to provide a reliable selector for testing table rows.

finnar-bin: @geodem127 why not use data-cy which is the proper cypress convention? I would prefer we follow cypress' convention instead of adding a class name which wouldn't be used for styling purposes.

geodem127: Good call. I've updated the attribute to data-cy to align with Cypress convention

finnar-bin (approved):

agalin920 (approved):

PR #3918, opened 2025-12-01
Cypress: Setup Fixture Seeding - Content Comments

geodem127:
- [x] created content-comments fixture
- [x]  updated element selectors


cypress/e2e/content/comments.js
```diff
@@ -4,25 +4,38 @@ const forceClick = { force: true };
...
+        //Set modelZUID as Cypress env variable for global test access
+        Cypress.env("modelZUID", model?.ZUID);
+        //Set itemZUID as Cypress env variable for global test access
+        Cypress.env("itemZUID", items[0]?.meta?.ZUID);
+        MODEL = model;
+        ITEMS = items;
+        FIELDS = fields;
+      }
+    );
+    cy.waitOn("/v1/content/models**", () => {
```
finnar-bin, line 16: What's with the double asterisk (*)? Is this a typo?

cypress/e2e/content/comments.js
```diff
@@ -4,25 +4,38 @@ const forceClick = { force: true };
...
-    cy.intercept("/v1/comments/*").as("getAllComments");
     cy.get(commentBox, { timeout: 50000 }).should("exist");
-    cy.get(commentBox).focus();
-    cy.get(commentBox).type("This is a new comment.");
+
+    cy.get(commentBox).focus().type("{selectAll}{del}This is a new comment.");
+
     cy.get('[data-cy="SubmitNewComment"]').click();
 
-    cy.wait("@getAllComments");
```
finnar-bin, line 30: Wouldn't this cause an issue if the comment hasn't been posted yet making the test flaky?

finnar-bin (changes requested):

cypress/e2e/content/comments.js
geodem127: I added the double asterisk because /v1/models* would only intercept requests that match the same path, possibly with query parameters or suffixes, but not additional path segments. Using ** ensures that requests to nested routes (e.g. /v1/models/123) are also intercepted.

cypress/e2e/content/comments.js
geodem127: yes it would. I added it back

finnar-bin (approved):

PR #3924, opened 2025-12-04
Blocks - Refetch content model data when a block model is generated

### LINKED ISSUE #3923: Blocks - Newly-created AI-generated block model does not show up on the subapp nav
## Problem

When generating a block model via AI and then clicking Navigate, the new block model does not show up on the blocks subapp nav. User has to refresh the page to see the newly-created block.

## Screenshots

https://github.com/user-attachments/assets/3689d7f9-697e-40f1-8439-7c0431959a32

## Steps to reproduce

1. Go to Content app
2. Open AI Drawer
3. Generate a new block model
4. Click Navigate button
5. See error

## Expected behavior

The generated block model should be present on the blocks subapp nav

finnar-bin:
Ensures that the content model data rtk cache is invalidated when a block model is generated via AI so that the subapp nav will immediately contain the newly-generated block.
Resolves #3923 

[Screencast_20251204_094345.webm](https://github.com/user-attachments/assets/d2772046-f927-482b-b764-794d7d5862d7)

### DIFF
--- a/src/shell/views/Shell/AIDrawer.tsx
+++ b/src/shell/views/Shell/AIDrawer.tsx
@@ -33,7 +33,10 @@ import ArrowDropDownRoundedIcon from "@mui/icons-material/ArrowDropDownRounded";
 import SettingsRoundedIcon from "@mui/icons-material/SettingsRounded";
 import InfoRoundedIcon from "@mui/icons-material/InfoRounded";
 import LanguageRoundedIcon from "@mui/icons-material/LanguageRounded";
-import { useGetLangsMappingQuery } from "../../services/instance";
+import {
+  useGetLangsMappingQuery,
+  useGetContentModelsQuery,
+} from "../../services/instance";
 import {
   codeSystemInstruction,
   contentSystemInstruction,
@@ -111,6 +114,7 @@ export const AIDrawer = () => {
 
   const [geminiGenerate, { isLoading, isError, data: aiResponse }] =
     useGeminiGenerationMutation();
+  const { refetch: refetchContentModels } = useGetContentModelsQuery();
 
   const responsesEndRef = useRef(null);
 
@@ -149,6 +153,11 @@ export const AIDrawer = () => {
               value: response.payload.value,
             },
           });
+        }
+
+        // Makes sure that the subapp sidebar data is refreshed
+        if (response.type === "NAVIGATE") {
+          refetchContentModels();
         }
       });
     } catch (error) {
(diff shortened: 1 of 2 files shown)


src/shell/views/Shell/AIDrawer.tsx
```diff
@@ -150,6 +154,11 @@ export const AIDrawer = () => {
             },
           });
         }
+
+        // Makes sure that the subapp sidebar data is refreshed
+        if (response.type === "NAVIGATE") {
```
agalin920: This means that everytime the model responds with a navigate action it refetches regardless if the user even clicked on the navigate button.

In addition the navigate action is context free it may navigate to settings for example so there are cases where there is no need to refetch content models

agalin920 (changes requested):

src/shell/views/Shell/AIDrawer.tsx
finnar-bin: @agalin920 I've updated it to only perform a refetch when the response path contains "blocks", this way it would only refetch when a new block is generated. I decided to perform the refetch as soon as the response is received from the mcp, this way even if the user was already on the blocks app the sidebar would automatically get updated instead of having to wait for the user to click the Navigate button before doing the refetch.

agalin920: This is a bit too brittle. Let's talk to @mattezra to return the created entities as part of the response so it can be applied to any entity that we may support in the future. 

For now simply ensuring blocks get refreshed when navigating to the blocks app should suffice

finnar-bin: >  return the crested entities as part of the response so it can be applied to any entity that we may support in the future.

@agalin920 I'm not sure I understand what this is for.

Also, are you suggesting the refetch be done once the navigate button is clicked? If so, what if the user decided to not click the navigate button, wouldn't that mean that they won't be able to see the new block item on the subapp sidebar until it gets refetched some other way?

finnar-bin: As discussed, I have moved the refetch call to the onClick of the navigate button

agalin920: @finnar-bin that wasn't really the solution discussed. The solution we want as part of this is to gurantee block model refetch when navigating to the blocks app

finnar-bin: @agalin920 updated the pr

agalin920 (approved):

PR #3926, opened 2025-12-04
Cypress: Setup Fixture Seeding - Content Meta

geodem127:
- [x] Created fixtures for content meta.
- [x]  added seeding task
- [x]  Updated element selector tagging.
- [x]  enabled previously skipped tests


cypress/e2e/content/meta.spec.js
```diff
@@ -1,27 +1,50 @@
...
+    cy.task("seed:content", "fixtures/meta.json").then(
+      ({ model, fields, items }) => {
+        MODEL["child"] = model;
+        FIELDS["child"] = fields;
+        ITEMS["child"] = items;
+        Cypress.env("ChildModelZUID", model?.ZUID);
+        Cypress.env("ChildItemZUID", items[0]?.meta?.ZUID);
+      }
+    );
+  });
```
finnar-bin, line 14: If any of these 2 seeds fail, will this test still proceed?

cypress/e2e/content/meta.spec.js
```diff
@@ -1,27 +1,50 @@
...
+
+  it("Modifies and saves Meta fields", () => {
+    cy.waitOn("/v1/content/models**", () => {
```
finnar-bin, line 17: Why are there 2 asterisks? Is this a typo?

cypress/e2e/content/meta.spec.js
```diff
@@ -1,27 +1,50 @@
...
     cy.get("[data-cy=metaLinkText]").should("exist");
     cy.get("[data-cy=metaTitle]").should("exist");
     cy.get("[data-cy=metaDescription]").should("exist");
     cy.get("[data-cy=metaKeywords]").should("exist");
     cy.get("[data-cy=sitemapPriority]").should("exist");
     cy.get("[data-cy=canonicalTag]").should("exist");
 
-    cy.get("[data-cy=itemParent] .Select").click();
-    cy.get('[data-value="7-f40360-7vcf5h"]').click();
+    cy.get("#parentZUID input").type(Cypress.env("ParentItemZUID"));
```
finnar-bin: We should ideally be using data-cy attribute selectors instead of IDs to keep things robust and follow [cypress best practices](https://docs.cypress.io/app/core-concepts/best-practices#Selecting-Elements).

cypress/e2e/content/meta.spec.js
```diff
@@ -42,40 +65,47 @@ describe("Content Meta", () => {
       .type("new Meta Title");
 
     cy.get("[data-cy=metaDescription]")
-      .find("textarea")
+      .find("textarea:eq(0)")
```
finnar-bin: Prefer to use cypress `first()`

cypress/e2e/content/meta.spec.js
```diff
@@ -42,40 +65,47 @@ describe("Content Meta", () => {
...
       .click()
       .clear()
       .type("new Meta Description");
 
     cy.get("[data-cy=metaKeywords]")
-      .find("textarea")
+      .find("textarea:eq(0)")
```
finnar-bin: Prefer to use cypress `first()`

cypress/e2e/content/meta.spec.js
```diff
@@ -42,40 +65,47 @@ describe("Content Meta", () => {
...
 
     cy.get("[data-cy=canonicalTag] >  div").click();
-    cy.get('[data-value="1"]').last().click();
+    cy.get('[data-option-index="1"]').last().click();
 
     cy.get("[data-cy=canonicalTag] >  div").click();
-    cy.get('[data-value="0"]').last().click();
+    cy.get('[data-option-index="0"]').last().click();
+
+    cy.get("#parentZUID input").clear().wait(500).type("e2e");
```
finnar-bin: We should ideally be using data-cy attribute selectors instead of IDs to keep things robust and follow [cypress best practices](https://docs.cypress.io/app/core-concepts/best-practices#Selecting-Elements).

cypress/e2e/content/meta.spec.js
```diff
@@ -42,40 +65,47 @@ describe("Content Meta", () => {
...
+    cy.get("ul.MuiAutocomplete-listbox").find("li").eq(0).click();
 
     cy.get("#SaveItemButton").click();
-    cy.contains("Saved a new ").should("exist");
+    cy.get("[data-cy=toast]").contains("Item Saved", {
```
finnar-bin: We have a `cy.GetBySelector()` custom command to easily target data-cy attributes.

cypress/e2e/content/meta.spec.js
```diff
@@ -42,40 +65,47 @@ describe("Content Meta", () => {
...
       cy.waitOn("/v1/env/nav", () => {
         cy.waitOn("/v1/search/items*", () => {
-          cy.visit("/content/6-675028-84dq4s/new");
+          cy.visit(`/content/${Cypress.env("ChildModelZUID")}/new`);
         });
       });
     });
 
-    cy.get("#12-7893a0-w4j9gk", { timeout: 5000 }).find("input").type(today);
+    cy.get('[data-cy="field:text"]', { timeout: 5000 })
```
finnar-bin: We have a `cy.GetBySelector()` custom command to easily target data-cy attributes.

cypress/e2e/content/meta.spec.js
```diff
@@ -85,30 +115,38 @@ describe("Content Meta", () => {
...
       });
     });
 
     cy.iframe("#wysiwyg_basic_ifr")
       .click()
       .type(`{selectall}{backspace}meta description`);
-    cy.get("#12-849844-t8v5l6").find("input").type(`meta title ${today}`);
-
-    cy.getBySelector("CreateItemSaveButton").click();
+    cy.get('[data-cy="field:text"]')
```
finnar-bin: We have a `cy.GetBySelector()` custom command to easily target data-cy attributes.

cypress/e2e/content/meta.spec.js
```diff
@@ -85,30 +115,38 @@ describe("Content Meta", () => {
...
 
     cy.waitOn("/v1/content/models*", () => {
-      cy.get('[role="tablist"]').find("button").eq(1).click();
+      cy.getBySelector("CreateItemSaveButton").wait(500).click();
     });
 
-    cy.contains("/page/otherpage/all-field-types/").should("exist");
+    cy.get('.MuiTabs-root:eq(0) [role="tablist"]').find("button").eq(1).click();
+
+    cy.contains("/e2e-", { matchCase: false }).should("exist");
```
finnar-bin: Why is `matchCase: false` necessary here?

finnar-bin (changes requested):

cypress/e2e/content/meta.spec.js
geodem127: Updated. Implemented content seeding for one item.

cypress/e2e/content/meta.spec.js
geodem127: /v1/content/models* would only intercept requests that match the same path, possibly with query parameters or suffixes, but not additional path segments. Using ** ensures that requests to nested routes are also intercepted.

cypress/e2e/content/meta.spec.js
geodem127: update: added data-cy attribute

cypress/e2e/content/meta.spec.js
geodem127: updated

cypress/e2e/content/meta.spec.js
geodem127: updated

cypress/e2e/content/meta.spec.js
geodem127: update: added data-cy attribute

cypress/e2e/content/meta.spec.js
geodem127: update: used cy.GetBySelector

cypress/e2e/content/meta.spec.js
geodem127: update: used cy.GetBySelector

cypress/e2e/content/meta.spec.js
geodem127: update: used cy.GetBySelector

cypress/e2e/content/meta.spec.js
geodem127: update: no longer necessary since i have used a pre-defined content as parent

cypress/e2e/content/meta.spec.js
finnar-bin: I'm not sure why you updated it, I just wanted to know if the test would still continue to run even if any of the 2 seed tasks fail

finnar-bin (approved):

cypress/e2e/content/meta.spec.js
geodem127: I updated the code to use a predefined model for a simpler and more maintainable approach.

To answer your question: If the first seeding fails, the second will also fail, causing the entire test to fail because the second task won't have a required parent model to be nested under.

PR #3927, opened 2025-12-04
Cypress: Setup Fixture Seeding - Content List item table

geodem127:
- [ ] Created fixtures for content List item table.
- [ ]  added seeding task
- [ ]  Updated element selector tagging.


cypress/e2e/content/item-list-table.spec.js
```diff
@@ -1,25 +1,38 @@
...
+        FIELDS = fields;
+      }
+    );
+  });
+
   it("Resolves internal link zuids", () => {
     cy.waitOn("/search/items*", () => {
-      cy.waitOn("/v1/content/models*", () => {
-        cy.visit("/content/6-a1a600-k0b6f0");
+      cy.waitOn("/v1/content/models**", () => {
```
finnar-bin, line 17: Why are there 2 asterisks?

cypress/e2e/content/item-list-table.spec.js
```diff
@@ -1,25 +1,38 @@
...
+        cy.visit(`/content/${Cypress.env("modelZUID")}`);
       });
     });
 
-    cy.getBySelector("SingleRelationshipCell", { timeout: 30000 })
+    // cy.getBySelector("SingleRelationshipCell")
+    cy.get(".MuiDataGrid-row .MuiDataGrid-cell:eq(2)")
```
finnar-bin: We should ideally be using data-cy attribute selectors to keep things robust and follow [cypress best practices](https://docs.cypress.io/app/core-concepts/best-practices#Selecting-Elements).

cypress/e2e/content/item-list-table.spec.js
```diff
@@ -1,25 +1,38 @@
...
       .first()
-      .contains(
-        "5 Tricks to Teach Your Pitbull: Fun & Easy Tips for You & Your Dog!",
-        { timeout: 15_000 }
-      );
+      .contains(ITEMS?.[0]?.web?.metaTitle, { matchCase: false });
```
finnar-bin: Why does matchCase need to be false? Shouldn't we be matching this exactly as to how the content was posted?

finnar-bin (changes requested):

cypress/e2e/content/item-list-table.spec.js
geodem127: /v1/content/models* would only intercept requests that match the same path, possibly with query parameters or suffixes, but not additional path segments. Using ** ensures that requests to nested routes are also intercepted.

cypress/e2e/content/item-list-table.spec.js
geodem127: updated

cypress/e2e/content/item-list-table.spec.js
geodem127: updated

finnar-bin (approved):

PR #3931, opened 2025-12-12
Global Search - Ensure that value exists before handling the selection

### LINKED ISSUE #3907: Error: Cannot trigger notification without a message
Sentry Issue: [MANAGER-UI-2GG](https://zestyio.sentry.io/issues/7044056353/?referrer=github_integration)

```
Error: Cannot trigger notification without a message
  at Z (./src/shell/store/notifications.ts:39:11)
  at <object>.{children#2}.onChange (./src/shell/components/GlobalSearch/index.tsx:379:22)
...
(12 additional frame(s) were not displayed)
```

finnar-bin:
Ensure that value exists before handling the selection
Resolves #3907

### DIFF
--- a/src/shell/components/GlobalSearch/index.tsx
+++ b/src/shell/components/GlobalSearch/index.tsx
@@ -260,10 +260,12 @@ export const GlobalSearch = () => {
         break;
     }
 
-    return consolidatedResults.sort(
-      (a, b) =>
-        new Date(b.updatedAt).getTime() - new Date(a.updatedAt).getTime()
-    );
+    return consolidatedResults
+      .filter((result) => !!result || Object.keys(result).length)
+      .sort(
+        (a, b) =>
+          new Date(b.updatedAt).getTime() - new Date(a.updatedAt).getTime()
+      );
   }, [
     contents,
     models,
--- a/src/shell/components/GlobalSearch/index.tsx
+++ b/src/shell/components/GlobalSearch/index.tsx
@@ -501,10 +501,7 @@ export const GlobalSearch = () => {
             /** This is when the user selects any of the suggestions */
 
             // null represents "X" button clicked
-            if (
-              !newVal ||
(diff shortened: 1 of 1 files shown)


src/shell/components/GlobalSearch/index.tsx
```diff
@@ -524,8 +524,11 @@ export const GlobalSearch = () => {
               }
 
               goToSearchPage(newVal);
-            } else {
-              if (newVal?.url) {
+            } else if (
```
agalin920: this else if is a bit difficult to reason about since it has a nested if/else and there is no matching else.

Can this not be refactored to be more readable? This is very difficult to really read what the core logic here is.

finnar-bin: I've updated the code and moved the handling in cases where the value is an empty object

src/shell/components/GlobalSearch/index.tsx
```diff
@@ -501,7 +501,10 @@ export const GlobalSearch = () => {
             /** This is when the user selects any of the suggestions */
 
             // null represents "X" button clicked
-            if (!newVal) {
+            if (
+              !newVal ||
+              (typeof newVal === "object" && !Object.keys(newVal)?.length)
```
agalin920: Is the conditional operator on Object.keys necessary?

Object.keys(newVal)?.length

I don't think there is a scenario where it doesn't return an array

agalin920 (changes requested):

src/shell/components/GlobalSearch/index.tsx
finnar-bin: @agalin920 The issue happened when there was a result that was just an empty object, I seem to be unable to replicate this anymore but I have updated the fix to filter out any empty objects from the results.

agalin920 (approved):

PR #3944, opened 2026-01-06
Blocks: Save changes on currently opened variant when Save button on preview is clicked

### LINKED ISSUE #3939: Blocks – Saving in the preview window after changing variants saves the last selected variant.
When saving from the preview window after switching block variants, the system saves the last selected variant instead of the originally intended one.

https://github.com/user-attachments/assets/6ad35756-d1d6-4790-94e6-d91ef21c4a4e

finnar-bin:
Keep a reference of the `onSave` and `onClose` to make sure that when invoked, the latest states of these functions are invoked.
Resolves #3939 

[Screencast_20260106_130048.webm](https://github.com/user-attachments/assets/6cfb7b18-5e8c-4e2b-b8e6-bd73ab9f9849)

### DIFF
--- a/src/apps/content-editor/src/app/components/Editor/PreviewMode/PreviewMode.js
+++ b/src/apps/content-editor/src/app/components/Editor/PreviewMode/PreviewMode.js
@@ -121,17 +121,36 @@ export default function PreviewMode(props) {
     route,
   ]);
 
+  // Storing these as a ref to make sure we're always invoking the latest state
+  // of these functions
+  const onSaveRef = useRef(props.onSave);
+  const onCloseRef = useRef(props.onClose);
+
+  useEffect(() => {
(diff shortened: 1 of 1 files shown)


finnar-bin:
@agalin920 so regarding our discussion during call earlier for this PR. I tried both of your suggestions and here are my findings


- Add props.onSave and props.onClose to the useEffect dependency
 
  - While this resolves the main issue, any changes made to the variant (like typing on the fields etc) causes the useEffect to rerun and rebind the event listener, this may present a performance issue.

- Create handler functions like `handleSave()` and `handleClose()` which contain the props

  - This does not resolve the main issue, it still invokes a stale copy of the functions passed in the props

agalin920 (approved):

PR #3950, opened 2026-01-12
Cypress: Setup Fixture Seeding - Content Redirects

geodem127:
- [x] Created fixtures for content redirects.
- [x] Implemented data seeding
- [x] Updated element selector tagging.
- [x] Removed timeouts
- [x] Moved item-publish and redirect-create setup into Cypress tasks (`api:publishItem`, `api:createRedirect`) instead of `cy.apiRequest`
- [x] Merged latest `dev` and resolved conflicts
- [x] Addressed remaining review feedback: fixed an unreachable `.should("have.length", 0)` assertion after `.contains()`, added the missing leading `/` on seeded redirect paths, and documented two minor nits (dead-code `REDIRECT_ITEMS`, hardcoded publish version)

Depends on #3899 — merged.

`Redirect Content Item` / `Stop Content Item Redirect` remain skipped: reproducibly flaky on the shared dev instance due to redirect/publish-indexing lag (the UI's "Currently Being Redirected" state doesn't reliably settle in time), independent of the fixture-seeding change here. See the in-code comment on those tests for details.

Two review threads intentionally left as-is (see inline replies for rationale):
- No `after()` cleanup for seeded models/redirects — the dev instance's nightly sync from prod already wipes all test data daily, and `seed:content`'s per-run UUID naming prevents same-day collisions, so explicit teardown adds API calls without benefit.
- The skipped `"Redirect Content Item"` test still wraps its commands in `.then()` — will refactor to top-level `it` commands when it's re-enabled, since it doesn't currently run in CI.


cypress/e2e/content/redirects.spec.js
```diff
@@ -1,164 +1,105 @@
...
-    cy.getElement(".MuiDataGrid-row").should("have.length", 2);
+    cy.get(".MuiDataGrid-row").should("have.length", 2);
   });
 
   it("should be able to edit a redirect", () => {
-    cy.getElement(".MuiDataGrid-cell")
-      .contains(`/${REDIRECTS[0].path}`, { matchCase: false })
+    cy.get(".MuiDataGrid-cell")
+      .contains(`${CURRENT_CONTENT?.name}/${REDIRECTS[0].path}`, {
+        matchCase: false,
```
finnar-bin: Is there any valid reason why were not matching the case?

cypress/e2e/content/redirects.spec.js
```diff
@@ -1,164 +1,105 @@
...
     cy.getBySelector("RedirectsFieldPath")
       .find("input")
       .clear()
-      .type(EDIT_REDIRECTS.path);
+      .type(`${CURRENT_CONTENT?.name}/${EDIT_REDIRECTS.path}`);
     cy.getBySelector("RedirectsCreateButton").click({ timeout: 15000 });
 
-    cy.contains(`/${EDIT_REDIRECTS.path}`, { timeout: 10000 }).should("exist");
+    cy.contains(`${CURRENT_CONTENT?.name}/${EDIT_REDIRECTS.path}`, {
+      timeout: 10000,
```
finnar-bin: Is this necessary? Cypress already has a default timeout

cypress/e2e/content/redirects.spec.js
```diff
@@ -1,164 +1,105 @@
...
+    }).should("exist");
   });
 
   it("should be able to delete a redirect", () => {
-    cy.getElement(".MuiDataGrid-cell")
-      .contains(`/${REDIRECTS[0].path}/updated`, { matchCase: false })
+    cy.get(".MuiDataGrid-cell")
+      .contains(`${CURRENT_CONTENT?.name}/${REDIRECTS[0].path}/updated`, {
+        matchCase: false,
```
finnar-bin: Same comment here regarding the case matching

cypress/e2e/content/redirects.spec.js
```diff
@@ -168,16 +109,16 @@ describe("Content item redirects", () => {
         "itemZUID"
       )}/redirects`
     );
-    cy.getElement('[data-cy="AddIncomingRedirectButton"]').click();
+    cy.get('[data-cy="AddIncomingRedirectButton"]').click();
```
finnar-bin: We already have a helper function that exists `cy.getBySelector`

cypress/e2e/content/redirects.spec.js
```diff
@@ -1,164 +1,105 @@
...
-    deleteTestContents();
-    deleteTestRedirects();
+    cy.task("seed:content", "fixtures/redirects.json").then(
+      ({ model, items }) => {
+        Cypress.env("contentZUID", model?.ZUID);
+        Cypress.env("itemZUID", items[0]?.meta?.ZUID);
+        CURRENT_CONTENT = model;
+        CONTENT_ITEMS = items;
+        publishItem(model?.ZUID, items[0]?.meta?.ZUID);
+        createTestRedirects(items[0]?.meta?.ZUID, model?.name);
```
finnar-bin: Wouldn't these be better of converted to cypress tasks as well?

finnar-bin (changes requested):

cypress/e2e/content/redirects.spec.js
geodem127: Removed `matchCase: false` here — no valid reason for it. The redirect path is built from `model?.name`, which `formatName()` always lowercases, and `REDIRECTS[0].path` is already lowercase, so case always matches exactly.

cypress/e2e/content/redirects.spec.js
geodem127: Removed the explicit `timeout: 10000` — Cypress's default command timeout (30000ms, set in cypress.config.js) already covers this.

cypress/e2e/content/redirects.spec.js
geodem127: Same fix as the other matchCase comment — removed, since case always matches here.

cypress/e2e/content/redirects.spec.js
geodem127: Migrated all remaining raw `cy.get('[data-cy=...]')` selectors in this file to `cy.getBySelector`.

cypress/e2e/content/redirects.spec.js
geodem127: Moved these into Cypress tasks (`api:publishItem`, `api:createRedirect` in cypress/plugins/seeds/content.ts), following the same pattern as `seed:content`, instead of driving them through `cy.apiRequest` from the spec.

cypress/e2e/content/redirects.spec.js
```diff
@@ -1,131 +1,65 @@
...
+        CURRENT_CONTENT = model;
+        cy.task("api:publishItem", {
+          modelZUID: model?.ZUID,
+          itemZUID: items[0]?.meta?.ZUID,
+        });
+        createTestRedirects(items[0]?.meta?.ZUID, model?.name);
+      }
+    );
+
+    cy.task("seed:content", "fixtures/content.json").then(
```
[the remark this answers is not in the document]
geodem127: Moved this seed call into the "Redirect Content Item" test body itself (it only runs when that test does), so it's no longer created on every pass while the test is skipped. While making this change I also caught and fixed a real bug it exposed: `REDIRECT_ITEMS[0]` was being read synchronously before the `cy.task().then()` callback had populated it — fixed by nesting the dependent commands inside the `.then()`.

cypress/e2e/content/redirects.spec.js
```diff
@@ -1,131 +1,65 @@
...
+        Cypress.env("redirectItemZUID", items[0]?.meta?.ZUID);
+
+        REDIRECT_ITEMS = items;
+
+        cy.task("api:publishItem", {
+          modelZUID: model?.ZUID,
+          itemZUID: items[0]?.meta?.ZUID,
+        });
+      }
+    );
```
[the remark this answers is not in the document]
geodem127: No cleanup needed here — the dev instance's data is reset by a nightly sync, so leaked test records don't persist/pollute beyond that window.

cypress/e2e/content/redirects.spec.js
```diff
@@ -284,147 +202,16 @@ describe("Content item redirects", () => {
...
-      )
-    );
-    if (!allPresent && attempts > 0) {
-      waitForRedirectsIndexed(attempts - 1);
-    }
-  });
-}
-
-function createTestRedirects(ZUID) {
+function createTestRedirects(ZUID, path) {
```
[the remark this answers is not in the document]
geodem127: Same as above — no after-hook cleanup needed since the instance resets nightly, so there's no need to track redirect ZUIDs for deletion.

cypress/e2e/content/redirects.spec.js
```diff
@@ -191,68 +112,67 @@ describe("Content item redirects", () => {
...
-        });
-    });
-
-    awaitRedirectsData(
-      `/content/${Cypress.env("contentZUID")}/${Cypress.env(
-        "itemZUID"
-      )}/redirects`
-    );
-    cy.getBySelector("RedirectContentItemButton").should("be.enabled").click();
+    cy.task("seed:content", "fixtures/content.json").then(
```
finnar-bin, line 123: Does this really need to be updated considering this whole block is skipped anyway?

finnar-bin (commented):
See comment

cypress/e2e/content/redirects.spec.js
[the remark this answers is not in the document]
geodem127: No cleanup added here by design — the dev instance is synced nightly from prod (see CLAUDE.md), which wipes all test data created during the day regardless of whether a spec explicitly deletes it. `seed:content` already appends a per-run UUID to model/content names, and the redirects created by `createTestRedirects` are keyed off that same unique model name, so there's no naming-collision risk within a day either. Adding `after()` teardown here would just be extra API calls for redirects that are already guaranteed to disappear by the next nightly sync — happy to revisit if that assumption changes (e.g. sync cadence changes or this spec starts running against a longer-lived instance).

cypress/e2e/content/redirects.spec.js
geodem127: Agreed — leaving the `.then()`-wrapped structure as-is for now since both tests are currently `it.skip`'d and don't run in CI. I'll refactor to top-level `it` commands (matching the pattern used by the other tests in this file) when these are re-enabled.

agalin920 (approved):

finnar-bin (approved):

PR #3958, opened 2026-01-14
[Media] Fix replace file permission check

### LINKED ISSUE #3936: Media - User cannot "Replace" assets in Media app despite having delete permissions
## Problem

Users with custom roles (and standard "Developer" roles) are unable to replace existing images in the Media app. They receive an error message or restriction when attempting to do so.

Currently, the UI appears to restrict the "Replace" action specifically to users with Admin or Owner role names. This behavior is inconsistent because these same users are able to Delete assets, suggesting they should have sufficient write permissions to replace them as well.

- [Slack conversation link](https://zesty-io.slack.com/archives/C050120G5/p1765298876135699)

cc: @bobakzahedi

agalin920:


src/apps/media/src/app/components/FileModal/ReplaceFileModal.tsx
```diff
@@ -45,10 +45,9 @@ export const ReplaceFileModal = ({
   const { data: currentUserRoles } = useGetCurrentUserRolesQuery();
   const uploads = useSelector((state: AppState) => state.mediaRevamp.uploads);
   const filesToUpload = uploads.filter((upload) => upload.status !== "failed");
-
   const canReplaceImage = currentUserRoles
     ?.filter((role) => role.entityZUID === instanceZUID)
-    ?.some((role) => ["admin", "owner"].includes(role.name?.toLowerCase()));
+    ?.some((role) => role.systemRole.super);
```
finnar-bin: What does `systemRole.super `mean? I thought we were only supposed to allow file replace for users who are also able to delete a file?

agalin920: @finnar-bin pushed wrong commit

finnar-bin (approved):

geodem127 (approved):

PR #3967, opened 2026-01-21
Content: Lazy load fetching of a media item data

### LINKED ISSUE #3934: N+1 API Call
Sentry Issue: [MANAGER-UI-2GR](https://zestyio.sentry.io/issues/7105319260/?referrer=github_integration)
|  |  |
| ------------- | --------------- |
| **Offending Spans** | http.client - https://media-manager.api.zesty.i...

finnar-bin:
Adds lazy loading to the api call to fetch a media item in the media field. While this does not totally resolve the issue in the sentry span logs, this at least prevents a surge of api calls to this endpoint when a content item with multiple media files is opened.
Resolves #3934 

[Screencast_20260121_125204.webm](https://github.com/user-attachments/assets/2eed919b-133d-4acb-9a47-3e8a52bc6fd5)


shrunyan:
Had you consider the [`loading` attribute](https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/loading) and if so why intersection usage over `loading`?

finnar-bin:
@shrunyan the change on this Pr is to add lazy loading to the api calls that fetch a media item's data. It's so we don't immediately make a bunch of api calls to `/media-manager.api.zesty.io/file/*` when a content item is being opened. The api calls are only made when those items are in view and those data will actually be needed.

shrunyan (approved):

PR #3968, opened 2026-01-22
Global: Update mcp domain

### LINKED ISSUE #3966: [AI Drawer] - Update mcp remote client URL
Update mcp-remote-client URL in manager-ui to use the new mcp-remote code:

dev: `https://mcp-remote-387953501748.us-central1.run.app/client`
stage: `https://mcp-remote-109811026457.us-central1.run.app/client`
prod: `https://mcp.content.one/client`

finnar-bin:
Created new rtk query for the new mcp domain
Added new mcp domain config and enabled in CSP
Resolves #3966 

[Screencast_20260122_083514.webm](https://github.com/user-attachments/assets/540150a8-66ed-4f5e-ab5e-1dcb54e45a9f)

### DIFF
--- a/src/shell/store/index.js
+++ b/src/shell/store/index.js
@@ -40,6 +40,7 @@ import { metricsApi } from "../services/metrics";
 import { cloudFunctionsApi } from "../services/cloudFunctions";
 import { marketingApi } from "../services/marketing";
 import { analyticsApi } from "../services/analytics";
+import { mcpApi } from "../services/mcp";
 
 // Middleware is applied in order of array
 const middlewares = [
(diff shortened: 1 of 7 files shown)


src/shell/store/index.js
```diff
@@ -55,9 +56,9 @@ const middlewares = [
   cloudFunctionsApi.middleware,
   mediaManagerApi.middleware,
   metricsApi.middleware,
-  cloudFunctionsApi.middleware,
```
agalin920, line 58: If we are removing it from the middleware should we just delete it?

agalin920 (approved):

src/shell/store/index.js
finnar-bin: @agalin920 Oh I'm not removing this completely, this is just a duplicate I removed since it's also already added on line 56

finnar-bin: But yeah, I just realized that I haven't removed the old query on cloudfunctions, just removed it now

PR #3986, opened 2026-02-05
Content: Also delete siblings of a deleted content item in a multi-lang instance

### LINKED ISSUE #3980: Deleted items not purged from in-memory cache
## Issue
After deleting an item from a multilang instance, which all of its siblings got deleted too, they will still be visible when on a relational field's dropdown menu where they are being referenced.

## Reproduce
1. Use a multilang instance
2. Delete an item that is being referenced by a relational field
3. Navigate to the relational field, and the items deleted are still visible in the dropdown list

finnar-bin:
When deleting a content item in a multi-lang instance, also delete the siblings that are on a different language from the cache.
Resolves #3980 

[Screencast_20260205_090332.webm](https://github.com/user-attachments/assets/d9eaac20-155a-4830-9fa0-a962a9e3b03b)

### DIFF
--- a/src/shell/store/content.js
+++ b/src/shell/store/content.js
@@ -128,7 +128,23 @@ export function content(state = {}, action) {
 
     case "REMOVE_ITEM":
       let removed = { ...state };
+      const itemToRemove = removed[action.itemZUID];
+      // Get the item ZUIDs of all the siblings for multi-lang instances
+      const siblingItems = itemToRemove?.siblings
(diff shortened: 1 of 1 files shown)


agalin920 (commented):
@finnar-bin is this always the case BE wise? If any locale gets deleted it will automatically delete all siblings?

finnar-bin:
@agalin920 yep, I confirmed it with @markelmad yesterday just to make sure.

agalin920 (approved):

PR #3994, opened 2026-02-09
Cypress: Code App - Seeding Task

geodem127:
- [x] added code app data seeding task
- [x] updated plugin
- [x] updated type decleration
- [x] fixed missing content-type header on the script fetch request
- [x] fixed fragile/incorrect getAuthToken() call (now reuses sdk.token)
- [x] merged latest dev
- [x] added `cleanup:code` task and teardown hook for seeded views/scripts/stylesheets
- [x] fixed `ActionButton` dropping `data-cy` in the inactive/disabled render branch

Note: Had to use native fetch for creating script files since the SDK doesn't support that yet

Test fixtures and e2e coverage for this task will be added in the PR that actually consumes `seed:code`.


cypress/plugins/seeds/code.ts
finnar-bin, line 1: What happens if the api call fails?

geodem127: Returns null; doesn't throw.

geodem127: Good catch — now throws instead of silently returning null (cf1740422).

agalin920 (dismissed):

finnar-bin (dismissed):

agalin920:
Checks failing

agalin920 (dismissed):

agalin920:
@geodem127 can we merge this soon? make sure to re run the test if it was skipped

finnar-bin (dismissed):

cypress/e2e/code/sidebar.spec.js
```diff
@@ -1,13 +1,154 @@
...
+    ].filter((file) => file.zuid);
+
+    if (files.length) {
+      cy.task("cleanup:code", files);
+    }
+  });
+
+  it("Displays documents/files correctly", () => {
+    cy.get('[data-cy="html"] [id="html-/code/file/views/__e2e__"]').as(
+      "e2eFolder"
```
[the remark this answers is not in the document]
geodem127: Verified against the live DOM (dumped actual tree-item ids via a diagnostic Cypress spec, not just static reading) — this locator is actually correct. `buildNavTree`/`resolveNavData` in `navCode.js` do give directory nodes a raw `path` (e.g. `__e2e__`), but `SideBar/index.tsx`'s `createTreeData` re-prefixes directory paths with `` `/code/file/${group}/` `` before they reach `NavTree`/`getItemId`:

```ts
const itemLocation: string = `/code/file/${group}`;
const filePath: string = isDir
  ? `${itemLocation}/${itemPath}`
  : `/${itemPath}`;
```

Confirmed the rendered id is `html-/code/file/views/__e2e__`, matching this spec's original selector. Ran `sidebar.spec.js` against the dev instance — 5/5 passing unmodified. No change needed here.

cypress/e2e/code/sidebar.spec.js
```diff
@@ -1,13 +1,154 @@
...
+  it("Displays a seeded stylesheet under SITE.CSS", () => {
+    expect(STYLESHEET_FILENAME).to.not.be.empty;
+    findSeededFileNode("css", "stylesheets", Cypress.env("stylesheetZUID"))
+      .scrollIntoView()
+      .should("be.visible");
   });
 });
+
+function findSeededFileNode(navId, group, zuid) {
+  cy.get(`[data-cy="${navId}"] [id="${navId}-/code/file/${group}/__e2e__"]`).as(
```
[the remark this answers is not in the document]
geodem127: Same root-cause analysis as the `__e2e__` locator above — see that reply. `createTreeData` in `SideBar/index.tsx` re-prefixes directory paths with `` `/code/file/${group}/` `` for every `FileNav` instance (views/css/js), so `findSeededFileNode`'s `` `${navId}-/code/file/${group}/__e2e__` `` selector is correct as written. Verified via a diagnostic spec dumping live DOM ids, and by running the full spec (5/5 passing) unmodified.

geodem127:
Addressed review feedback (`aa4961d36`):

**Fixed:**
- `cypress/plugins/seeds/code.ts` — replaced `Date.now()` with `uuidv4()` for filename uniqueness, per CLAUDE.md's requirement to avoid timestamp collisions across parallel CI runners (matches `content.ts`'s existing pattern).
- `cypress/index.d.ts` — removed the `Partial<>` wrapping on `seed:code`'s return type; the implementation always returns a full `WebView | Script | Stylesheet` or throws, so `Partial<>` just forced unnecessary null-checks on callers.

**Investigated and declined** — the latest review's 2 blocking findings (folder-locator id mismatch in `sidebar.spec.js`, lines 49 and 132):
Traced the actual DOM by seeding real data and dumping the live tree-item ids in a throwaway diagnostic spec. `buildNavTree`/`resolveNavData` in `navCode.js` do give directory nodes a raw `path` (e.g. `__e2e__`), but `SideBar/index.tsx`'s `createTreeData` re-prefixes directory paths with `` `/code/file/${group}/` `` before they reach `NavTree`. Confirmed rendered id is `html-/code/file/views/__e2e__`, matching the spec's original selectors exactly. Ran `sidebar.spec.js` against the dev instance both before and after — 5/5 passing, unmodified. Replied inline on both threads with the trace.

Other repeatedly-flagged items (retry logic on seed calls, `.MuiTreeItem-content` class selectors) are unchanged — advisory only, not part of the current blocking set, and would require broader component changes out of scope here.

`npx tsc --noEmit` clean.

agalin920 (approved):

PR #4020, opened 2026-03-16
Content | Schema: Repeater Field

### LINKED ISSUE #3991: Manager-UI - Repeater Field
## Problem
Users have requested a way to add repeatable groups of fields to their content items. To simplify the process of adding content such as:
- Slides
- Clients
- Events

## Solution
We are introducing the repeater field to our schema and content apps. As a way to define and add a group of fields to a content item. These will be saved as JSON data, customers can reference with a Parsley iterable.

## Additional notes

Add any other context about the feature request here.

finnar-bin:
**Changes:**
- Added a new field type in schema for the field repeater
- Allow users to add rows to the repeater field in the content app
Resolves #3991 

**Dependency:**
- https://github.com/zesty-io/material/pull/121
- https://github.com/zesty-io/instances-api/pull/730

**Preview**

[Screencast_20260317_143201.webm](https://github.com/user-attachments/assets/2173a1e8-c00d-46bc-8e62-f2fa0dd068c2)


src/shell/components/FieldTypeRepeater/index.tsx
```diff
@@ -0,0 +1,419 @@
...
+      {
+        ...GRID_REORDER_COL_DEF,
+        width: 28,
+        minWidth: 28,
+      },
+      ...mappedColumns,
+    ];
+  }, [baseColumns, rowSelectionModel]);
+
+  const calculatedHeight = useMemo(() => {
```
agalin920: This seems very complex. What is the purpose of this? Would Autosizer or react-window not resolve what you are attempting to achieve?

src/shell/components/FieldTypeRepeater/index.tsx
```diff
@@ -0,0 +1,419 @@
...
+  };
+
+  return (
+    <>
+      <Box
+        sx={{
+          height: calculatedHeight,
+          position: "relative",
+
+          // An ancestor has removed all scrollbars so we're re-enabling them here
```
agalin920, line 308: Can you explain this a bit more?

src/apps/schema/src/app/views/Model.tsx
```diff
@@ -63,10 +64,12 @@ export const Model = () => {
         path="/schema/:id/fields/:fieldId"
         render={() => {
           return (
-            <AddFieldModal
-              mode="update_field"
-              onModalClose={() => history.push(`/schema/${id}/fields`)}
-            />
+            <VisibilityProvider>
```
agalin920: This seems like a bandaid solution that seems like tech debt. Its naming is also very generic. What is the main purpose of this?

src/apps/schema/src/app/components/AddFieldModal/index.tsx
```diff
@@ -59,6 +61,11 @@ export const AddFieldModal = ({ onModalClose, mode, sortIndex }: Props) => {
       fullScreen={viewMode === "fields_list"}
       sx={{
         my: "20px",
+        // This is used when adding a sub-field in a repeater field,
+        // that flow opens a new modal and we want this modal to still remain open
+        // so context is not lost while being hidden so it doesn't interfere
+        // with the other modal
+        display: isHidden ? "none" : "block",
```
agalin920: This comment doesnt really explain this. It says we dont want to hide anything but this seems like it is adding hidden functionality

agalin920 (changes requested):

src/shell/components/FieldTypeRepeater/index.tsx
finnar-bin: There's [this ancestor that did a * selector](https://github.com/zesty-io/manager-ui/blob/b78f622718672d581ad4175b495aec37841ec08b/src/apps/content-editor/src/app/views/ItemEdit/Content/Content.js#L92-L99) to remove the scrollbar on every child component but since we needed the scrollbar on this table I had to override it locally.

src/apps/schema/src/app/views/Model.tsx
finnar-bin: So, what's going on here is that the when you add/edit a schema field it opens the `<AddFieldModal />` which is a dialog. When the opened field is a repeater field and they click add/edit subfield it opens another dialog on top of the `<AddFieldModal />`, since we want the user to focus on the newly opened dialog sitting on top of the `<AddFieldModal />`  there needs to be a way to "hide" this. I chose to not unmount the `<AddFieldModal />` when the other dialog is opened since all the state context is on `<AddFieldModal />` and we do not want to lose it once the dialog for the subfield add/edit is closed.

src/apps/schema/src/app/components/AddFieldModal/index.tsx
finnar-bin: This comment is the same explanation to my reply above. Instead of unmounting the `<AddFieldModal />` when the subfield add/edit dialog is opened on top of it, I just hid this modal in order to not lose all the state context which is stored in this component. `<AddFieldModal />` just "hides" when the subfield add/edit dialog is opened on top of it.

src/shell/components/FieldTypeRepeater/index.tsx
finnar-bin: I removed this runtime height calculation with a more css-aligned solution to achieve the same UI need

agalin920 (approved):

src/apps/schema/src/app/views/Model.tsx
finnar-bin: As discussed last week I removed the unnecessary hiding of the parent modal

src/apps/schema/src/app/components/AddFieldModal/index.tsx
finnar-bin: As discussed last week I removed the unnecessary hiding of the parent modal

finnar-bin:
Note: Some tests are currently failing due to the backend changes to support the repeater field has not yet been deployed

geodem127 (approved):

PR #4025, opened 2026-03-19
Blocks: Support rendering of block variants without names

### LINKED ISSUE #3989: TypeError: Cannot read properties of null (reading 'toLowerCase')
Sentry Issue: [MANAGER-UI-2HQ](https://zestyio.sentry.io/issues/7235130729/?referrer=github_integration)

```
React ErrorBoundary TypeError: Cannot read properties of null (reading 'toLowerCase')
  at BlockModel (./src/apps/blocks/views/BlockModel.tsx:17:20)
  at None (div)
  at None (G$)
  at None (div)
  at None (div)
...
(37 additional frame(s) were not displayed)

TypeError: Cannot read properties of null (reading 'toLowerCase')
  at <object>.{children#2}.children (./src/apps/blocks/views/BlockModel.tsx:86:56)
  at Array.filter (<anonymous>)
  at BlockModel (./src/apps/blocks/views/BlockModel.tsx:86:20)
...
(8 additional frame(s) were not displayed)
```

finnar-bin:
Fixes #3989

### Changes:
- Prevents app from crashing when a block variant doesn't have a meta title/variant name
- Allow users to create block variants with no variant name in the UI

### Preview:

[Screencast_20260319_092154.webm](https://github.com/user-attachments/assets/2ee8e219-7a8b-4cb7-ad7a-90ddfcc67c66)

### DIFF
--- a/cypress/e2e/blocks/tests/blocks.spec.js
+++ b/cypress/e2e/blocks/tests/blocks.spec.js
@@ -85,6 +85,19 @@ describe("All Blocks Tests", () => {
       timeout: 15_000,
     }).should("exist");
   });
+
+  it("creates a variant with no variant name", () => {
+    AllBlocksPage.visit();
+    cy.contains(CypressTestBlock).click(TIMEOUT);
+    cy.getBySelector("create-variant-button-header").click();
+    cy.getBySelector("variant-name-input").clear();
+    cy.getBySelector("create-variant-confirm-button").click();
+    cy.contains(new RegExp(`${CypressTestBlock}:\\s*`), {
+      timeout: 15_000,
+    }).should("exist");
+    cy.contains(CypressTestBlock).click(TIMEOUT);
+    cy.getBySelector("block-variant-card").should("have.length", 2);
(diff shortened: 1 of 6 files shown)


cypress/e2e/blocks/tests/blocks.spec.js
```diff
@@ -85,6 +85,19 @@ describe("All Blocks Tests", () => {
       timeout: 15_000,
     }).should("exist");
   });
+
+  it("creates a variant with no variant name", () => {
```
agalin920, line 89: What is the need to allow the creation of a variant with no variant name? Does product no longer want this enforced?

finnar-bin: @bobakzahedi [decided that we should maintain BE parity](https://github.com/zesty-io/manager-ui/issues/3989#issuecomment-4067584923), so since the api doesn't require it it shouldn't be required in the ui as well.

agalin920 (approved):

PR #4028, opened 2026-03-25
 [Content]: Schedule Unpublish Button

### LINKED ISSUE #3151: Content: Schedule Unpublish Button
# Problem 
Currently, there is no way in the platform for users to set a specific date or time for when a content item should automatically be unpublished. This limits control over content availability, especially for time-sensitive materials such as promotions, events, or seasonal content.

Without the ability to schedule when content should be unpublished, users must manually monitor and remove outdated content, leading to potential mistakes and inefficiencies.

## Opportunity
Users need a feature that allows users to schedule when content should be unpublished. This would provide users with more control over content lifecycle management, streamline workflows, and reduce manual effort in managing content availability.

# Solution

**1. Add Unpublish Scheduling Option During Content Publishing**
When a user publishes a content item, we should prompt them with an option to set an unpublish date. This would allow the user to specify when the content should automatically be unpublished after a set period.

**2. Add Schedule Unpublish Option to the Content Item View**
After publishing, users should be able to view and edit the unpublish date for a content item in the view content item view. This allows users to modify the unpublish date later if necessary.

** Future Idea: Bulk Unpublish Scheduling**
Allow users to select multiple content items and apply an unpublish date in bulk. This is especially useful for time-sensitive campaigns or events where many items need to be unpublished at the same time.

# Figma
https://www.figma.com/design/zOI7oSH3fG1XDmmPzsP6PX/Zesty-3.0-Revamp?node-id=5235-90699&t=oHRL90DLORLZQhbE-1

geodem127:
## What

Adds **Schedule Unpublish** and **Cancel Scheduled Unpublish** flows to the content item editor.

## Changes

**New functionality**
- New menu items in `ItemEditHeaderActions` (published state only): "Schedule Unpublish" / "Unschedule Unpublish"
- `SchedulePublish` now accepts `scheduledAction="unpublish"` prop and routes to a dedicated `ScheduleUnpublishDialog` component
- `ScheduledUnpublishIndicator` badge on the published label (`data-cy="ScheduledUnpublishIndicator"`)
- `PublishStatus` component shows scheduled-unpublish badge when the active version has a future `unpublishAt`
- Item edit header tooltips (saved / published / scheduled) now resolve and display the acting user's name instead of a raw ZUID

**Architecture**
- `SchedulePublish` refactored into a smart container (`index.tsx`) and two dumb presentation components:
  - `SchedulePublishDialog` — publish scheduling flow (exact pre-PR behaviour)
  - `ScheduleUnpublishDialog` — unpublish scheduling flow (new)
- All state, handlers, and computed values live in the smart container; dialogs receive only props
- `TooltipTitle` now takes a resolved `userName` prop instead of `userZUID` + the full `users` list; each caller (`ItemEditHeaderActions`, `PublishStatus`) resolves the name itself

**Bug fixes**
- `unpublish()` thunk now re-throws in `.catch()` so `handleUnscheduleUnpublish` step-2 is correctly gated on step-1 success
- `handleSchedulePublish` / `handleScheduleUnpublish`: `onScheduleSuccess` moved from `.finally()` to `.then()` so it only fires on success, not on API error
- `handleUnschedulePublish`: `.catch()` added to prevent unhandled promise rejection leaking to Sentry (error notification is already handled by the thunk)
- `handleSchedulePublish` / `handleScheduleUnpublish`: `.catch()` added for the same reason
- `hasScheduledUnpublish` wrapped in `!!()` so its type is `boolean` rather than `string | boolean | undefined`
- `Audit` type imported in `ItemEditHeaderActions` to satisfy `noImplicitAny`
- Cancel-unpublish: pre-deletes conflicting scheduled publish before POSTing `unpublishAt: "never"` (avoids API "already has scheduled publish event" error)
- `publish()` thunk notification correctly shows "Cancelled scheduled unpublish for …" for the `publishAt: "now" / unpublishAt: "never"` path (success and error branches)
- `handleUnscheduleUnpublish`: fixed `throw new Error(response?.error?.message)` — `publish()`/`unpublish()` resolve their `.catch()` to `{ error: message }` where `message` is a plain string, so `.message` was always `undefined`. Now throws `response?.error` directly so the real error text is preserved.
- `publish()` thunk `.catch()`: added the missing branch for a failed `handleScheduleUnpublish` call (`publishAt: "now"` + a future `unpublishAt`). It previously fell through to the generic "Error publishing …" toast instead of an unpublish-specific one.
- Fixed a bug flagged by negative-QA: `hasScheduledUnpublish` (`ItemEditHeaderActions`), `isAlreadyScheduledUnpublish` (`SchedulePublish`), and the `PublishStatus` badge all gated visibility on the currently loaded/edited version matching the published version's version number. Saving a new draft bumps `meta.version`, so the badge and "Unschedule Unpublish" menu label disappeared even though the scheduled unpublish was still armed on the published version, with no way to cancel it from the UI. All three now gate on `item.publishing.isPublished`, which reflects the published version's own state independent of what's currently loaded/edited.
- `handleScheduleUnpublish` now mirrors the `hasAnyScheduledPublish` pre-deletion guard already used by `handleUnscheduleUnpublish`: it deletes a conflicting scheduled publish before POSTing the new `unpublishAt` schedule, instead of letting the API reject it with "already has a scheduled publish event." `ScheduleUnpublishDialog` already warns the user that scheduling an unpublish cancels a pending scheduled publish — this makes that promise actually hold. Flagged in review.

**Cypress**
- New specs: _Schedules an item for unpublishing_, _Cancels a scheduled unpublish_
- Each test is self-contained with `cy.visit()` + `cy.wait([items, publishings])`
- "Schedules" test cancels the scheduled unpublish at the end to avoid state leaking into subsequent tests
- "Cancels" test has a defensive cleanup block at the start for robustness across reruns

https://github.com/user-attachments/assets/03fe1af5-cea0-4f4d-bb63-f19e03248a47

> **NOTE**: This PR is required to resolve a failing test https://github.com/zesty-io/manager-ui/pull/4153


src/shell/components/ScheduleUnpublish/index.tsx
agalin920, line 1: Is this not duplicating a ton of stuff we use for SchedulePublish? I would imagine there would be a lot of shared code between them

agalin920 (changes requested):

src/shell/components/ScheduleUnpublish/index.tsx
geodem127: Refactored `SchedulePublish` component to support scheduled un-publishing

src/apps/content-editor/src/app/views/ItemEdit/components/ItemEditHeader/ItemEditHeaderActions.tsx
```diff
@@ -305,7 +313,7 @@ export const ItemEditHeaderActions = ({
   const itemState = (() => {
     if (item?.dirty) {
       return ITEM_STATES.dirty;
-    } else if (item?.scheduling?.isScheduled) {
+    } else if (item?.scheduling?.isScheduled && !item?.publishing?.publishAt) {
```
finnar-bin: why is `item?.publishing?.publishAt` being added? I would assume isScheduled already covers this.

src/apps/content-editor/src/app/views/ItemEdit/components/ItemEditHeader/ItemEditHeaderActions.tsx
```diff
@@ -532,133 +540,192 @@ export const ItemEditHeaderActions = ({
...
+                    users?.find(
+                      (user: any) =>
+                        user.ZUID === activePublishing?.publishedByUserZUID
+                    )?.firstName
+                  }{" "}
+                  {
+                    users?.find(
+                      (user: any) =>
+                        user.ZUID === activePublishing?.publishedByUserZUID
+                    )?.lastName
```
finnar-bin: I would honestly prefer to have this as a variable instead of inline

src/apps/content-editor/src/app/views/ItemEdit/components/ItemEditHeader/ItemEditHeaderActions.tsx
```diff
@@ -532,133 +540,192 @@ export const ItemEditHeaderActions = ({
...
+                  }
```
finnar-bin: Same for this one, prefer to have this as a variable

src/apps/content-editor/src/app/views/ItemEdit/components/ItemEditHeader/ItemEditHeaderActions.tsx
finnar-bin, line 1: I'm aware that this file has been this way from the beginning but I honestly find it hard to follow the rendering logic with all the nested ternary operators. Might be worth looking at refactoring this. Thoughts @agalin920 ?

src/shell/components/SchedulePublish/index.tsx
```diff
@@ -71,29 +73,32 @@ export const SchedulePublish = ({
...
+    setIsLoading(true);
 
     dispatch(
       publish(
         item?.meta?.contentModelZUID,
         item?.meta?.ZUID,
         {
-          publishAt: publishAtUtcStr,
+          publishAt: isForUnpublish ? "now" : publishAtUtcStr,
+          unpublishAt: isForUnpublish ? publishAtUtcStr : "never",
```
finnar-bin: Why do these need to be changed given that this function is supposed to only handle publishing?

src/shell/store/content.js
```diff
@@ -872,9 +872,17 @@ export function publish(modelZUID, itemZUID, data, meta = {}) {
...
+          (!!data.publishAt && data.publishAt !== "now") ||
+          (data.publishAt === "now" &&
+            !!data?.unpublishAt &&
+            data?.unpublishAt !== "never")
+            ? `Scheduled ${title} to ${
+                data.publishAt === "now" && data?.unpublishAt !== "never"
+                  ? "unpublish"
+                  : "publish"
+              } on ${meta.localTime} in the ${meta.localTimezone} timezone`
+            : `Published ${title} now`;
```
finnar-bin: Prefer to have this as an if-else for readability. Kinda hard to follow heavily-nested ternary operations like this.

src/shell/components/SchedulePublish/index.tsx
```diff
@@ -112,13 +117,32 @@ export const SchedulePublish = ({
   const handleUnschedulePublish = () => {
     setIsLoading(true);
 
-    dispatch(
-      unpublish(
+    let actionDispatch = null;
+
+    if (isForUnpublish) {
+      actionDispatch = publish(
```
finnar-bin: I'm not sure why we're doing a publish action when we're supposed to unpublish something?

finnar-bin (changes requested):

src/apps/content-editor/src/app/views/ItemEdit/components/ItemEditHeader/ItemEditHeaderActions.tsx
agalin920: Yea it has grown to have many cases now so if possible a refactor would be good

agalin920 (approved):

src/apps/content-editor/src/app/views/ItemEdit/components/ItemEditHeader/ItemEditHeaderActions.tsx
```diff
@@ -532,133 +540,192 @@ export const ItemEditHeaderActions = ({
...
-                size="small"
-                onClick={(e) => {
-                  setPublishMenu(e.currentTarget);
+            )}
+          </Tooltip>
+          {hasScheduledUnpublish && (
+            <Tooltip
+              title={
+                <div>
+                  v{activePublishing?.version} scheduled to unpublish on{" "}
```
agalin920: Double "on" render here.

The literal "on" is already in the string, so non-Today/Yesterday dates render "scheduled to unpublish on on Mar 30…". The publish-side tooltip uses published (no trailing "on"), which is why that one works. Drop the literal "on" and let the conditional supply it.

agalin920 (changes requested):

src/apps/content-editor/src/app/views/ItemEdit/components/ItemEditHeader/ItemEditHeaderActions.tsx
geodem127: `isScheduled` only covers items scheduled for **publishing**,  meaning they're not yet live. `item?.publishing?.publishAt` is needed to also catch items scheduled for **unpublishing**, which are currently *published* items with a future takedown date. Since those two states are mutually exclusive, `isScheduled` alone wouldn't catch the latter case.

src/shell/components/SchedulePublish/index.tsx
geodem127: Scheduled unpublishing is just publishing with an `unpublishAt` property. Rather than spinning up a separate function to handle scheduled unpublishing, we can extend this one to recognize `unpublishAt` alongside `publishAt`. No need to duplicate the logic.

src/shell/components/SchedulePublish/index.tsx
geodem127: Rather than deleting the existing publish record (which would immediately unpublish the content) and recreating it, we publish again without the `unpublishAt` field, this overrides the current publishing entry in place.

src/apps/content-editor/src/app/views/ItemEdit/components/ItemEditHeader/ItemEditHeaderActions.tsx
geodem127: updated

src/apps/content-editor/src/app/views/ItemEdit/components/ItemEditHeader/ItemEditHeaderActions.tsx
geodem127: updated

src/apps/content-editor/src/app/views/ItemEdit/components/ItemEditHeader/ItemEditHeaderActions.tsx
geodem127: refactored deeply nested ternary operators.

src/shell/store/content.js
geodem127: updated

src/apps/content-editor/src/app/views/ItemEdit/components/ItemEditHeader/ItemEditHeaderActions.tsx
geodem127: updated

src/apps/content-editor/src/app/views/ItemEdit/components/ItemEditHeader/ItemEditHeaderActions.tsx
finnar-bin: I'm kind of confused why the property is called `publishAt` if the item is scheduled for unpublishing.

src/shell/components/SchedulePublish/index.tsx
finnar-bin: I suggest adding a comment on why this is the case in order to avoid confusion

agalin920 (dismissed):

src/apps/content-editor/src/app/views/ItemEdit/components/ItemEditHeader/ItemEditHeaderActions.tsx
geodem127: I added this check to filter out already-published items

src/shell/components/SchedulePublish/index.tsx
geodem127: totally agree. added comment

finnar-bin: I don't see the comment

agalin920:
## QA — Gaps & concerns + suggested Cypress coverage

Verified the core **Schedule Unpublish** flow end-to-end against a live published item (open menu → Schedule Unpublish → set date → schedule → indicator shows → Unschedule → state restored). The `POST …/publishings` payloads are correct (`publishAt:"now"` + `unpublishAt:<UTC>` to schedule, `unpublishAt:"never"` to cancel). The feature works. A few things to address:

> Note: "Solution item 1/2" below refer to the two numbered bullets in issue #3151's **Solution** section — not to other tickets.

### Gaps & concerns
- **Unschedule-Unpublish dialog omits the scheduled date.** The body renders *"v241 is scheduled to unpublish on ⟨blank⟩ in (GMT-08:00) Pacific Time…"*. Root cause: `src/shell/components/SchedulePublish/index.tsx:155` derives `scheduledLocalText` only from `item.scheduling.publishAt`, but a scheduled *unpublish* date lives in `item.publishing.unpublishAt`, so it resolves to `""`. Cosmetic (cancel still works, and the header indicator tooltip is wired to `unpublishAt`), but the confirmation should show the date being cancelled.
- **"Invalid Time" label persists** on the time field even after a valid future time is selected (the UTC preview computes correctly and the confirm button stays enabled). Looks like a stale validation/aria label on `FieldTypeDateTime`'s time sub-field — worth a look as it can confuse users.
- **No edit-in-place** for an existing scheduled unpublish date. Issue #3151's Solution item 2 mentions "modify the unpublish date later", but currently once a schedule exists the dialog shows only the info-alert + "Cancel Scheduled Unpublish" (no datetime picker), so changing a date is a two-step cancel + re-schedule rather than an edit.
- **Publish-time prompt not included.** Issue #3151's Solution item 1 ("prompt to set an unpublish date *during* content publishing") isn't part of this change — unpublish scheduling is a separate post-publish menu action and `ConfirmPublishModal` is untouched. Flagging in case it's expected here vs. a follow-up.

### Suggested Cypress coverage
The new cases in `cypress/e2e/content/actions.spec.js` ("Schedules an item for unpublishing", "Cancels a scheduled unpublish") cover the indicator appearing/disappearing. Recommended additions in that spec:
- **Assert the API payload on schedule** — intercept `POST …/publishings` and assert the body is `{ publishAt: "now", unpublishAt: <UTC string>, version }`. Locks in the publish-vs-unpublish payload branching that is the heart of this change.
- **Assert the schedule dialog labeling** — after clicking `UnpublishScheduleButton`, assert the header reads "Schedule Unpublish:", the field label reads "Unpublish on", and the confirm button reads "Schedule Unpublish". Guards the publish/unpublish copy from regressing.
- **Assert the date is shown in the Unschedule dialog** (would catch the blank-date defect above) — open `UnschedulePublishButton` and assert the body text contains the scheduled date, not an empty string.
- **Assert the menu toggles** to "Unschedule Unpublish" once a schedule exists, and that the header indicator tooltip shows the unpublish date and publisher name.

src/apps/content-editor/src/app/views/ItemEdit/components/ItemEditHeader/ItemEditHeaderActions.tsx
geodem127: The `publishAt` field on `item.publishing` is part of the API's publishing record shape — it's the timestamp for when the item was published (or will be). It's not specific to unpublish scheduling. When `item.publishing.publishAt` is set, the item has an active/live publishing record. `item.scheduling` is a separate object representing a pending future schedule; `item.scheduling.isScheduled` is true only when the item hasn't gone live yet.

So the guard `!item?.publishing?.publishAt` is filtering to: "item has a scheduled future publish and is NOT currently live". Without it, a published item with a scheduled unpublish would also match `isScheduled` and incorrectly land in the `scheduled` state.

The naming (`publishAt`) comes from the API response shape — we don't control those field names.

src/shell/components/SchedulePublish/index.tsx
geodem127: Comment added in `handleUnschedulePublish` — see the latest commit.

agalin920:
> ## QA — Gaps & concerns + suggested Cypress coverage
> Verified the core **Schedule Unpublish** flow end-to-end against a live published item (open menu → Schedule Unpublish → set date → schedule → indicator shows → Unschedule → state restored). The `POST …/publishings` payloads are correct (`publishAt:"now"` + `unpublishAt:<UTC>` to schedule, `unpublishAt:"never"` to cancel). The feature works. A few things to address:
> 
> > Note: "Solution item 1/2" below refer to the two numbered bullets in issue #3151's **Solution** section — not to other tickets.
> 
> ### Gaps & concerns
> * **Unschedule-Unpublish dialog omits the scheduled date.** The body renders _"v241 is scheduled to unpublish on ⟨blank⟩ in (GMT-08:00) Pacific Time…"_. Root cause: `src/shell/components/SchedulePublish/index.tsx:155` derives `scheduledLocalText` only from `item.scheduling.publishAt`, but a scheduled _unpublish_ date lives in `item.publishing.unpublishAt`, so it resolves to `""`. Cosmetic (cancel still works, and the header indicator tooltip is wired to `unpublishAt`), but the confirmation should show the date being cancelled.
> * **"Invalid Time" label persists** on the time field even after a valid future time is selected (the UTC preview computes correctly and the confirm button stays enabled). Looks like a stale validation/aria label on `FieldTypeDateTime`'s time sub-field — worth a look as it can confuse users.
> * **No edit-in-place** for an existing scheduled unpublish date. Issue [Content: Schedule Unpublish Button #3151](https://github.com/zesty-io/manager-ui/issues/3151)'s Solution item 2 mentions "modify the unpublish date later", but currently once a schedule exists the dialog shows only the info-alert + "Cancel Scheduled Unpublish" (no datetime picker), so changing a date is a two-step cancel + re-schedule rather than an edit.
> * **Publish-time prompt not included.** Issue [Content: Schedule Unpublish Button #3151](https://github.com/zesty-io/manager-ui/issues/3151)'s Solution item 1 ("prompt to set an unpublish date _during_ content publishing") isn't part of this change — unpublish scheduling is a separate post-publish menu action and `ConfirmPublishModal` is untouched. Flagging in case it's expected here vs. a follow-up.
> 
> ### Suggested Cypress coverage
> The new cases in `cypress/e2e/content/actions.spec.js` ("Schedules an item for unpublishing", "Cancels a scheduled unpublish") cover the indicator appearing/disappearing. Recommended additions in that spec:
> 
> * **Assert the API payload on schedule** — intercept `POST …/publishings` and assert the body is `{ publishAt: "now", unpublishAt: <UTC string>, version }`. Locks in the publish-vs-unpublish payload branching that is the heart of this change.
> * **Assert the schedule dialog labeling** — after clicking `UnpublishScheduleButton`, assert the header reads "Schedule Unpublish:", the field label reads "Unpublish on", and the confirm button reads "Schedule Unpublish". Guards the publish/unpublish copy from regressing.
> * **Assert the date is shown in the Unschedule dialog** (would catch the blank-date defect above) — open `UnschedulePublishButton` and assert the body text contains the scheduled date, not an empty string.
> * **Assert the menu toggles** to "Unschedule Unpublish" once a schedule exists, and that the header indicator tooltip shows the unpublish date and publisher name.

@geodem127  please state the changes that have been introduced due to this review

geodem127:
@agalin920 — addressed in 64058eed3. Here's what was changed:

**Gaps & concerns:**

- ✅ **Blank date in Unschedule Unpublish dialog** — already fixed in a prior commit via \`dateText = isForUnpublish ? item.publishing.unpublishAt : item.scheduling.publishAt\`, so \`scheduledLocalText\` correctly shows the unpublish date.

- ✅ **"Invalid Time" label**  
- ℹ️ **No edit-in-place for existing scheduled unpublish** — will need to reach out to product since design for this scenario has not been included.
- ℹ️ **Publish-time unpublish prompt** — will need to reach out to product since design for this scenario has not been included.

**Cypress coverage additions:**

- ✅ **API payload assertion** — \`cy.wait(publishItem).then()\` asserts \`publishAt === "now"\`, \`unpublishAt\` matches UTC datetime format, and \`version\` is a number.
- ✅ **Dialog labeling assertion** — within \`SchedulePublishModal\`, asserts "Schedule Unpublish:" header, "Unpublish on" field label, and button text "Schedule Unpublish".
- ✅ **Scheduled date in Unschedule dialog** — asserts "scheduled to unpublish on" text and a matching date string (\`/\w{3} \d{1,2}, \d{4} at \d{1,2}:\d{2}/\`) are present.
- ✅ **Menu toggle label** — after scheduling, opens the menu and asserts \`UnpublishScheduleButton\` contains "Unschedule Unpublish".

Both new tests are now fully self-contained (\`cy.visit\` + \`cy.wait\`); the "Cancels" test schedules an unpublish before cancelling it.

agalin920:
> noted as a gap vs issue Content: Schedule Unpublish Button #3151 Solution item 2. Not addressed in t

What does "noted as a gap vs issue" mean? And how do we intend to address it?

geodem127:
> > noted as a gap vs issue Content: Schedule Unpublish Button #3151 Solution item 2. Not addressed in t
> 
> What does "noted as a gap vs issue" mean? And how do we intend to address it?

@agalin920, apologies for the confusion. I’ll follow up with the Product team, as this scenario hasn't been accounted for in the design.

src/shell/components/SchedulePublish/index.tsx
```diff
@@ -52,227 +42,229 @@ export const SchedulePublish = ({
...
+      );
+      onUnscheduleSuccess?.();
+    } catch {
+      // Error notification is handled by the thunk; swallow here so the
+      // finally block always closes the dialog cleanly.
+    } finally {
+      setIsLoading(false);
+      onClose();
+    }
+  };
```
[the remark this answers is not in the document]
geodem127: Done — replaced both `(dispatch as Function)` casts with `// @ts-expect-error untyped action` to match the rest of the file, and added a comment in the `hasAnyScheduledPublish` branch explaining the partial-state scenario and why a rollback isn't feasible.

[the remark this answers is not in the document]
geodem127: Small correction to my previous reply: `@ts-expect-error` can't be used for the `await dispatch(...)` form — the idiom works in `handleSchedulePublish` / `handleUnschedulePublish` because calling `.then()` on the `AnyAction` return type *is* a hard TS error that gets suppressed. With `await`, TypeScript emits only a ts(80007) suggestion ("await has no effect on the type"), not an error, so `@ts-expect-error` itself becomes an error ("unused directive").

Current state: both `(dispatch as Function)` casts are removed; the calls are plain `await dispatch(...)`. They work correctly at runtime — thunks return Promises — and `tsc --noEmit` is clean. The ts(80007) hints remain but are suggestions only.

src/apps/content-editor/src/app/views/ItemEdit/components/ItemEditHeader/ItemEditHeaderActions.tsx
```diff
@@ -311,15 +336,32 @@ export const ItemEditHeaderActions = ({
...
-    } else if (activePublishing?.version === item?.meta.version) {
+    } else if (
+      item?.publishing?.isPublished &&
+      item?.publishing?.version === item?.meta.version
+    ) {
       return ITEM_STATES.published;
     } else {
       return ITEM_STATES.draft;
     }
   })();
```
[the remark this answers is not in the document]
geodem127: can only schedule one at a time

src/shell/components/SchedulePublish/index.tsx
```diff
@@ -52,227 +41,222 @@ export const SchedulePublish = ({
...
 
-  const handleSchedulePublish = () => {
-    setIsLoading(true);
+  const latestChangeCreator = users?.find(
+    (user) => user.ZUID === item?.web?.createdByUserZUID
+  );
+  const creatorName = `${latestChangeCreator?.firstName ?? ""} ${
+    latestChangeCreator?.lastName ?? ""
+  }`.trim();
+  const savedAgo = item?.web?.updatedAt
```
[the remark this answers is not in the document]
geodem127: this is intentional

cypress/e2e/content/actions.spec.js
```diff
@@ -238,6 +238,170 @@ describe("Actions in content editor", () => {
...
+
+    cy.getBySelector("ScheduleUnpublishModal")
+      .should("exist")
+      .within(() => {
+        cy.getBySelector("ScheduleUnpublishButton").should("exist");
+        cy.getBySelector("ScheduleUnpublishButton").trigger("click");
+      });
+
+    cy.wait(publishItem);
+    cy.wait(publishings);
```
finnar-bin, line 383: why not just use cy.wait([...])?  Some waits below are already doing this.

src/apps/content-editor/src/app/views/ItemEdit/components/ItemEditHeader/ItemEditHeaderActions.tsx
```diff
@@ -156,6 +157,21 @@ export const ItemEditHeaderActions = ({
   const activePublishing = itemPublishings?.find(
     (itemPublishing) => itemPublishing._active
   );
+
+  const lastItemUpdateAudit = itemAudit?.find(
```
finnar-bin, line 162: Isn't this going to return the first match instead of the "last" item like what the variable name suggests?

src/apps/content-editor/src/app/views/ItemEdit/components/ItemEditHeader/ItemEditHeaderActions.tsx
```diff
@@ -156,6 +157,21 @@ export const ItemEditHeaderActions = ({
...
+    (audit: Audit) => audit.action === 2 || audit.action === 1
+  );
+
+  const hasScheduledPublish =
+    !!item?.scheduling?.isScheduled &&
+    new Date(item?.scheduling?.publishAt).getTime() > Date.now();
```
finnar-bin, line 168: Isn't the date comparison here a bit unnecessary? Shouldn't .isScheduled be enough?

src/apps/content-editor/src/app/views/ItemEdit/components/ItemEditHeader/ItemEditHeaderActions.tsx
```diff
@@ -311,7 +327,10 @@ export const ItemEditHeaderActions = ({
   const itemState = (() => {
     if (item?.dirty) {
       return ITEM_STATES.dirty;
-    } else if (item?.scheduling?.isScheduled) {
+    } else if (
+      item?.scheduling?.isScheduled &&
+      item?.scheduling?.version === item?.meta.version
```
finnar-bin: Could've just used the variable created above (`hasScheduledUnpublish`) since it's basically the same logic check

src/apps/content-editor/src/app/views/ItemEdit/components/ItemEditHeader/TooltipTitle.tsx
```diff
@@ -0,0 +1,37 @@
...
+}) => {
+  const formatted = dataTime ? formatDate(dataTime) : "";
+  const showOn =
+    formatted &&
+    !formatted.includes("Today") &&
+    !formatted.includes("Yesterday");
+  const user = users?.find((u) => u.ZUID === userZUID);
+  const userName = user
+    ? `${user.firstName || ""} ${user.lastName || ""}`.trim()
+    : "";
```
finnar-bin: Wouldn't it be better to have the parent components do all these calculations so this component is just a dumb component that just renders the provided data?

src/apps/content-editor/src/app/views/ItemEdit/components/ItemEditHeader/ItemEditHeaderActions.tsx
```diff
@@ -496,19 +518,16 @@ export const ItemEditHeaderActions = ({
...
-                  (user) => user.ZUID === item?.meta?.createdByUserZUID
-                )?.lastName}
-            </div>
+            <TooltipTitle
+              text={`v${item?.meta?.version} saved`}
+              dataTime={item?.meta?.updatedAt || ""}
+              userZUID={
+                lastItemUpdateAudit?.actionByUserZUID ||
+                item?.meta?.createdByUserZUID ||
+                ""
```
finnar-bin: Is there really any scenario that the first 2 won't have a userZUID that would would warrant the need for the empty string fallback?

src/apps/content-editor/src/app/views/ItemEdit/components/ItemEditHeader/ItemEditHeaderActions.tsx
```diff
@@ -545,35 +564,20 @@ export const ItemEditHeaderActions = ({
...
-                    (user: any) =>
-                      user.ZUID === activePublishing?.publishedByUserZUID
-                  )?.lastName
+              <TooltipTitle
+                text={`v${activePublishing?.version} published`}
+                dataTime={activePublishing?.publishAt || ""}
+                userZUID={
+                  activePublishing?.publishedByUserZUID ||
+                  item?.meta?.createdByUserZUID ||
+                  ""
```
finnar-bin: Same question here

src/apps/content-editor/src/app/views/ItemEdit/components/ItemEditHeader/PublishStatus.tsx
```diff
@@ -31,13 +30,12 @@ export const PublishStatus = ({ currentVersion }: PublishStatusProps) => {
...
-
-    if (user) {
-      return `${user.firstName} ${user.lastName}`;
-    }
-  };
+  const scheduledUnpublishing = itemPublishings?.find(
+    (item) =>
+      item._active &&
+      item?.unpublishAt &&
+      new Date(item?.unpublishAt).getTime() > Date.now()
```
finnar-bin: Kind of weird that you added conditional chaining on the last 2 but not on the first check

src/shell/store/content.js
```diff
@@ -876,16 +876,21 @@ export function publish(modelZUID, itemZUID, data, meta = {}) {
...
+          message = `Cancelled scheduled unpublish for ${title}`;
+        }
 
-        return dispatch(
-          notify({
-            message,
-            kind: "save",
-          })
-        );
+        return dispatch(notify({ message, kind: "save" }));
```
finnar-bin: `kind: save` isn't a thing

src/shell/store/content.js
```diff
@@ -956,12 +967,8 @@ export function unpublish(modelZUID, itemZUID, publishZUID, options = {}) {
...
           ? `Error Unscheduling version ${options.version}`
           : `Error Unpublishing ${title}`;
-        return dispatch(
-          notify({
-            message,
-            kind: "error",
-          })
-        );
+        dispatch(notify({ message, kind: "error" }));
+        throw err;
```
finnar-bin: Why is the error being thrown?

finnar-bin (dismissed):

src/apps/content-editor/src/app/views/ItemEdit/components/ItemEditHeader/TooltipTitle.tsx
geodem127: Multiple components use this. this enforces uniform formatting across all implementations.

src/apps/content-editor/src/app/views/ItemEdit/components/ItemEditHeader/ItemEditHeaderActions.tsx
geodem127: Audits are listed newest first.

src/apps/content-editor/src/app/views/ItemEdit/components/ItemEditHeader/ItemEditHeaderActions.tsx
geodem127: backup check for audits flagged as 'Scheduled' that have past dates.

src/apps/content-editor/src/app/views/ItemEdit/components/ItemEditHeader/ItemEditHeaderActions.tsx
geodem127: I've seen a few instances where the user field was missing and it displayed 'undefined' in the UI - likely due to corrupt data. Wouldn't it be safer to just add the || "" fallback now rather than waiting for the issue to come up again? And does this really have any significant downside?

src/apps/content-editor/src/app/views/ItemEdit/components/ItemEditHeader/ItemEditHeaderActions.tsx
geodem127: I've seen a few instances where the user field was missing and it displayed 'undefined' in the UI - likely due to corrupt data. Wouldn't it be safer to just add the || "" fallback now rather than waiting for the issue to come up again? And does this really have any significant downside?

src/shell/store/content.js
geodem127: I see that this is currently still in use. Should I assume it's being deprecated?

https://github.com/zesty-io/manager-ui/blob/3b39859dd68868fb913d0e4c97adf5193a0e017b/src/shell/store/content.js#L832-L837

https://github.com/zesty-io/manager-ui/blob/3b39859dd68868fb913d0e4c97adf5193a0e017b/src/shell/store/content.js#L893

https://github.com/zesty-io/manager-ui/blob/3b39859dd68868fb913d0e4c97adf5193a0e017b/src/apps/content-editor/src/app/components/ReorderNav/ReorderNav.js#L57-L62

https://github.com/zesty-io/manager-ui/blob/3b39859dd68868fb913d0e4c97adf5193a0e017b/src/apps/code-editor/src/store/headers.js#L63-L69

as well as in these files:
/src/apps/content-editor/src/app/components/ReorderNav/ReorderNav.js
/src/apps/code-editor/src/store/headers.js
/src/apps/content-editor/src/app/views/ItemEdit/ItemEdit.js
/src/apps/content-editor/src/app/views/ItemEdit/Content/Actions/Widgets/WidgetPurgeItem/WidgetPurgeItem.js
/src/apps/content-editor/src/app/views/ItemEdit/Content/Actions/Widgets/WorkflowRequest/WorkflowRequest.js
/src/apps/content-editor/src/app/views/LinkCreate/LinkCreate.js
/src/apps/content-editor/src/app/views/LinkEdit/LinkEdit.js

src/shell/store/content.js
geodem127: removed

cypress/e2e/content/actions.spec.js
geodem127: updated

src/apps/content-editor/src/app/views/ItemEdit/components/ItemEditHeader/ItemEditHeaderActions.tsx
geodem127: 'hasScheduledUnpublish' is using the 'item?.publishing?.version'

src/apps/content-editor/src/app/views/ItemEdit/components/ItemEditHeader/PublishStatus.tsx
geodem127: updated

src/apps/content-editor/src/app/views/ItemEdit/components/ItemEditHeader/TooltipTitle.tsx
finnar-bin: Not sure what you mean by "multiple components use this" as I'm seeing you're rendering these variables directly below, not passing the variables to child components.

src/apps/content-editor/src/app/views/ItemEdit/components/ItemEditHeader/ItemEditHeaderActions.tsx
finnar-bin: Probably worth checking in with backend as to why these user ZUIDs are undefined. Might be an issue on their end, and if that's the case it's probably worth fixing.

src/shell/store/content.js
finnar-bin: the notify type doesn't mention "save" as a valid value.

```
type NotifyArgs = {
  kind: "warn" | "error" | "success";
  HTML?: unknown;
  heading?: string;
  message: string;
};
```

geodem127:
Please note: the platform is designed to permit scheduling for a single item only.

src/shell/store/content.js
geodem127: updated

src/apps/content-editor/src/app/views/ItemEdit/components/ItemEditHeader/ItemEditHeaderActions.tsx
geodem127: removed "" fallback

src/apps/content-editor/src/app/views/ItemEdit/components/ItemEditHeader/ItemEditHeaderActions.tsx
geodem127: removed "" fallback

src/apps/content-editor/src/app/views/ItemEdit/components/ItemEditHeader/TooltipTitle.tsx
geodem127: moved the calculations to the parent component

src/apps/content-editor/src/app/views/ItemEdit/components/ItemEditHeader/PublishStatus.tsx
```diff
@@ -100,6 +110,43 @@ export const PublishStatus = ({ currentVersion }: PublishStatusProps) => {
...
+        scheduledUnpublishing.version === currentVersion && (
+          <Tooltip
+            enterDelay={1000}
+            enterNextDelay={1000}
+            title={
+              <TooltipTitle
+                text={`v${scheduledUnpublishing.version} scheduled to unpublish`}
+                dateTime={scheduledUnpublishing.unpublishAt || ""}
+                userName={getUserNameByZUID(
+                  activePublishing?.publishedByUserZUID
```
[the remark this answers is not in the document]
geodem127: had an issue when ipushed mr PR

agalin920 (dismissed):

src/shell/components/SchedulePublish/index.tsx
```diff
@@ -52,227 +42,239 @@ export const SchedulePublish = ({
...
 
-  const tzLabel =
-    TIMEZONES.find((tz) => tz.id === guessedTz)?.label || guessedTz;
+  const handleScheduleUnpublish = () => {
+    setIsLoading(true);
+    const { publishAtUtcStr, localPretty } = formatPayloadTimes(
+      selectedUtc,
+      publishTimezone
+    );
+    dispatch(
```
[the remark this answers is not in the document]
geodem127: schedule unpublish is only available for published items. items scheduled for publishing will not show this option

agalin920:
@geodem127 code blocker and tests failing

src/shell/components/SchedulePublish/index.tsx
```diff
@@ -52,227 +42,239 @@ export const SchedulePublish = ({
...
+      .then((response) => {
+        if (!response?.error) {
+          onScheduleSuccess?.();
+        }
+      })
+      .finally(() => {
+        setIsLoading(false);
+        onClose();
+      });
+  };
```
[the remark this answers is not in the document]
geodem127: Schedule unpublish will only be available if there are no other scheduled items and the active item has published status.

agalin920 (dismissed):

agalin920:
@geodem127 please make sure negative QA runs against this

geodem127:
> @geodem127 please make sure negative QA runs against this

Negative QA manually triggered for the latest commit.

src/shell/components/SchedulePublish/index.tsx
```diff
@@ -52,227 +42,243 @@ export const SchedulePublish = ({
...
```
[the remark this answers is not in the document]
geodem127: Good catch — fixed in 6f99adf10.

`handleScheduleUnpublish` now mirrors the `hasAnyScheduledPublish` pre-deletion guard already used by `handleUnscheduleUnpublish`: it deletes the conflicting scheduled publish first (throwing on failure), then POSTs the new `unpublishAt` schedule, all wrapped in try/catch/finally so the dialog closes cleanly either way. Removed the stale comment claiming this path was unreachable.

Verified: `tsc --noEmit` clean, and ran `content/actions.spec.js` twice with vs. without this change (isolating it via `git stash`) — 18/18 passing both ways, so no regression to the existing schedule-publish flows it sits next to.

geodem127:
Re: the negative-QA finding above ("Saving a new draft after scheduling an unpublish makes the schedule invisible and unmanageable") — fixed in 79df78daf.

Root cause confirmed: `hasScheduledUnpublish` (`ItemEditHeaderActions`), `isAlreadyScheduledUnpublish` (`SchedulePublish`), and the `PublishStatus` badge's render gate all keyed off `item.publishing.version === item.meta.version` (or `currentVersion`) — the currently loaded/edited version — instead of the published version's own state. Saving a new draft bumps `meta.version` past the published version, so all three flipped false even though the schedule was still armed server-side, exactly as the report showed via `item.publishing`.

Fix: all three now gate on `item.publishing.isPublished` instead, which reflects whether the *published* version itself is live and has a future `unpublishAt` — independent of what's currently loaded in the editor.

Verified live (not just unit-level):
- Ran a Playwright script against the dev instance: scheduled an unpublish → badge appears → edited the item and hit **Save** (not Publish, bumping `meta.version`) → badge **still visible**, survives a full page reload → "Unschedule Unpublish" still cancels it correctly.
- Also confirmed `VersionSelector` (which I'd initially missed — it renders for all content types, not just blocks) lets you switch back to the published version after saving a draft, which restores `itemState === "published"` and re-exposes the cancel menu, as intended by the original design.
- Ran `content/actions.spec.js` — both new specs (_Schedules an item for unpublishing_, _Cancels a scheduled unpublish_) pass.

Manually re-triggered this negative-QA workflow against the latest commit to confirm.

agalin920 (dismissed):

finnar-bin (dismissed):

src/shell/components/SchedulePublish/index.tsx
```diff
@@ -52,227 +42,267 @@ export const SchedulePublish = ({
...
-              </Typography>
-            </Box>
+      await dispatch(
+        publish(
+          item?.meta?.contentModelZUID,
+          item?.meta?.ZUID,
+          {
+            publishAt: "now",
+            unpublishAt: publishAtUtcStr,
+            version: item?.meta?.version,
```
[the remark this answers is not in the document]
geodem127: Just to clarify, the active item must be published for the “Schedule Unpublish” option to be visible. It won’t be available for items that are currently unpublished.

agalin920:
@geodem127  please run ci on this

src/apps/content-editor/src/app/views/ItemEdit/components/ItemEditHeader/ItemEditHeaderActions.tsx
```diff
@@ -832,16 +827,24 @@ export const ItemEditHeaderActions = ({
...
+          onUnpublishNow={() => {
+            setScheduledPublishDialogOpen(false);
+            setScheduledAction(null);
+            setUnpublishDialogOpen(true);
           }}
           onUnscheduleSuccess={() => {
             if (publishAfterUnschedule) {
               setIsConfirmPublishModalOpen(true);
             }
           }}
```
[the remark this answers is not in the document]
geodem127: Fixed — gated the `onUnscheduleSuccess` check on `scheduledAction === "publish"` in addition to `publishAfterUnschedule`, so a stale flag from an abandoned publish flow can no longer trigger the publish-confirm modal during an unpublish-unschedule flow.

agalin920 (approved):

PR #4039, opened 2026-04-16
[Workflows] - Default workflow permissions do not have owner permissions by default

geodem127:
- [x] Add owner as the default role value when creating a new status label

https://github.com/user-attachments/assets/5d04d444-4ee5-4723-9a9f-c1f2551eb111


agalin920:
@geodem127 The ticket seems to point to "Default workflow statuses" This seems to be addressing "when creating a new status label". Am i missing something?

geodem127:
> @geodem127 The ticket seems to point to "Default workflow statuses" This seems to be addressing "when creating a new status label". Am i missing something?

My bad. Got confused with the headers. It is referring to the default status labels.

geodem127:
NOTE: This would require backend changes because no default role is assigned to the default status labels.

finnar-bin:
@geodem127 I suggest marking this as draft if this requires a backend change

geodem127:
requires backend changes. no ui changes required

PR #4045, opened 2026-04-24
Content: Show referenced blocks in code and freestyle

### LINKED ISSUE #4005: Content / Blocks - Add the ability see all blocks and their variants used in a page
When Blue Cross Blue Shield Global manages their site in the content app, they want to see Blocks linked to their content, so they can quickly make adjustments without going to Freestyle.

## The Issue
Blue Cross Blue Shield’s on demand developers have been building their site using Freestyle. Due to this, content editors have difficulty viewing which Blocks are being used when working in the Content App.

## The Solution
We want to scan for Parsley references to Blocks in code and freestyle to show on content items. This will enable BCBS to view the Blocks used on their pages, and help with Block visibility when we get further with Studio-Mode.

**Note:** 
- We will want to identify Blocks linked by code and Freestyle.
- Exclude any Blocks which use the Block Selector Field

## Design
https://www.figma.com/design/oiesa7e6Q5mcGEAyOjYCtJ/Blocks-App?node-id=1250-50551&t=O4zAW40GpDpgtRIr-1

--

Original Feature Request

## Problem

When editing a page in freestyle:

1. You have to go through the freestyle page and wait for it to load. 

2. Manually look for the block and variant you have to edit 

3. Click it from the editor 

4. Once the said block is focused, you have to then click edit content from the sidebar to get to the block variant you want to edit. 

The customer noted that this process has too many clicks and takes too long. 

## Solution

Provide visibility into which blocks (and their variants) are currently used on a page directly within the Content Page view.

This would allow users to:

1. Quickly see all blocks used on the page.

2. Identify the exact variant associated with each block.

3. Navigate directly to edit the selected block/variant without going through the Freestyle canvas first.

This reduces unnecessary navigation and improves editing efficiency.

Image

finnar-bin:
Show all referenced blocks in the code template and freestyle when viewing a content item.
Resolves #4005 

image

### DIFF
--- a/src/apps/content-editor/src/app/components/UsedBlocks/extractBlockReferences.ts
+++ b/src/apps/content-editor/src/app/components/UsedBlocks/extractBlockReferences.ts
@@ -1,0 +1,147 @@
+/**
+ * extractBlockReferences.ts
+ *
+ * Parses a Zesty code file (e.g. something.html) and returns all block
+ * references found in it, covering every known syntax pattern.
+ *
+ * Supported patterns:
+ *  1. Block selector field:       {{block({this.field_name})}}
+ *  2. Basic path:                 {{block('/-/block/name.html')}}
+ *  3. Path with version:          {{block('/-/block/name.html?version=5')}}
+ *  4. Path with variant:          {{block('/-/block/name.html?variant=7-xxx')}}
+ *  5. Positional params:          {{block('/-/block/name.html', 'variantZUID')}}
+ *                                 {{block('/-/block/name.html', 'variantZUID', '4')}}
+ *  6. Full preview URL:           https://instance.preview.zesty.io/-/block/name.html
+ *                                 https://instance.preview.zesty.io/-/block/name.html?variant=7-xxx
+ */
+
+export interface BlockReference {
+  raw: string;
+  field: string | null;
+  blockName: string | null;
+  variant: string | null;
+  version: string | null;
+}
+
+/**
+ * Parses comma-separated Parsley block() arguments, respecting single and double quotes.
+ * e.g. "'/-/block/foo.html', '7-abc123', '2'" => ['/-/block/foo.html', '7-abc123', '2']
+ */
+function parseArgs(str: string): string[] {
+  const args: string[] = [];
+  let current = "";
+  let inQuote = false;
+  let quoteChar = "";
+
+  for (let i = 0; i < str.length; i++) {
+    const ch = str[i];
+    if (!inQuote && (ch === '"' || ch === "'")) {
+      inQuote = true;
+      quoteChar = ch;
+    } else if (inQuote && ch === quoteChar) {
+      inQuote = false;
+      args.push(current.trim());
+      current = "";
+    } else if (!inQuote && ch === ",") {
+      if (current.trim()) args.push(current.trim());
+      current = "";
+    } else if (inQuote) {
+      current += ch;
+    } else if (ch !== " ") {
+      current += ch;
+    }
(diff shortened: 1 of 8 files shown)


src/apps/content-editor/src/app/components/UsedBlocks/index.tsx
```diff
@@ -0,0 +1,270 @@
...
+      const uniqueBlockReferences = allBlockReferences.filter((ref) => {
+        const key = `${ref.blockName}|${ref.variant}|${ref.version}`;
+        if (seen.has(key)) return false;
+
+        seen.add(key);
+
+        return true;
+      });
+
+      if (!uniqueBlockReferences.length) return;
```
agalin920: This causes IsBuildingReferences to remain true? is that desired?

src/apps/content-editor/src/app/components/UsedBlocks/extractBlockReferences.ts
```diff
@@ -0,0 +1,142 @@
...
+    // {{block(this.selector)}} and {{block({this.some_field})}}
+    if (argsStr.startsWith("this.") || argsStr.startsWith("{this.")) continue;
+
+    const args = parseArgs(argsStr);
+    const fullPath = args[0] ?? "";
+
+    // Full URL passed as argument —
+    // {{block('https://instance/-/block/name.html')}} or {{block(https://...)}}
+    if (fullPath.startsWith("http://") || fullPath.startsWith("https://")) {
+      const url = new URL(fullPath);
```
agalin920: this could throw if the path is invalid or broken. consider wrapping it in a try/catch to skip it

src/apps/content-editor/src/app/components/UsedBlocks/index.tsx
```diff
@@ -0,0 +1,270 @@
...
+            (block) => block.name === ref.blockName.replace(".html", "")
+          );
+
+          if (blockModel) {
+            blockZUIDs.add(blockModel.ZUID);
+          }
+        }
+      });
+
+      // Fetch all items for each block model in parallel, preserving the (blockId → items)
```
agalin920, line 136: This could be very heavy, when working with content its best practice to use the redux store patterns to be able to resolve quickly and efficiently. Any reason we cant use it here?

agalin920 (changes requested):

src/apps/content-editor/src/app/components/UsedBlocks/index.tsx
finnar-bin: Right, updated the code

src/apps/content-editor/src/app/components/UsedBlocks/extractBlockReferences.ts
finnar-bin: Added handling in case url is malformed

src/apps/content-editor/src/app/components/UsedBlocks/index.tsx
finnar-bin: I was leaning more towards making sure that the data is always present hence opted for the api calls. But yeah, I do understand that this could end up with a lot of api calls if a content item is referencing a lot of blocks. I'll check out the redux store route you suggested.

finnar-bin: @agalin920 Updated to use redux store to get the variant data. If it doesn't exist, it attempts to fetch it to see if it exists on the database.

agalin920 (approved):

PR #4048, opened 2026-04-30
Amplitude: Enable session replay plugin

agalin920:
## Summary
- Add `@amplitude/plugin-session-replay-browser` as a dependency
- Register the plugin in `src/shell/views/Shell/Shell.tsx` (after `user.email` is loaded), gated on:
  - `window.CONFIG.ENV === "production"`
  - `!isZestyEmail(user.email)` — excludes internal users
- Extend `isZestyEmail` so it also matches `@content.one` (in addition to `@zesty.io`)
- Plugin is added at most once per page load via a module-level guard

## Why gate in code
Sampling could in theory be controlled via Amplitude's Targeted Sampling dashboard rules, but a dashboard misconfiguration would fail open and capture sessions we don't want (lower envs, internal staff). Gating in code fails closed and keeps the policy version-controlled alongside the rest of the app.

## Test plan
- [ ] Verify the bundle builds without errors (`npm run build:dev`)
- [ ] In production, log in as an external user → replay appears in Amplitude
- [ ] In production, log in as an `@zesty.io` or `@content.one` user → no replay captured
- [ ] Load the app in stage / development → no replay captured
- [ ] Confirm `isZestyEmail` widening doesn't regress the existing import in `GlobalTopbar.tsx`


src/shell/views/Shell/Shell.tsx
```diff
@@ -42,7 +42,18 @@ import { registerNavigate } from "../../../engine/navigator";
 import { AIDrawer } from "./AIDrawer";
 import { useLocalStorage } from "react-use";
 import * as amplitude from "@amplitude/analytics-browser";
+import { sessionReplayPlugin } from "@amplitude/plugin-session-replay-browser";
 import instanceZUID from "../../../utility/instanceZUID";
+import { isZestyEmail } from "../../../utility/isZestyEmail";
+
+let sessionReplayAdded = false;
+function maybeEnableSessionReplay(email?: string) {
```
shrunyan, line 50: Legibility request; Can we get this cleaned up into a single if check which when passing sets replay

shrunyan (dismissed):

finnar-bin (dismissed):

finnar-bin (approved):

PR #4053, opened 2026-05-06
Cypress: Code App - Editor

geodem127:
- [x] shows correct file contents
- [x] show version in file name
- [x] displays save/publish status
- [x] status changes when editing contents
- [x] show suggestions
- [x] show confirmation dialog when exiting without saving changes.
- [x] Save
- [x] Publish
- [x] Show Diff
- [x] Delete File

---

Closing out the remaining 7 checklist items (no GitHub issue backs this PR — planned directly from the checklist above). All 7 behaviors already existed in the app; this is purely new Cypress coverage in `cypress/e2e/code/editor.spec.js`, plus small additive `data-cy` attributes on the Diff Versions button, More Options menu/Delete File item, Differ version selects/Load/Save Version buttons, and the Delete dialog (no behavior changes).

Known follow-up, not a blocker: ad hoc manual UI verification against the live dev instance hit a one-off case where the Publish button didn't visually flip to "Published" immediately after a successful publish (toast + 200 response confirmed), while the underlying Cypress "Publish" test itself passes reliably. Looks like a pre-existing eventual-consistency gap between the publish/purge-cache call and the live-status refetch, likely surfaced by repeated automated hits against the shared dev instance rather than a regression from this change.


cypress/fixtures/code/editor-suggestions.json
finnar-bin, line 1: Is there any reason why this is separated from the other fixture file that this PR also introduced?

cypress/plugins/seeds/code.ts
```diff
@@ -61,6 +64,16 @@ module.exports = function code(config) {
 
   // CODE TASK MAPPING
   return {
-    "seed:code": (path: string) => seedCode(path),
+    // Accepts either a plain fixture path (existing call sites) or
+    // `{ path, overrides }` when a caller needs to merge in values that can
+    // only be known at test run time (e.g. a freshly-created contentModelZUID).
+    "seed:code": (
+      args: string | { path: string; overrides?: Record<string, any> }
```
finnar-bin: Why not just make the expected arguments always an object with `overrides` being optional instead of accepting 2 different argument types? Kind of adds a mental overhead to always remember this function accepts 2 different argument types

src/apps/code-editor/src/app/components/TopBar/DifferActions.tsx
```diff
@@ -35,6 +35,17 @@ interface FileVersion {
...
+// MUI's `SelectDisplayProps` type (`React.HTMLAttributes<HTMLDivElement>`) has
+// no `data-*` index signature, so a `data-cy` object literal fails the excess
+// property check when assigned inline. Declaring it as a typed constant first
+// avoids that (structural typing still allows the extra property).
+const codeOneSelectDisplayProps = {
+  "data-cy": "code-app-differ-version-select-one",
+} as HTMLAttributes<HTMLDivElement>;
+const codeTwoSelectDisplayProps = {
+  "data-cy": "code-app-differ-version-select-two",
+} as HTMLAttributes<HTMLDivElement>;
```
finnar-bin: I would honestly prefer just directly passing this to the component instead of arbitrarily storing these in a variable

finnar-bin (changes requested):

cypress/fixtures/code/editor-suggestions.json
geodem127: Good catch — there wasn't actually a reason. Removed `editor-suggestions.json` and pointed the suggestions test at the existing `editor.json` fixture; the test only needs *some* file to bind a `contentModelZUID` to, it doesn't depend on the fixture's content or view `type`. (c51bde192)

cypress/plugins/seeds/code.ts
geodem127: Agreed, removed the dual-argument-type API entirely — `seed:code` now only takes a plain fixture path string again. The one call site that needed a runtime-only value (`contentModelZUID` from a freshly-created model) now seeds the file plainly, then binds the model via a follow-up `cy.apiRequest` PUT to `/web/views/{ZUID}`, matching the pattern already used elsewhere in the suite. (c51bde192)

src/apps/code-editor/src/app/components/TopBar/DifferActions.tsx
geodem127: Done — removed the module-level `codeOneSelectDisplayProps`/`codeTwoSelectDisplayProps` constants and inlined the cast directly at each `SelectDisplayProps` call site. (c51bde192)

agalin920 (approved):

finnar-bin (approved):

PR #4078, opened 2026-05-12
Sentry: Ignore N+1 API calls from transactions

### LINKED ISSUE #4057: Exclude N+1 api calls from being flagged by sentry
Check out project settings to find a way to pattern match these non-issues to prevent them from muddying up the board

Image

finnar-bin:
## Summary

Certain endpoints in the app are called repeatedly as a natural consequence of how the UI renders — they are not genuine performance problems. Without filtering, these calls generate N+1 transaction noise in Sentry that obscures real issues.

This PR introduces a `beforeSendTransaction` hook that suppresses N+1 span noise for known-safe endpoints. Rather than dropping the entire transaction, only duplicate spans are removed — the first occurrence of each matched endpoint is preserved so the call remains visible in Sentry.

The ignored endpoint patterns are:

- `/v1/search/items`
- `/v1/instances/:id/comments`
- `/v1/content/models/:id/items`

These were confirmed as non-issues required for UI rendering. Key implementation details:

- Deduplication is keyed by the **matched regex pattern** (not the full URL), so endpoints with variable path segments (e.g. different model ZUIDs in `/v1/content/models/:id/items`) are correctly grouped rather than treated as distinct calls.
- Endpoint patterns match on **path only** (no hardcoded domain) so filtering works consistently across all environments.
- Uses `spanToJSON` from `@sentry/core` to read span properties, which is the correct API for accessing span data in this version of the SDK.

Resolves #4057


shrunyan:
`accounts.api.zesty.io/v1/instances/:id/comments` should include a code comment noting that this endpoint currently is part of the allow list but should get update to a batch.

finnar-bin:
@shrunyan added the code comment for the comments endpoint

src/utility/sentry.js
```diff
@@ -59,6 +68,21 @@ if (["stage", "production"].includes(__CONFIG__?.ENV)) {
...
+        const { op, description } = spanToJSON(span);
+        return (
+          op === "http.client" &&
+          N1_IGNORED_ENDPOINTS.some((pattern) =>
+            pattern.test(description ?? "")
+          )
+        );
+      });
+
+      if (isIgnoredN1) return null;
```
shrunyan: Should we be dropping the whole event or just removing matching spans? What if there is a legitimate error in the event along with a span that contains one of these requests, wouldn't that get swallowed with this code?

Is there no indication of it being N1 on the event in the front-end? Does that determination happen on the sentry side?

finnar-bin: I have updated the code so that the entire event isn't dropped. Instead we're just filtering out duplicate spans for the ignored endpoints. This way, the span still gets recorded on sentry while preventing the N+1 noise.

src/utility/sentry.js
```diff
@@ -1,6 +1,15 @@
 import * as Sentry from "@sentry/react";
+import { spanToJSON } from "@sentry/core";
```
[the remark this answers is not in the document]
finnar-bin: Don't think it exists in this version we're using

agalin920:
@finnar-bin do we really need to do this app level? Can this not be simply configured in project settings in sentry to not capture?

agalin920 (changes requested):
see comment

geodem127 (approved):

finnar-bin:
@agalin920 Looked into this a bit. Inbound Filters (Project Settings > Inbound Filters) only has a fixed set of server-side toggles: browser extension errors, localhost events, legacy browsers, web crawlers, plus error message/release/IP filters. There's no way to match on transaction or span URLs there, so it can't target these specific endpoints, that feature just isn't built for this.

There is a setting for N+1 API Calls detector sensitivity under Project Settings > Performance, but it's project wide. Turning it down would quiet N+1 detection across every endpoint, not just the three we've confirmed are fine, so we'd lose sensitivity for real N+1 regressions elsewhere too.

So app level filtering is really the only way to target just these known safe endpoints without dulling detection everywhere else.

agalin920 (approved):

PR #4100, opened 2026-05-14
Schema - Fix Add HTTP Header button losing rows on double-click

### LINKED ISSUE #4094: Schema - Integration field 'Add HTTP Header' button loses rows on double-click due to Date.now() key collision + stale closure
## Problem

The "Add HTTP Header" button in the integration field's Connect dialog has a race condition that **drops rows when the user double-clicks the button**. A normal impatient user double-tapping the button gets only ONE new row instead of two. The race fires whenever two clicks land within the same millisecond — which is reachable through real user interaction (double-click), keyboard repeat, or assistive tech.

## Source

`src/shell/components/FieldTypeIntegration/IntegrationFieldConfigure/ConnectToApi.tsx:336-346`

```tsx
<Button
  data-cy="addHeaderButton"
  fullWidth
  variant="outlined"
  startIcon={<AddCircleIcon />}
  onClick={() => {
    const keyId: string = Date.now().toString();   // ← (a) collides if two clicks happen in same ms
    focusRef.current = keyId;
    setHeadersLocal({
      ...headersLocal,                              // ← (b) closes over stale state, not functional setter
      [keyId]: {
        key: "",
        value: "",
      },
    });
  }}
>
  Add HTTP Header
</Button>
```

Two compounding bugs:

1. **Key collision** — `Date.now().toString()` returns the same value if both clicks land in the same millisecond. Modern browsers can dispatch double-click events <1ms apart. The second `setHeadersLocal` overwrites the first new row's key.

2. **Stale closure spread** — `setHeadersLocal({...headersLocal, [keyId]: ...})` reads `headersLocal` from the React closure captured at render time, not the latest value. Under React 18 batching, two synchronous click handlers running back-to-back both spread the same `headersLocal` snapshot, so the second update overwrites the first.

Either bug alone would cause the issue. Both together compound it.

## Reproduction

**Manual repro (real user path):**

1. Schema → Add Field → Integration → Connect to API
2. Double-click the "Add HTTP Header" button (just like a user might if they're impatient)
3. **Observed:** only 1 new row is added (count goes 1 → 2 instead of 1 → 3)

**Programmatic repro (more reliable):**

```js
// In devtools console while the Connect dialog is open
const btn = document.querySelector('[data-cy="addHeaderButton"]');
btn.click(); btn.click(); btn.click();
document.querySelector('[data-cy="integrationHeadersContainer"]').children.length;
// Expected: 4 (1 initial + 3 added)
// Actual:   2 (1 initial + only 1 added — three rapid clicks collapse to one)
```

## Fix

```tsx
import { v4 as uuidv4 } from "uuid"; // already imported at the top of this file

onClick={() => {
  const keyId = uuidv4();                         // ← unique per click regardless of timing
  focusRef.current = keyId;
  setHeadersLocal(prev => ({                      // ← functional updater, no stale closure
    ...prev,
    [keyId]: { key: "", value: "" },
  }));
}}
```

Both bugs fix with the same single change.

## Severity

**Minor.** User-reachable but non-data-destructive — they just need to click again to add more rows. No silent corruption or crash. Polish item.

## Test coverage gap

Cypress integration spec doesn't double-click the Add Header button. Adding a `cy.get('[data-cy="addHeaderButton"]').dblclick()` followed by an assertion that 2 new rows were added would catch this.

geodem127:
## Summary
The root cause was a **stale closure** on `headersLocal` inside the click handler: the handler captured a snapshot of headers at render time, so a second click in the same render cycle spread `{...staleHeaders, [uuid2]: ...}`, silently overwriting the row added by the first click.

Two complementary fixes:
- **Functional state updater** — uses `(prev) => ({...prev, [id]: ...})` so each click composes from the latest state rather than a stale closure capture. This is the primary fix.
- **`uuidv4()` for row keys** — replaces `Date.now().toString()` as defense-in-depth: two `Date.now()` calls in the same millisecond produce the same key, masking the stale-closure bug as a silent overwrite. Unique UUIDs surface any future key collision as a visible problem.

## Test plan
- [x] Open Schema app, navigate to an integration field, and open the API configuration
- [x] Double-click "Add HTTP Header" rapidly — confirm both rows appear instead of one being lost
- [x] Verify headers still save and persist correctly

### DIFF
--- a/src/shell/components/FieldTypeIntegration/Configure/ConnectToApi.tsx
+++ b/src/shell/components/FieldTypeIntegration/Configure/ConnectToApi.tsx
@@ -30,7 +30,6 @@ import { FieldWrapper } from "../Shared/FieldWrapper";
 import { IntegrationRequestHeaders } from "../../../services/types";
 import { validateUrl } from "../utils";
 import useIntegrationField from "../useIntegrationField";
-import { v4 as uuidv4 } from "uuid";
 
 const CONNECTION_STATUSES: {
   [key: string]: {
@@ -94,7 +93,7 @@ const ConnectToApi = ({
   setActiveStep: (step: number) => void;
   closeForm?: () => void;
 }) => {
-  const focusRef = useRef<string>("url");
+  const focusRef = useRef<number | "url">("url");
   const { data, status, fetchApiData } = useIntegrationField();
 
   const [isValidUrl, setIsValidUrl] = useState(true);
@@ -103,27 +102,23 @@ const ConnectToApi = ({
   const [endpointLocal, setEndpointLocal] = useState<string>(endpoint || "");
 
   const [headersLocal, setHeadersLocal] = useState<
-    Record<string, { key: string; value: string }>
+    { key: string; value: string }[]
   >(() => {
     if (!headers || Object.keys(headers).length === 0) {
-      const id = uuidv4();
-      return { [id]: { key: "", value: "" } };
+      return [{ key: "", value: "" }];
     }
-
-    return Object.entries(headers).reduce((acc, [key, value]) => {
-      const id = uuidv4();
-      acc[id] = { key, value: String(value) };
-      return acc;
-    }, {} as Record<string, { key: string; value: string }>);
+    return Object.entries(headers).map(([key, value]) => ({
+      key,
+      value: String(value),
+    }));
   });
 
   const handleNext = () => {
-    const headersLocalValues = Object.values(headersLocal);
-    const headersWithKeys = headersLocalValues.filter((i) => !!i?.key);
-    const reqHeaders = !headersWithKeys?.length
+    const headersWithKeys = headersLocal.filter((h) => !!h.key);
+    const reqHeaders = !headersWithKeys.length
       ? null
-      : headersWithKeys?.reduce((acc: any, obj: any) => {
-          acc[obj.key] = obj.value;
+      : headersWithKeys.reduce((acc: any, { key, value }) => {
+          acc[key] = value;
           return acc;
         }, {});
 
@@ -141,13 +136,12 @@ const ConnectToApi = ({
   const handleApiConnect = useCallback(() => {
     setReqAborted(false);
     setApiData(null);
-    const headersLocalValues = Object.values(headersLocal);
-    const headersWithKeys = headersLocalValues.filter((i) => !!i?.key);
-    const options = !headersWithKeys?.length
+    const headersWithKeys = headersLocal.filter((h) => !!h.key);
+    const options = !headersWithKeys.length
       ? {}
       : {
-          headers: headersWithKeys?.reduce((acc: any, obj: any) => {
-            acc[obj.key] = obj.value;
+          headers: headersWithKeys.reduce((acc: any, { key, value }) => {
+            acc[key] = value;
             return acc;
           }, {}),
         };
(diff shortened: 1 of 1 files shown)


finnar-bin (dismissed):

src/shell/components/FieldTypeIntegration/Configure/ConnectToApi.tsx
```diff
@@ -334,15 +334,12 @@ const ConnectToApi = ({
             variant="outlined"
             startIcon={<AddCircleIcon />}
             onClick={() => {
-              const keyId: string = Date.now().toString();
+              const keyId: string = uuidv4();
```
agalin920, line 330: what are these keys used for? if this is just local is there any reason we dont just use array position index?

agalin920 (changes requested):

src/shell/components/FieldTypeIntegration/Configure/ConnectToApi.tsx
geodem127: These keys were used before to uniquely identify each header row for targeted updates and deletions — a different approach using an object map. Since it's no longer necessary, I've replaced it with a plain array and index-based operations.

src/shell/components/FieldTypeIntegration/Configure/ConnectToApi.tsx
finnar-bin, line 1: This seems like a huge change compared to last time I reviewed this. How does this exactly resolve the issue again? The PR description doesn't seem to match the actual code changes.

geodem127: Hey @finnar-bin, fair point — the PR description is outdated and doesn't reflect what the code actually does now. I'll update it.

Here's what happened: as noted in [this comment](https://github.com/zesty-io/manager-ui/pull/4100#discussion_r3262252476), the original code stored headers as an object map keyed by UUID (`{ [uuid]: { key, value } }`) so that each row had a stable identity for targeted updates and deletions. The first fix kept that shape but swapped `Date.now().toString()` for `uuidv4()` to avoid same-millisecond key collisions on rapid double-clicks. When @agalin920 asked whether the UUID keys were actually necessary, I realized the object map itself was the unnecessary part — switching to a plain array made the whole collision problem moot and let index-based operations replace the keyed lookups.

src/shell/components/FieldTypeIntegration/Configure/ConnectToApi.tsx
```diff
@@ -244,106 +238,87 @@ const ConnectToApi = ({
...
           >
-            {Object.entries(headersLocal)?.map((entry, i) => {
-              const entryId: string = entry[0];
-              const headerKey = entry[1]?.key || "";
-              const headerValue = entry[1]?.value || "";
-              return (
+            {headersLocal.map((header, i) => (
+              <Grid
+                data-cy={`integrationHeadersContainerRow-${i}`}
+                key={i}
```
finnar-bin: Are users able to delete headers? Because if yes, using the array index as keys isn't really stable when it's constantly changing.

src/shell/components/FieldTypeIntegration/Configure/ConnectToApi.tsx
```diff
@@ -244,106 +238,87 @@ const ConnectToApi = ({
...
+              >
+                <Grid size={8}>
+                  <TextField
+                    autoFocus={i === focusRef.current}
+                    className="keyInput"
+                    fullWidth
+                    size="small"
+                    placeholder="Key"
+                    value={header.key}
+                    onChange={(e: any) => {
```
finnar-bin: `e` should be properly typed

src/shell/components/FieldTypeIntegration/Configure/ConnectToApi.tsx
```diff
@@ -244,106 +238,87 @@ const ConnectToApi = ({
...
+                    className="valueInput"
+                    size="small"
+                    placeholder="Value"
+                    value={header.value}
+                    onChange={(e: any) => {
+                      setHeadersLocal(
+                        headersLocal.map((h, idx) =>
+                          idx === i ? { ...h, value: e.target.value } : h
+                        )
+                      );
```
finnar-bin: This could potentially cause an issue with stale states. Prefer to use `prevState`.

finnar-bin (changes requested):

src/shell/components/FieldTypeIntegration/Configure/ConnectToApi.tsx
geodem127: Reverted. The array-index approach was a mistake — headers can be deleted, so index-based keys are unstable. The code is back to the `Record<uuid, {key, value}>` shape from the first commit: each row gets a UUID on creation and on load, `key={"header-${entryId}"}` is the stable React key, and `focusRef` tracks the UUID of the newly added row instead of its index.

Also fixed while here: `onChange` handlers now use the functional `setHeadersLocal((prev) => ...)` form (stale closure risk), and event types are `React.ChangeEvent<HTMLInputElement>` instead of `any`.

src/shell/components/FieldTypeIntegration/Configure/ConnectToApi.tsx
geodem127: Yes they can. Reverted to UUID keys — each row gets a stable `uuidv4()` id on init and on add, so deletions don't shift keys.

src/shell/components/FieldTypeIntegration/Configure/ConnectToApi.tsx
geodem127: Fixed — typed as `React.ChangeEvent<HTMLInputElement>`.

src/shell/components/FieldTypeIntegration/Configure/ConnectToApi.tsx
geodem127: Fixed — both `onChange` handlers now use `setHeadersLocal((prev) => ...)`.

geodem127:
Regarding the minor observation: always rendering at least one row and hiding the first row's delete button is intentional per design — the field should never present an empty state with no header rows visible.

geodem127:
PR description updated to lead with the stale closure as the primary root cause and frame `uuidv4` as defense-in-depth.

Nits applied:
- URL `TextField` `onInput/(e: any)` → `onChange/(e: ChangeEvent<HTMLInputElement>)`
- Both `reduce` accumulators in `handleNext`/`handleApiConnect` typed as `Record<string, string>`

Cypress: acknowledged as a pre-existing structural issue (nested `it()` calls inside `connectToEndpoint()` never register under Mocha). Fixing that requires restructuring the spec file — tracking as a separate follow-up.

geodem127:
PR #4153 fixes failing and flaky tests

agalin920 (approved):

finnar-bin (approved):

PR #4106, opened 2026-05-15
Schema - Fix integration field Resync missing falsy-to-falsy drift transitions

### LINKED ISSUE #4096: Content - Integration field Resync drift detection silently misses falsy↔non-falsy value transitions due to falsy-check fallback in keyPathValuesToString
## Problem

The integration field's "Resync" feature (the Refresh icon that appears next to a row in the selection dialog when the remote item has drifted from the locally-saved snapshot) uses a string-hash comparison that **collapses `null`/`undefined`/`0`/`""`/`false` into the same hash value**. As a result, drift detection silently fails for any change between two falsy values — most importantly, `null ↔ 0` and `0 ↔ undefined` transitions go undetected and the user is never prompted to refresh stale data.

## Source

`src/shell/components/FieldTypeIntegration/utils.ts:22-38`

```ts
export const keyPathValuesToString = (
  item: ApiDataProps,
  keyPaths: IntegrationKeyPaths
) => {
  const validValues = Object.values(keyPaths)
    ?.filter((value) => {
      if (Array.isArray(value)) return value?.length > 0;
      return value !== "";
    })
    ?.flat();
  const idParts = validValues?.map((key) => {
    const value = item?.[key] || "";    // ← line 33: 0/null/undefined/false/"" all become ""
    return typeof value === "string" ? value?.replace(/\s+/g, "") : value;
  });

  return idParts?.join(";");
};
```

Called from `IntegrationFieldSelect/ItemSelectionDialog.tsx:108-112`:

```tsx
const localItem =
  value?.find((val) => val?.[keyPaths?.itemId] === item?._itemId) || null;
const remoteItemData = keyPathValuesToString(item, keyPaths);
const localItemData = keyPathValuesToString(localItem, keyPaths);
const hasUpdates = !!localItem && remoteItemData !== localItemData;   // ← Resync icon condition
```

## How it manifests

| Local snapshot value | Remote API value | Hash comparison | User sees Refresh icon? | Reality |
|---|---|---|---|---|
| `5` | `5` | `5 === 5` | No | ✓ correct (no drift) |
| `5` | `7` | `5 !== 7` | **Yes** | ✓ correct (drift detected) |
| `5` | `0` | `5 !== ""` | **Yes** | ✓ correct (drift detected because local is non-falsy) |
| `0` | `5` | `"" !== 5` | **Yes** | ✓ correct |
| **`0`** | **`null`** | `"" === ""` | **No** | ❌ **wrong — drift not detected** |
| **`null`** | **`0`** | `"" === ""` | **No** | ❌ **wrong — drift not detected** |
| **`""`** | **`null`** | `"" === ""` | **No** | ❌ **wrong — drift not detected** |
| **`false`** | **`null`** | `"" === ""` | **No** | ❌ **wrong — drift not detected** |
| **`0`** | **`false`** | `"" === ""` | **No** | ❌ **wrong — drift not detected** |

So if a remote API used to return `weight: 0` for unknown players and now returns `weight: null`, the integration field saved snapshots will continue showing the stale `0` and the user will never be prompted to refresh.

## Reproduction

This is hard to repro through the UI alone because it requires server-side mutation of the API response between two reads. But the source-code logic is mechanically deterministic — see the table above.

If you want a live repro: edit `/api/qa-edge-falsy.json` (or any seeded endpoint) to return values that flip between `0` and `null` for a given player's saved keyPath, save an item with version A, swap the response to version B, reopen the selection dialog, and verify the Refresh icon does not appear when it should.

## Fix

Replace the falsy-fallback with explicit nullish handling and stringify all values consistently:

```ts
const idParts = validValues?.map((key) => {
  const value = item?.[key];
  if (value === null || value === undefined) return '<null>';   // distinguishable sentinel
  if (typeof value === 'string') return value.replace(/\s+/g, '');
  return JSON.stringify(value);   // covers 0, false, true, numbers, nested objects
});
```

This way:
- `0` → `"0"`
- `null` → `"<null>"`
- `false` → `"false"`
- `""` → `""`

All four serialize differently, so transitions between any two are detected.

## Severity

**Minor (semantic).** Edge case requiring specific value transitions in real-world APIs. Most APIs are consistent about whether they return `0` or `null` for a given field, so the mismatch is uncommon. But when it does happen, the user has no signal that their local data is stale — and no manual "refresh all" option to fall back on.

## Related

- Same falsy-bug family as #4093 (DisplayCard) and #4086 / #4083 (search crash). All three share the root cause: the integration field's data layer doesn't distinguish "missing" from "falsy".

## Test coverage gap

No Cypress test exercises Resync at all (the `cypress/e2e/schema/integration.spec.js` doesn't cover the Refresh-icon flow). Adding a test that mutates an API response and verifies the Refresh icon appears (with `cy.intercept`) would also have caught this if the test included a falsy↔falsy transition.

geodem127:
## Summary

Fixes #4096

- Replaces the `item?.[key] || ""` falsy-fallback in `keyPathValuesToString` with explicit nullish handling, so `null`, `undefined`, `0`, `false`, and `""` all produce distinct serialized values
- Also filters `null` key-path entries before mapping to avoid a TypeScript index-type error

## Test plan

- [x] Open an integration field selection dialog with a saved item whose remote API response has changed a field value from `0` → `null` (or any falsy↔falsy transition) and confirm the Resync (Refresh) icon now appears
- [x] Confirm no regression for non-falsy drift cases — Refresh icon still appears when e.g. `5` → `7`
- [x] Confirm no Refresh icon for unchanged items


src/shell/components/FieldTypeIntegration/utils.ts
```diff
@@ -26,12 +26,14 @@ export const keyPathValuesToString = (
   const validValues = Object.values(keyPaths)
     ?.filter((value) => {
       if (Array.isArray(value)) return value?.length > 0;
-      return value !== "";
+      return value !== "" && value !== null;
```
finnar-bin: Why only guard for `null`? I would assume we'd also want to look out if it happens to be `undefined` as well?

src/shell/components/FieldTypeIntegration/utils.ts
```diff
@@ -26,12 +26,14 @@ export const keyPathValuesToString = (
...
     })
     ?.flat();
   const idParts = validValues?.map((key) => {
-    const value = item?.[key] || "";
-    return typeof value === "string" ? value?.replace(/\s+/g, "") : value;
+    const value = item?.[key];
+    if (value === null || value === undefined) return "<null>";
```
finnar-bin, line 35: What is `<null>` used for?

finnar-bin (changes requested):

src/shell/components/FieldTypeIntegration/utils.ts
geodem127: Good catch —  wasn't guarded because the fields are typed as nullable (`string | null`) but the optional modifier (`?`) also allows `undefined`. Updated the filter to include `value !== undefined` so undefined key paths don't slip through.

src/shell/components/FieldTypeIntegration/utils.ts
geodem127: It's a sentinel string that preserves positional identity in the composite key used for drift detection. When a field value is `null` or `undefined`, omitting it would collapse the key and make two records with different field shapes look identical (e.g. `"foo;baz"` vs `"foo;bar"` become indistinguishable if the middle field is missing). Using `"<null>"` keeps the slot in place so the comparison in `hasUpdates` stays accurate.

finnar-bin (dismissed):

cypress/e2e/code/all-files-page.spec.js
```diff
@@ -62,11 +62,20 @@ describe("All Files Page", () => {
...
     cy.getBySelector("AllFilesTable")
-      .find('[data-cy="AllFilesRow"]:eq(0)')
-      .should("contain.text", TEST_DATA[0].filename, matchcase);
+      .find('[data-cy="AllFilesRow"] div div:eq(1)')
+      .then(($cells) => {
+        const times = $cells
+          .map((i, cell) => {
+            const text = cell.innerText.trim();
+            return parseInt(text.split(" ")[0]);
+          })
```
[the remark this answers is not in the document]
geodem127: will be addressed in this PR: https://github.com/zesty-io/manager-ui/pull/4153

geodem127:
PR #4153 fixes failing and flaky tests  including the ones mentioned

2. Time-unit comparison is unreliable (see inline comment at lines 72–74)

agalin920 (approved):

finnar-bin (approved):

geodem127:
The Integration Fields component has undergone refactoring. The utils file is being removed, with its functions being relocated to the component itself. These updates will be ported to the other PR.

Refactor PR: https://github.com/zesty-io/manager-ui/pull/4118

geodem127:

geodem127:
Closed: fix covered in PR: #4118

PR #4108, opened 2026-05-15
Content - Fix DisplayCard rendering placeholder for falsy values like 0

### LINKED ISSUE #4093: Content - Integration field DisplayCard renders placeholder text when value is the number 0 (or empty string / false) due to falsy-check fallback
## Problem

`DisplayCard` uses a JavaScript falsy-check (`||`) to substitute a placeholder when a value is missing. The check incorrectly fires for valid-but-falsy values: **`0` (number), `""` (empty string), and `false` (boolean)**.

The most visible instance: a player with `jerseyNo: 0` configured as the card's heading renders as **"Add Heading"** instead of "0", suggesting the field is empty when it actually contains a valid value.

This is the **same family of bug as the search-crash already fixed in #4083** (`?.trim()` on a non-string) — both stem from the integration field's display layer not properly handling the full range of JSON value types. Four sites in `DisplayCard.tsx` exhibit the same pattern.

## Source

`src/shell/components/FieldTypeIntegration/components/DisplayCard.tsx:67, 74, 82, 159`

```tsx
const headingValue = loading ? (
  <Skeleton width={type === "simple" ? "90%" : "55%"} height="26px" />
) : typeof heading === "boolean" ? (
  String(heading)
) : (
  heading || "Add Heading"          // ← line 67: 0/""/null all become placeholder
);

const subHeadingValue = loading ? (
  <Skeleton width="90%" />
) : typeof subHeading === "boolean" ? (
  String(subHeading)
) : (
  subHeading || "Add Subheading"    // ← line 74
);

const detailValue = loading ? (
  <Skeleton width="70px" />
) : typeof detail === "boolean" ? (
  String(detail)
) : (
  detail || "Add Detail"            // ← line 82
);

// inside details rendering, line 159:
{loading ? (
  <Skeleton sx={{ width: "95%" }} />
) : (
  item?.key || "+ Add Detail"        // ← line 159
)}
```

Note the inconsistency: each of these has a `typeof X === "boolean"` branch that uses `String(X)` (correct, will render `"true"`/`"false"`). But the author missed the equivalent `typeof X === "number"` branch — so `0` falls through to the `||` fallback and shows the placeholder.

The `keyPathValuesToString` helper at `utils.ts:33` also has the same pattern (see related issue for drift detection).

## Type system also lies

`DisplayCardProps` declares:

```tsx
export type DisplayCardProps = {
  type: IntegrationTypes;
  heading: string;            // ← claims string-only
  subHeading?: string;        // ← claims string-only
  thumbnail?: string;
  ...
};
```

But the function explicitly branches on `typeof heading === "boolean"` (line 64-65), proving the runtime types include booleans. Numbers also flow through. The TypeScript signature gives a false sense of safety.

## Reproduction

1. Schema → create integration field type=`text` → endpoint `https://<instance>.preview.stage.zesty.io/api/generic.json` → set Heading = `jerseyNo` (a number)
2. Save the field. Open a content item using it. Click "Select Remote Items".
3. Scroll to **Malik Anderson** (jerseyNo=`0`)
4. **Observed:** card displays "Add Heading" placeholder text instead of "0"
5. All other players show their numeric jersey numbers correctly because they're non-zero (truthy)

## Fix

Replace each `value || placeholder` with explicit nullish handling and string coercion:

```tsx
const renderValue = (value: unknown, placeholder: string) => {
  if (value === null || value === undefined || value === '') return placeholder;
  return String(value);   // covers number, boolean, string uniformly
};

const headingValue = loading ? <Skeleton .../> : renderValue(heading, 'Add Heading');
// ...etc
```

Also tighten the `DisplayCardProps` types — `heading: string | number | boolean | null` reflects reality. Or better, type as `unknown` and force the renderValue helper at the boundary.

## Severity

**Minor.** Visual-only, no data loss or crash. But it makes integration-field outputs look broken whenever a numeric keyPath value is `0` (jersey number 0, weight 0, count 0, score 0, etc.) — common enough in real-world data that this will surface frequently.

## Related

- #4086 / #4083 (the search-crash from the same type-coercion family)
- A1 — `keyPathValuesToString` at `utils.ts:33` has the same falsy bug, breaking Resync drift detection for null↔0 transitions

geodem127:
Fixes #4093

## ROOT CAUSE:
 - `heading || "Add Heading"` (and the same pattern for `subHeading`, `detail`, and `item.key/value`) used a JavaScript falsy-check fallback, causing valid-but-falsy values like `0`, `false`, and `""` to be treated as missing and replaced with placeholder text
 - `DisplayCardProps` typed `heading`, `subHeading`, and `detail` as `string` despite the component already branching on `typeof X === "boolean"` at runtime, giving a false sense of type safety

## FIXES:
 - Added a `renderValue(value, placeholder)` helper that only substitutes the placeholder for `null | undefined | ""`, and returns `String(value)` for everything else — covering numbers, booleans, and strings uniformly
 - Replaced all four falsy-fallback sites (`heading`, `subHeading`, `detail`, `item.key`, `item.value`) with `renderValue` calls, also removing the now-redundant `typeof X === "boolean"` branches
 - Updated `DisplayCardProps` types for `heading`, `subHeading`, `detail`, and `details` to reflect the actual runtime value types (`string | number | boolean | null`)

### DIFF
--- a/src/shell/components/FieldTypeIntegration/Shared/DisplayCard.tsx
+++ b/src/shell/components/FieldTypeIntegration/Shared/DisplayCard.tsx
@@ -36,7 +36,10 @@ const DisplayCard = ({
   showPlayIcon,
   loading,
 }: DisplayCardProps) => {
-  const renderValue = (value: unknown, placeholder: string) => {
+  const renderValue = (
+    value: string | number | boolean | null | undefined,
+    placeholder: string
+  ) => {
     if (value === null || value === undefined || value === "")
       return placeholder;
     return String(value);
--- a/src/shell/components/FieldTypeIntegration/Shared/DisplayCard.tsx
+++ b/src/shell/components/FieldTypeIntegration/Shared/DisplayCard.tsx
@@ -14,12 +14,12 @@ import { IntegrationTypes } from "../../../services/types";
 
 export type DisplayCardProps = {
   type: IntegrationTypes;
-  heading: string;
-  subHeading?: string;
+  heading: string | number | boolean | null;
+  subHeading?: string | number | boolean | null;
   thumbnail?: string;
   rootPath?: string | null;
-  detail?: string;
-  details?: Record<string, string | number>[];
+  detail?: string | number | boolean | null;
+  details?: Record<string, string | number | boolean | null>[];
   mediaVariant?: "rounded" | "square";
   showPlayIcon?: boolean;
   loading?: boolean;
@@ -36,6 +36,12 @@ const DisplayCard = ({
   showPlayIcon,
   loading,
 }: DisplayCardProps) => {
+  const renderValue = (value: unknown, placeholder: string) => {
+    if (value === null || value === undefined || value === "")
+      return placeholder;
+    return String(value);
+  };
+
   const [noImage, setNoImage] = useState(false);
   const isVideoType = ["video", "youtube", "mux"].includes(type);
   const isSpecialType = ["shopify", "youtube", "mux", "classy"].includes(type);
@@ -61,25 +67,19 @@ const DisplayCard = ({
 
   const headingValue = loading ? (
     <Skeleton width={type === "simple" ? "90%" : "55%"} height="26px" />
-  ) : typeof heading === "boolean" ? (
-    String(heading)
   ) : (
-    heading || "Add Heading"
+    renderValue(heading, "Add Heading")
   );
   const subHeadingValue = loading ? (
     <Skeleton width="90%" />
-  ) : typeof subHeading === "boolean" ? (
-    String(subHeading)
   ) : (
-    subHeading || "Add Subheading"
+    renderValue(subHeading, "Add Subheading")
   );
   const thumbnailValue = thumbnail || null;
   const detailValue = loading ? (
     <Skeleton width="70px" />
-  ) : typeof detail === "boolean" ? (
-    String(detail)
   ) : (
-    detail || "Add Detail"
+    renderValue(detail, "Add Detail")
   );
 
   const renderCard = () => {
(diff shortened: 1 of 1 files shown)


agalin920 (dismissed):

src/shell/components/FieldTypeIntegration/Shared/DisplayCard.tsx
```diff
@@ -36,6 +36,12 @@ const DisplayCard = ({
   showPlayIcon,
   loading,
 }: DisplayCardProps) => {
+  const renderValue = (value: unknown, placeholder: string) => {
```
finnar-bin: Why is the type of the value parameter `unknown` when the arguments being passed when this function is used all have clearly defined types above?

finnar-bin (changes requested):

src/shell/components/FieldTypeIntegration/Shared/DisplayCard.tsx
geodem127: Fixed — changed to `string | number | boolean | null | undefined`. The `undefined` covers optional props (`subHeading`, `detail`) and optional-chained record accesses (`item?.key`, `item?.value`); everything else matches the `DisplayCardProps` union directly.

agalin920 (approved):

finnar-bin (approved):

geodem127:

geodem127:
The Integration Fields component has been refactored. The necessary changes for the fix are included in PR #4118

geodem127:
Closed — fix available in PR [#4118.](https://github.com/zesty-io/manager-ui/pull/4118)

PR #4115, opened 2026-05-19
Schema - Fix reconfigure display options not reflecting updated type

### LINKED ISSUE #4112: Manager UI - Reconfigure Integration Display Options Not Saving
## Problem

When a user reconfigures display options for an integration field, they changes are not saved

## Screenshots

If applicable, add the following:

- https://www.loom.com/share/a2b19617c72549088719d3e2a6074bdf

## Steps to reproduce

1. Go to 'Schema'
2. Create an integration field connection
3. Click on reconfigure display options
4. Edit card type
5. Reconfigure field mapping
6. Save changes
7. Edits are not saved

A proposed fix or expected behavior (if applicable).

- Save user edits

geodem127:
## ROOT CAUSE:
- `Details.tsx` passed `fieldData?.settings?.integrationFieldConfig` as the prop to `IntegrationFieldConfigure` for update fields (the initial snapshot, never updated mid-session), which was the root cause of both issues below
- `Configure/index.tsx` displayed the card type from the `integrationFieldConfig` **prop**, so after reconfiguring the \"Display Items as\" label always showed the original type instead of the newly selected one
- `FieldFormProvider.tsx` validation read `integrationFieldConfig` from `fieldData.settings` (not `formData`) for update fields, causing the \"Incomplete API Configuration\" error to persist even after a valid reconfiguration was completed

## FIXES:
- **`Details.tsx`**: Remove `isUpdateField` branch for `integrationFieldConfig`; always read from `formData[\"integrationFieldConfig\"]` so the prop passed to `IntegrationFieldConfigure` reflects in-session reconfiguration (closes the root cause)
- **`Configure/index.tsx`**: `onClose` resets `endpoint`/`headers`/`type`/`keyPaths` back to prop values so dismissing the wizard without saving does not leave stale values in the read-only display
- **`Configure/index.tsx`**: `isConnected` derived from local state (`!!endpoint && !!type`), keeping the button label and connected-state section in sync with the live wizard state
- **`Configure/index.tsx`**: Move initial state sync for update fields from `useState` initializer into `useEffect` so a deferred `integrationFieldConfig` prop (not yet available on first render) is handled correctly; guard config card with `isUpdate || isConnected` to prevent a flash while the effect hasn't run yet
- **`ConfigureDisplayOptions.tsx`**: Remove redundant `closeForm?.()` from `handleSave` — `onSave` already calls `setIsFormOpen(false)`, and calling `onClose` afterward would reset local state to stale prop values before the parent re-renders with the new config
- **`FieldFormProvider.tsx`**: Always read `formData.integrationFieldConfig` for validation (removes the `isUpdateField` branch that read from `fieldData.settings` instead)

## TESTING PLAN:
- `npx tsc --noEmit` — zero errors in any changed file ✅
- `npm start` — webpack bundle compiled successfully, no errors ✅
- Acceptance criteria: `./node_modules/.bin/cypress run --spec "cypress/e2e/schema/integration.spec.js"`
- New E2E specs added:
  - Cancelling mid-reconfigure leaves the displayed type unchanged
  - Updates the displayed type after a successful reconfigure (and verifies PUT body)

## SCREEN RECORDING:

fix-4112-validation


src/shell/components/FieldTypeIntegration/Configure/index.tsx
```diff
@@ -180,7 +180,7 @@ const IntegrationFieldConfigure = ({
               <InputBase
                 data-cy="integrationApiUrl"
                 readOnly
-                value={integrationFieldConfig?.endpoint}
```
finnar-bin, line 183: What's the purpose of the fallback value here?

src/shell/components/FieldTypeIntegration/Configure/index.tsx
```diff
@@ -198,7 +198,7 @@ const IntegrationFieldConfigure = ({
               <InputBase
                 data-cy="integrationDisplayType"
                 readOnly
-                value={integrationFieldConfig?.type}
```
finnar-bin, line 201: Same comment here, why add a fallback value?

src/shell/components/FieldTypeIntegration/Configure/index.tsx
geodem127: Good catch. The fallback is redundant — this section only renders when `isConnected` is `true`, which already guarantees `integrationFieldConfig?.endpoint` is truthy, and `endpoint` state is initialized from that prop on mount. After a reconfiguration the state holds the updated value, which is what we want to display. Simplified to use the state variable directly.

src/shell/components/FieldTypeIntegration/Configure/index.tsx
geodem127: Same fix applied here — the fallback to `integrationFieldConfig?.type` is redundant for the same reason: the display only renders when `isConnected` is `true` (guaranteeing the prop is truthy), and `type` state is initialized from that prop on mount. Simplified to `value={type}`.

geodem127:
PR #4153 fixes failing and flaky tests

agalin920 (dismissed):

finnar-bin (dismissed):

agalin920 (approved):

finnar-bin (approved):

PR #4117, opened 2026-05-19
Cypress: Improve spec speed and reliability

### LINKED ISSUE #4105: Improve cypress test runner speed
- Set a budget of 30mins
- Look into parallelization, cypress has a plugin for parallelization

### Top 20 slowest specs (with fixable signals)

  Rank: 1
  Time: 260s
  Spec: content/content.spec.js
  it: 44
  cy.visit: 8
  hard-waits: 500ms
  Diagnosis: Largest spec — should be split into 2–3 files. ~5.9s/test is acceptable;
  volume
    is the issue.
  ────────────────────────────────────────
  Rank: 2
  Time: 248s
  Spec: content/actions.spec.js
  it: 16
  cy.visit: 10
  hard-waits: 0
  Diagnosis: ~15.5s per test — almost certainly auto-retried assertions or unmocked
  network.
     Worth profiling.
  ────────────────────────────────────────
  Rank: 3
  Time: 224s
  Spec: reports/activity-log/activity-log.spec.js
  it: 18
  cy.visit: 24
  hard-waits: 0
  Diagnosis: More cy.visit calls than tests → many full page reloads. Reuse navigation.
  ────────────────────────────────────────
  Rank: 4
  Time: 194s
  Spec: schema/integration.spec.js
  it: 15
  cy.visit: 2
  hard-waits: 0
  Diagnosis: ~13s/test on schema CRUD. Profile network.
  ────────────────────────────────────────
  Rank: 5
  Time: 193s
  Spec: studio/studio-wrapper.spec.js
  it: 19
  cy.visit: 1
  hard-waits: 0
  Diagnosis: Iframe-based (cypress-iframe is in deps). Iframe waits are slow — confirm
    there’s no cy.wait(timeMs) masquerading as cypress-iframe internals.
  ────────────────────────────────────────
  Rank: 6
  Time: 133s
  Spec: seo/redirects/redirects.spec.js
  it: 11
  cy.visit: 10
  hard-waits: 10,000ms
  Diagnosis: 2× cy.wait(5000) — high-confidence wins. Replace with cy.intercept +
  assertion.
    (this is the file you have selected)
  ────────────────────────────────────────
  Rank: 7
  Time: 121s
  Spec: settings/tabs.spec.js
  it: 6
  cy.visit: 9
  hard-waits: 4,000ms
  Diagnosis: 4 × cy.wait(1000) — replace each. ~20s/test on 6 tests suggests reload-heavy
    flow.
  ────────────────────────────────────────
  Rank: 8
  Time: 119s
  Spec: blocks/tests/starter-blocks.spec.js
  it: 10
  cy.visit: 2
  hard-waits: 0
  Diagnosis: ~12s/test, likely block install/build waits — cy.intercept opportunity.
  ────────────────────────────────────────
  Rank: 9
  Time: 113s
  Spec: code/all-files-page.spec.js
  it: 8
  cy.visit: 6
  hard-waits: 0
  Diagnosis: ~14s/test, 6 reloads on 8 tests.
  ────────────────────────────────────────
  Rank: 10
  Time: 93s
  Spec: schema/field.spec.js
  it: 20
  cy.visit: 1
  hard-waits: 6,200ms
  Diagnosis: 3 hard waits (longest 5s) + 20 tests in one file = also a split candidate.
  ────────────────────────────────────────
  Rank: 11
  Time: 92s
  Spec: settings/workflows.spec.js
  it: 14
  cy.visit: 2
  hard-waits: 3,500ms
  Diagnosis: 3 hard waits.
  ────────────────────────────────────────
  Rank: 12
  Time: 90s
  Spec: content/meta.spec.js
  it: 4
  cy.visit: 4
  hard-waits: 500ms
  Diagnosis: ~22s per test on only 4 tests — very high per-test cost. Profile.
  ────────────────────────────────────────
  Rank: 13
  Time: 88s
  Spec: schema/repeater-field.spec.js
  it: 20
  cy.visit: 1
  hard-waits: 0
  Diagnosis: Volume; split candidate.
  ────────────────────────────────────────
  Rank: 14
  Time: 74s
  Spec: content/workflows.spec.js
  it: 4
  cy.visit: 1
  hard-waits: 0
  Diagnosis: ~18s/test.
  ────────────────────────────────────────
  Rank: 15
  Time: 73s
  Spec: blocks/tests/blocks.spec.js
  it: 7
  cy.visit: 0
  hard-waits: 0
  Diagnosis:
  ────────────────────────────────────────
  Rank: 16
  Time: 72s
  Spec: schema/models.spec.js
  it: 8
  cy.visit: 5
  hard-waits: 0
  Diagnosis:
  ────────────────────────────────────────
  Rank: 17
  Time: 69s
  Spec: content/navigation.spec.js
  it: –
  cy.visit: –
  hard-waits: –
  Diagnosis:
  ────────────────────────────────────────
  Rank: 18
  Time: 59s
  Spec: content/item-list-table.spec.js
  it: –
  cy.visit: –
  hard-waits: –
  Diagnosis:
  ────────────────────────────────────────
  Rank: 19
  Time: 54s
  Spec: content/redirects.spec.js
  it: –
  cy.visit: –
  hard-waits: –
  Diagnosis:
  ────────────────────────────────────────
  Rank: 20
  Time: 53s
  Spec: content/used-blocks.spec.js
  it: –
  cy.visit: –
  hard-waits: –
  Diagnosis:

  Free wins (do these first)

  1) Five “specs” are not specs. specPattern is cypress/e2e/**/*.{js,jsx,ts,tsx}, so these
  page-object classes get loaded as specs and each pays Cypress startup overhead (~1–5s
  each, ~10s total):

  - cypress/e2e/blocks/pages/AllBlocksPage.js
  - cypress/e2e/blocks/pages/BlockPage.js
  - cypress/e2e/content/pages/ContentItemPage.js
  - cypress/e2e/schema/pages/SchemaPage.js

  Fix: tighten the pattern.

  specPattern: “cypress/e2e/**/*.spec.{js,jsx,ts,tsx}”

  :warning:  One file relies on the loose pattern: cypress/e2e/content/comments.js has 6 real tests
   but no .spec. suffix. Rename it to comments.spec.js before tightening the pattern.

  2) Kill all cy.wait(<ms>) in the top specs. From the audit there are 25,200ms of hard
  waits concentrated in 7 specs — and that’s just the floor (each one usually slows the
  test by more than the wait because of the retries it was masking). High-value targets:

  ┌─────────────────────────────────┬──────────────┬───────┐
  │              Spec               │ hard-wait ms │ calls │
  ├─────────────────────────────────┼──────────────┼───────┤
  │ seo/redirects/redirects.spec.js │ 10,000       │ 2     │
  ├─────────────────────────────────┼──────────────┼───────┤
  │ media/files.spec.js             │ 7,000        │ 4     │
  ├─────────────────────────────────┼──────────────┼───────┤
  │ schema/field.spec.js            │ 6,200        │ 3     │
  ├─────────────────────────────────┼──────────────┼───────┤
  │ settings/tabs.spec.js           │ 4,000        │ 4     │
  ├─────────────────────────────────┼──────────────┼───────┤
  │ settings/workflows.spec.js      │ 3,500        │ 3     │
  ├─────────────────────────────────┼──────────────┼───────┤
  │ schema/activity-log.spec.js     │ 3,000        │ 1     │
  ├─────────────────────────────────┼──────────────┼───────┤
  │ releases/members.spec.js        │ 2,000        │ 1     │
  └─────────────────────────────────┴──────────────┴───────┘

  Standard replacement pattern:

  cy.intercept(“POST”, “**/redirects”).as(“createRedirect”);
  clickSaveButton();
  cy.wait(“@createRedirect”);

  3) Split the two giant specs. content/content.spec.js (44 its) and
  schema/field.spec.js/schema/repeater-field.spec.js (20 its each) are also the natural
  shard boundaries for whatever parallelization scheme you adopt next — splitting them
  gives a load-balancer something to work with.

  Things worth profiling (per-test cost is anomalously high)

  These three are the most suspicious because the per-test cost is way out of line with the
   rest of the suite:

  - content/meta.spec.js — 22s/test on 4 tests
  - content/workflows.spec.js — 18s/test on 4 tests
  - content/actions.spec.js — 15.5s/test on 16 tests

  Run with DEBUG=cypress:* npx cypress run --spec cypress/e2e/content/meta.spec.js locally
  and look for repeated unmocked network calls or defaultCommandTimeout-eating retries.

  Expected impact

  - Removing the 4 non-spec files + renaming comments.js → ~10s shaved, near zero risk.
  - Replacing all cy.wait(ms) in the listed specs → ~30–60s shaved (the wait itself plus
  the retries the wait was hiding).
  - Splitting content/content.spec.js and the two schema 20-test specs → no time saved
  standalone, but unlocks better parallelization.
  - Combined with the 6-way matrix shard from the previous message → expect ~8–10 min total
   CI.

  Want me to open a PR for the spec-pattern fix and the cy.wait replacements in
  seo/redirects/redirects.spec.js? That’s the lowest-risk, highest-confidence starting
  point.

✻ Sautéed for 1m 54s

※ recap: You’re speeding up a ~1-hour Cypress run; I audited timings from the latest green
  CI and identified the slowest specs plus quick wins. Next: open a PR tightening the spec
  pattern and replacing hard waits in seo/redirects. (disable recaps in /config)

finnar-bin:
## Summary

- Replaced `cy.wait(N)` hard waits with deterministic intercept/assertion waits across multiple spec files
- Fixed a `cy.intercept(url, "DELETE")` stub in `redirects.spec.js` that was preventing actual server deletions (should be `cy.intercept("DELETE", url)`)
- Fixed a Cypress alias reset bug in `workflows.spec.js` where `cy.wrap().as()` in `before()` is cleared before the first `it()` runs — replaced with a closure variable
- Switched `models.spec.js` to use `uuidv4()` for unique model names and derive the reference ID dynamically
- Updated CLAUDE.md to document preferred selector strategy (`getBySelector`), ban on hard waits, and preference for `uuidv4` over timestamps for unique test data
- Isolated publish-related specs (`content/actions`, `content/list`, `content/redirects`, `settings/workflows`) to a dedicated runner (runner 4) to prevent cross-runner interference caused by `workflows.spec.js` creating a publish-blocking workflow label
- Added `all_tests_passed` aggregator job to CI — with test parallelization, matrix runners register as `run_tests (N)` checks rather than a single `run_tests` job, so any branch protection rule requiring `run_tests` never resolves. The new job aggregates all runner results into a single status check that branch protection can target instead.

## Measured results (headless Electron, dev instance)

| Spec | Before | After | Saved |
|------|--------|-------|-------|
| `seo/redirects/redirects.spec.js` | 133s | 103s | **30s** |
| `settings/workflows.spec.js` | 92s | 90s | 2s |

Remaining changed specs (`field.spec.js`, `media/folders.spec.js`, `content/singlePageAnalytics.spec.js`) have deterministic savings from removed hard waits but the delta is within dev API variance on a single run.

## Deterministic savings floor

| Source | Detail | Saved |
|--------|--------|-------|
| Hard waits removed | 21,700ms across 19 call sites | **21.7s** |


cypress.config.js
```diff
@@ -36,4 +36,5 @@ module.exports = defineConfig({
     specPattern: "cypress/e2e/**/*.spec.{js,jsx,ts,tsx}",
     testIsolation: false,
   },
+  retries: 1,
```
agalin920, line 41: What does this do?

cypress/e2e/content/workflows.spec.js
```diff
@@ -111,7 +109,10 @@ describe("Content Item Workflows", () => {
     cy.get("body").type("{esc}");
   });
 
-  it("Cannot publish a content item if label with allowPublish is missing", () => {
+  // Skipped: creating an allowPublish:true label on the shared dev instance blocks
+  // concurrent specs from publishing until cleanup runs. Needs a mock-based approach
+  // to avoid cross-runner interference.
+  it.skip("Cannot publish a content item if label with allowPublish is missing", () => {
```
agalin920: Could we not simply move this spec into the concurrent spec that is dependent? Not a fan of the mock based approach but this one is tricky. Can we get a ticket to re add this linked?

cypress/e2e/search/search-page.spec.js
```diff
@@ -73,15 +73,7 @@ describe("Global Search: Search Page", () => {
...
-    cy.waitOn(
-      "https://8-f48cf3a682-7fthvk.api.dev.zesty.io/v1/search/items*",
-      () => {
-        cy.visit(
-          "/search?q=somerandomstringthatdoesnotexist&sort=created&user=5-84d1e6d4ae-s3m974&datePreset=today"
-        );
-      }
-    );
-
+    // already at the filtered URL from the previous test
```
agalin920: Shouldnt we try to decouple tests from previous tests as much as possible? Especially with the "retries" you added if this fails and this is attempted to retry it wont be in the state it needs to be

cypress/e2e/reports/activity-log/activity-log.spec.js
```diff
@@ -58,17 +80,11 @@ describe("Reports > Activity Log > Home", () => {
...
-      const threeMonthsAgo = addMonths(today, -3);
-      cy.location("search").should(
-        "not.eq",
-        `?from=${format(threeMonthsAgo, "yyyy-MM-dd")}&to=${format(
-          today,
-          "yyyy-MM-dd"
-        )}`
-      );
+    it("Filters block items", () => {
+      // already at /resources with default date range and data loaded from previous test
```
agalin920: Originally visited ?from=2020-07-14&to=2020-07-16 to verify URL params are read on initial visit. Now it runs after the previous test mutated the URL via filter clicks, so the
  assertion passes for a different reason URL mutated by filters, not by visit params

agalin920 (changes requested):

cypress.config.js
finnar-bin: Some tests are failing due to API timeouts and intermittent latency. Implementing an automatic retry mechanism will allow these tests a second chance to pass, reducing flaky failures and saving us from having to manually rerun the entire CI pipeline.

cypress/e2e/content/workflows.spec.js
finnar-bin: Yeah not really a fan of mocks as well, just added it as a comment as a possible fix. I'm still looking into ways to add this without affecting other tests. I'll look around if there's a way that a certain spec could have this as a dependency to avoid cross-runner issues.

cypress/e2e/search/search-page.spec.js
finnar-bin: I see what you mean. I went out and removed what I thought were redundant `cy.visit()` calls since these added to the overall test runtime but yeah I understand that this causes a huge reliance on the fact that the previous test passes. I'll revert these.

cypress/e2e/reports/activity-log/activity-log.spec.js
finnar-bin: The previous test only loaded the page and asserted if it automatically added the `from` and `to` url search params. Then this test asserts the other filters.

cypress/e2e/content/workflows.spec.js
finnar-bin: I think I may have a solution for this, which is to group tests that perform publish actions and this test which adds the publish label in 1 runner, this way we avoid cross-runner interference.

finnar-bin: Updated the PR, added a dedicated test runner which runs content/workflow.spec.js and tests that involves publishing content items. This way we can avoid cross-runner interference with regards to content item publishing.

cypress/e2e/search/search-page.spec.js
finnar-bin: Reverted changes to the `cy.visit()` so that we retain decoupling of each test

agalin920 (dismissed):

agalin920 (dismissed):

agalin920 (approved):

PR #4119, opened 2026-05-20
Cypress: Add missing integration field regression tests

### LINKED ISSUE #4104: Cypress - Add missing integration field regression tests
## Background

A manual regression suite (R1–R15) was run against the local dev build during the FieldTypeIntegration folder restructure (PR #4103) to confirm no functional regressions. Comparing those checks against the existing `cypress/e2e/schema/integration.spec.js`, **5 of the 15 manual checks have no automated coverage**, and **9 of the 11 open integration-field defects** (#4086, #4088–#4096) lack any Cypress assertion that would catch the bug if it regressed after a fix.

Filing this as a single issue so the team can prioritize/triage. Each row below is independently actionable.

## Audit: manual regression vs current Cypress coverage

| ID | Manual regression test | Currently in cypress | Notes |
|---|---|---|---|
| R1 | Field renders on item route | ✓ implicit | "Create Content Item" visits the item route |
| R2 | Selection dialog opens | ✓ | Lines 110–113 |
| R3 | Search "Marcus" → 1 row | ✓ | Lines 115–124 |
| R4 | Search "xxxxxx" → no-results | ✓ | Lines 125–131 |
| R5 | Search Again clears + focuses | ✓ | Lines 133–139 |
| **R6** | **`maxValue` selection lockout disables checkboxes 6–10** | **❌ MISSING** | The `maxValue` schema setting feeds `IntegrationFieldSelect.maxItems` via `Field.tsx:805`. Setting `maxValue=N` should disable checkboxes once N items are selected. No test today. |
| R7 | Done commits selections to field | ✓ | Lines 140–158 |
| R8 | Save persists ("Save Item" toast) | ✓ | Lines 209–216 |
| **R9** | **Reload preserves saved selections** | **❌ MISSING** | The Save Item test asserts the toast but never reloads to confirm the snapshot survives a page refresh. A round-trip assertion is the only way to catch silent data-loss between client state and persisted record. |
| R10 | Drag-reorder | ✓ | Lines 174–194 (uses `dragstart`/`drop` + DataTransfer) |
| R11a | More-options → View Raw JSON | ✓ | Lines 159–172 |
| R11b | More-options → Remove | ✓ | Lines 196–207 |
| R12 | Field-edit page renders | ✓ implicit | Reconfigure test visits this route |
| R13 | Reconfigure dialog opens at Step 1 (Display Type) | ✓ partial | Asserts dialog exists but doesn't assert which step is active or that current type is pre-selected |
| **R14** | **Reconfigure Step 2 pre-fills existing keyPath values** | **❌ MISSING** | The test (line 220) changes one value (`subHeading` → `position`) and asserts the PUT body, but never asserts that the *unchanged* keyPaths (Item ID, Heading) retained their original values. The bug this would catch: a regression where the Step 2 form opens empty on update, forcing users to re-enter every keyPath. |
| **R15** | **Image-type display rendering with real thumbnails in selection dialog** | **❌ MISSING** | Field creation for `image` type is tested (lines 56–75) but the **runtime rendering** of the selection dialog with image cards (broken-image fallback, CardMedia onLoad/onError) is not. |

## Audit: open defects vs preventive Cypress coverage

For each open issue, what test would have caught the bug and prevent its regression after a fix:

| Issue | Suggested Cypress test | Where to add |
|---|---|---|
| **#4086** Search crash on non-string keyPath | Configure a field with `heading = jerseyNo` (numeric key in `cypress/fixtures/integration/generic.json`). Visit content item, open selector, type any character in the search box. Assert: dialog stays open, no Sentry "Oops" screen, filter result count is consistent. **Critical gate** — the original bug crashed the entire content editor; without this test, the same regression would slip through. | New test under `describe("Item Selection")` |
| #4087 Reconfigure persists changes | ✓ Partially covered by line 220 test added in #4084. Could expand to also flip the `type` from `simple` → `text`, change `endpoint` (after #4089 fixed), and re-PUT headers. | Expand existing `describe("Reconfigure Display Options")` |
| #4088 Green debug border | Computed-style assertion on the dialog paper: `expect(border).not.to.contain('green')`. Or full visual snapshot with `cypress-image-snapshot`. | New `it("does not have debug styling")` under `describe("Item Selection")` |
| #4089 Endpoint immutable in Reconfigure | Open Reconfigure → Step 1 → assert that a `Back` button exists alongside Cancel/Next. (Will fail until the bug is fixed; prevents future regression once it is.) | New test under `describe("Reconfigure Display Options")` |
| **#4091** Non-unique itemId silent corruption | Configure with `itemId = position` (the fixture has 2 players each with `Shooting Guard` / `Point Guard` / etc.). Click checkbox 0. Assert the title bar shows `"1 Selected"` AND `cy.get('[data-cy="integrationSelectCard"] input:checked')` returns exactly 1 item, not 2. | New test under `describe("Item Selection")` |
| #4092 Empty Configure on bad API shapes | Parameterize over four mocked responses (`null`, `[]`, `{name:"x"}`, `[1,2,3]`) using `cy.intercept`. For each: walk through Connect → Display Type → Next, then assert the Configure step shows an actionable error (e.g. `cy.get('[data-cy="integrationApiShapeError"]')`). | New `describe("Unsupported API shapes")` block |
| #4093 DisplayCard `0` shows placeholder | Use a fixture item with `jerseyNo: 0`. Configure heading=`jerseyNo`. Open selector. Assert the card text contains `"0"` and does NOT contain `"Add Heading"`. | New test under `describe("Display rendering")` (new describe) |
| #4094 Add HTTP Header double-click race | `cy.get('[data-cy="addHeaderButton"]').dblclick();` then assert 2 new rows added (3 total). | New test under existing `describe("Add Field")` |
| #4095 Public API double-encodes | After Save Item, `cy.request("/-/instant/<itemZUID>.json")` and assert `Array.isArray(body.data[0].content.players_simple)` is true. (Will fail until the public API stops string-encoding.) | New test under `describe("Item Selection")` after Save |
| #4096 Drift detection misses falsy↔falsy | Save snapshot with a falsy value, intercept the next selector-open fetch to return a different falsy value, reopen selector, assert the Refresh (resync) icon is present on the changed row. | New `describe("Resync drift detection")` block |

## Spec-quality fix unrelated to coverage

Lines **279 and 289** of `cypress/e2e/schema/integration.spec.js` declare `it()` blocks **inside the `connectToEndpoint` helper function**:

```js
function connectToEndpoint(endpoint, type, apiData) {
  ...
  if (type === "simple") {
    it("Add HTTP Headers", () => { ... });   // ← line 279
    it("Remove HTTP Headers", () => { ... }); // ← line 289
  }
  ...
}
```

This is a Mocha anti-pattern — `it()` may only be called during the `describe()` registration phase, not from inside another test's body. These two tests **never actually register and run**. They should be either:
- Moved into the top-level `describe("Add Field")` block as proper standalone tests, or
- Deleted if they're superseded by other coverage

This is independent of the new tests above; flagging it because it surfaces during the same audit.

## Suggested order of implementation

If anyone picks this up incrementally, the highest-value tests to add first (in this order):

1. **#4086 numeric-keyPath search** — the regression check for the worst bug found in the integration field
2. **R9 reload-persistence** — the most obvious coverage cliff
3. **R14 Reconfigure Step 2 pre-fill assertion** — strengthens the existing reconfigure test against the same bug class
4. **R6 `maxValue` lockout** — verifies a feature that's wired but invisible without a test
5. Then everything else as bandwidth allows

## Test fixtures needed

A few of these tests need fixture data not present in `cypress/fixtures/integration/generic.json` today:
- **#4093** needs an item with `jerseyNo: 0` (Malik Anderson would qualify if his number were changed; currently he's `0` already on stage but not in the fixture — verify)
- **#4091** needs at least 2 items sharing the same `position` (current fixture has Jalen Thompson and Hiroshi Tanaka both as "Shooting Guard" — already suitable)
- **#4086** needs the same fixture (Heading is configured to `jerseyNo`, no fixture changes needed)
- **#4092** needs no fixture changes — uses `cy.intercept` to mock the four bad-shape responses

geodem127:
### **RCA:**
- Two `it()` blocks nested inside the `connectToEndpoint()` helper function were never registered by Mocha and silently skipped on every test run (Mocha anti-pattern)
- Five coverage gaps identified in the post-#4103 regression audit: R6 (maxValue lockout), R9 (reload persistence), R14 (Reconfigure Step 2 pre-fill), #4094 (double-click header race), and the nested-`it()` anti-pattern

### **FIX:**
- **Anti-pattern removed** — deleted `if (type === "simple") { it()... }` from `connectToEndpoint()`; tests moved into a proper `context("HTTP Headers")` with `beforeEach` setup
- **R9** — added `"Reload preserves saved selections"` after `"Save Item"`: reloads the page and asserts 2 persisted cards remain
- **R14** — added pre-fill assertions in `"Reconfigure Display Options"` to verify `itemId` and `heading` retain their original values before changing `subHeading`
- **#4094** — added `"Double-click does not add duplicate header rows"`: `dblclick` on `addHeaderButton` should yield exactly 2 rows
- **R6** — added `"Disables unselected checkboxes once the maxValue limit is reached"` in a new `describe("maxValue Lockout")` block; uses `seed:content` with a new fixture (`fixtures/integration/maxvalue.json`) that seeds a dedicated model with `maxValue: 2` and a properly configured integration field

### **TESTING PLAN:**
- New fixture `cypress/fixtures/integration/maxvalue.json` seeds a model with title + integration field configured with `maxValue: 2`, `endpoint`, `type: "simple"`, and `keyPaths`
- Existing spec runs unchanged; new tests added to the end of the suite and to `context("HTTP Headers")`
- R6 test intercepts `**/get-url?url=*` with `genericApi` fixture data to avoid external network dependency
- Verified spec structure: 20 `it()` blocks, zero nested `it()` in helper functions

___

> ### NOTE: PR required https://github.com/zesty-io/manager-ui/pull/4100, this fixes the issue with "Add header" buttton

 
Closes #4104


cypress/e2e/schema/integration.spec.js
finnar-bin, line 1: use the `cy.getBySelector` function

cypress/e2e/schema/integration.spec.js
```diff
@@ -239,6 +298,47 @@ describe("Integration Field", () => {
...
+
+      cy.get(
+        `[data-cy="field:players"] [data-cy="integrationSelectItemsButton"]`
+      ).click();
+      cy.wait("@maxValueGetUrl");
+      cy.get('[data-cy="integrationSelectionFormDialog"]').should("exist");
+
+      cy.get(
+        '.integrationSelectionFormListContainer [data-cy="integrationSelectCard"]:eq(0) input'
+      ).check({ force: true });
```
finnar-bin: Why is `force` needed here? If the UI isn't behaving as expected without this then wouldn't that mean there's something wrong that needs to be resolved?

cypress/e2e/schema/integration.spec.js
```diff
@@ -239,6 +298,47 @@ describe("Integration Field", () => {
...
+      cy.get(
+        '.integrationSelectionFormListContainer [data-cy="integrationSelectCard"]:eq(1) input'
+      ).check({ force: true });
```
finnar-bin: same comment here

finnar-bin (changes requested):

cypress/e2e/schema/integration.spec.js
```diff
@@ -98,6 +98,57 @@ describe("Integration Field", () => {
...
+          .should("have.length", 1);
+      });
+
+      it("Double-click does not add duplicate header rows", () => {
+        cy.get('[data-cy="addHeaderButton"]').dblclick();
+
+        cy.get('[data-cy="integrationHeadersContainer"]')
+          .children()
+          .should("have.length", 2);
+      });
```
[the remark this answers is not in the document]
geodem127: Added a comment in the test documenting this dependency: the assertion relies on `ConnectToApi.tsx` using the stale-closure spread (`{ ...headersLocal, [keyId]: ... }`) rather than the functional updater, so both click handlers from `dblclick` capture the same snapshot and the second write overwrites the first. The comment explicitly notes that switching to the functional updater would break this test even though that refactor is correct.

cypress/fixtures/integration/maxvalue.json
```diff
@@ -0,0 +1,43 @@
...
+    },
+    {
+      "label": "Players",
+      "name": "players",
+      "datatype": "integration",
+      "sort": 1,
+      "settings": {
+        "maxValue": 2,
+        "integrationFieldConfig": {
+          "endpoint": "https://8xbq19z1-dev.preview.stage.zesty.io/api/generic.json",
```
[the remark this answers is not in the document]
geodem127: Fixed. Changed the endpoint to `"https://example.invalid/generic.json"` to make it clear the value is intentional placeholder data — the test mocks the response via `cy.intercept`, so the real URL was just misleading.

cypress/e2e/schema/integration.spec.js
```diff
@@ -214,6 +265,14 @@ describe("Integration Field", () => {
...
       cy.get('[data-cy="toast"]').contains("Created Item");
     });
+
+    it("Reload preserves saved selections", () => {
+      cy.reload();
+
+      cy.get('[data-cy="integrationListValueContainer"]')
+        .children()
+        .should("have.length", 2);
+    });
```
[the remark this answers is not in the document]
geodem127: Fixed. Added `cy.intercept("**/get-url?url=*", { statusCode: 200, body: genericApi })` before `cy.reload()`, consistent with the `maxValue Lockout` and `Reconfigure Display Options` tests. The assertion is now purely about persistence behavior and doesn't depend on the external endpoint being reachable during CI.

cypress/e2e/schema/integration.spec.js
geodem127: Done. Converted all `cy.get('[data-cy="..."]')` calls to `cy.getBySelector("...")` throughout the spec. Compound selectors with nested class-based children were refactored to use `.find()` chaining (e.g. `.find(".MuiCheckbox-root")`, `.find(".draggableCard")`).

cypress/e2e/schema/integration.spec.js
geodem127: Removed. Replaced `.find("input").check({ force: true })` with `.find(".MuiCheckbox-root").click()` — clicking the visible MUI checkbox span doesn't require `force`. Applied the same fix to all checkbox interactions in the spec, including the pre-existing "select 3 Items from the list" test.

cypress/e2e/schema/integration.spec.js
geodem127: Same fix applied here as well.

cypress/e2e/schema/integration.spec.js
```diff
@@ -98,68 +104,136 @@ describe("Integration Field", () => {
         });
       });
     });
+
+    context("HTTP Headers", () => {
+      beforeEach(() => {
+        cy.visit(`/schema/${Cypress.env("modelZUID")}/fields`);
```
finnar-bin, line 110: Is it really necessary to visit this url on each test item? This adds a significant time in the overall test runtime.

finnar-bin (dismissed):

agalin920 (approved):

geodem127:

geodem127:
The Integration Fields component has been refactored. The necessary changes for the fix are included in PR #4118.

geodem127:
Closed — fix available in PR [#4118.](https://github.com/zesty-io/manager-ui/pull/4118)

PR #4145, opened 2026-06-03
[Studio] - Support cross-region layout drag-and-drop

agalin920:
## Summary

Lifts the single-code-region constraint on Studio layout drag-and-drop so a block can be dragged out of one webview and dropped into another in one gesture, with both files saved sequentially.

- `useLayoutReorderState`: pending state is now keyed by region (`{ regions: Record<codeId, RegionState>, primaryCodeId }`), with a save loop over every affected webview. `mapSourceByLayoutStructure` does a donor-injection pass (pulls a moved block's source from the donor template), a removal pass (drops blocks no longer in the structure), and an anchored re-parent pass that uses `insertBefore` against the first existing layout-id's `previousSibling` so surrounding non-layout content (`<h1>`, Parsley `{{include ...}}` directives, blank lines) keeps its position.
- `StudioWrapper`: `useMultiPermission` gates UPDATE/PUBLISH across every affected codeId; replaced the single-codeId `mappedSource` boolean with a derived `hasPendingLayoutChanges` flag.
- `use-permissions`: extracted `hasPermission` as a non-hook helper and added `useMultiPermission` for AND-across-zuids checks.

### DIFF
--- a/src/apps/content-editor/src/app/views/ItemEdit/hooks/useLayoutReorderState.ts
+++ b/src/apps/content-editor/src/app/views/ItemEdit/hooks/useLayoutReorderState.ts
@@ -1,4 +1,4 @@
-import { useCallback, useRef, useState } from "react";
+import { useCallback, useMemo, useRef, useState } from "react";
 import { notify } from "../../../../../../../shell/store/notifications";
 import { LayoutBreadcrumbItem } from "./studioTypes";
 
@@ -16,8 +16,7 @@ const isLayoutBreadcrumbItem = (
       typeof (segment as LayoutBreadcrumbItem).label === "string"
   );
 
-type LayoutReorderState = {
-  codeId: string | null;
+type LayoutRegionState = {
   selector: string;
   orderedLayoutIds: string[];
   layoutStructure: LayoutStructureItem[];
@@ -25,11 +24,32 @@ type LayoutReorderState = {
   mappedSource: string;
 };
 
+type LayoutReorderState = {
+  regions: Record<string, LayoutRegionState>;
+  primaryCodeId: string;
+};
+
+// Apply a new layoutStructure to a cached template source. Used both for
+// single-region reorders and for the per-region post-processing of a cross-
+// region drag, where blocks may have moved between sibling/nested templates.
+//
+//  - donorSourceById: layoutId → outerHTML pulled from OTHER regions' cached
+//    templates. When a layoutId is referenced by the destination region's new
+//    structure but is missing from its cached template (because it was just
+//    dragged in from the donor region), the donor markup is injected here so
+//    the re-parent pass can place it correctly. With globally-unique
+//    layoutIds, donor lookups are unambiguous.
+//
+//  - Removal pass: any [data-layout-id] node whose id is NOT in the
+//    incoming layoutStructure is removed. This is how the SOURCE region of a
+//    cross-region drag loses the moved block. For single-region reorders the
+//    structure always covers every block, so the pass is a no-op.
 const mapSourceByLayoutStructure = (
   source: string,
-  layoutStructure: LayoutStructureItem[]
-) => {
-  if (!source || !layoutStructure?.length) return source;
+  layoutStructure: LayoutStructureItem[],
+  donorSourceById?: Map<string, string>
+): string => {
+  if (!source) return source;
 
   const parser = new DOMParser();
   const doc = parser.parseFromString(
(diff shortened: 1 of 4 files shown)


src/shell/hooks/use-permissions.js
```diff
@@ -20,54 +58,20 @@ export function usePermission(action, zuid = instanceZUID) {
...
-
-      default:
-        return false;
-    }
-  }, [user, role, action, zuid]);
+    if (!zuids?.length) return true;
+    return zuids.every((zuid) => hasPermission(user, role, action, zuid));
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [user, role, action, key]);
 }
```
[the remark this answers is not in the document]
agalin920: On the three sub-points:

1. **Empty array returns `true`** — vacuous-truth convention, same as `[].every(...)`. The only caller gates on `hasPendingLayoutChanges` before any UI-visible decision, so the empty case never reaches user action. Leaving as-is.
2. **Memo key is order-sensitive** — true in principle, but `pendingLayoutCodeIds` is 1–2 entries in practice and the "recomputation" is one `Array.every` over `hasPermission`. The cost is negligible. Leaving as-is.
3. **eslint-disable justification** — fair. The disable is there because `zuids` array identity changes every render but the joined `key` captures its content. I can add an inline justification comment in a follow-up if you'd like.

src/apps/content-editor/src/app/views/ItemEdit/hooks/useLayoutReorderState.ts
```diff
@@ -390,115 +561,158 @@ export const useLayoutReorderState = ({
...
+            layoutId: current.layoutId,
             breadcrumb: withCodeIdBreadcrumbRoot(
-              codeId,
+              primaryCodeId,
               reorderedBreadcrumb,
-              codeFileNameById[codeId]
+              codeFileNameById[primaryCodeId]
             ),
           };
         });
```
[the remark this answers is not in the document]
agalin920: The bridge enforces this at the source: `bridge.js` only sets `draggable="true"` on `.studio-selected` elements, and the `dragstart` handler early-returns on non-selected targets. So when a `REORDER_OUTPUT` reaches the host, the dragged block is always the selected one — `msg.selectedLayoutId === current.layoutId` is implicitly guaranteed. A host-side guard would be redundant under the current bridge contract. Leaving as-is.

src/apps/content-editor/src/app/views/ItemEdit/hooks/useLayoutReorderState.ts
```diff
@@ -39,6 +59,38 @@ const mapSourceByLayoutStructure = (
...
+    });
+
+    expectedLayoutIds.forEach((layoutId) => {
+      if (existingLayoutIds.has(layoutId)) return;
+      const donor = donorSourceById.get(layoutId);
+      if (!donor) return;
+      const holder = doc.createElement("div");
+      holder.innerHTML = donor;
+      const node = holder.firstElementChild;
+      if (node) root.appendChild(node);
```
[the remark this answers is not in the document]
agalin920: This is not a possible case

src/apps/content-editor/src/app/views/ItemEdit/hooks/useLayoutReorderState.ts
```diff
@@ -232,125 +337,191 @@ export const useLayoutReorderState = ({
...
+  const formatSavedFileNames = useCallback(
+    (results: Array<{ codeId: string; webView: any }>) => {
+      const names = results.map(
+        ({ codeId, webView }) =>
+          codeFileNameById[codeId] || webView?.fileName || codeId
+      );
+      if (names.length === 0) return "";
+      if (names.length === 1) return names[0];
+      if (names.length === 2) return `${names[0]} and ${names[1]}`;
+      return `${names.slice(0, -1).join(", ")}, and ${names[names.length - 1]}`;
```
finnar-bin, line 414: Why is the last element being removed here?

src/apps/content-editor/src/app/views/ItemEdit/hooks/useLayoutReorderState.ts
```diff
@@ -390,115 +561,158 @@ export const useLayoutReorderState = ({
...
+      // region so the header reflects the new home, even on cross-region drag.
+      if (primaryCodeId) {
         const reorderedBreadcrumb = Array.isArray(msg.selectedLayoutBreadcrumb)
           ? msg.selectedLayoutBreadcrumb.filter(isLayoutBreadcrumbItem)
           : [];
 
         onSelectedLayoutBreadcrumbChange((current) => {
-          if (!current || current.codeId !== codeId) return current;
+          if (!current) return current;
           if (!reorderedBreadcrumb.length) return current;
```
finnar-bin: Kind of a nitpick but could probably just combine these 2

src/shell/hooks/use-permissions.js
```diff
@@ -10,6 +10,44 @@ const getRole = (state) => state.userRole;
...
 
+export function hasPermission(user, role, action, zuid = instanceZUID) {
+  // "With great power comes great responsibility" - Benjamin Franklin Parker
+  if (user?.staff) {
+    return true;
+  }
+
+  if (role?.systemRole?.super) {
+    return true;
+  }
```
finnar-bin: Same nitpick here, probably could just combine these

src/apps/content-editor/src/app/views/ItemEdit/hooks/useLayoutReorderState.ts
agalin920: Not removed  `slice(0, -1)` takes all names except the last so they can be comma-joined, then the last is appended with "and": `"a, b, and c"`. The earlier `length === 1`/`=== 2` branches handle the shorter cases, so this line only runs for 3+ files.

src/apps/content-editor/src/app/views/ItemEdit/hooks/useLayoutReorderState.ts
agalin920: Done — combined into `if (!current || !reorderedBreadcrumb.length) return current;` in b4d0dec9e.

src/shell/hooks/use-permissions.js
agalin920: Done — merged into `if (user?.staff || role?.systemRole?.super) return true;` in b4d0dec9e.

src/apps/content-editor/src/app/views/ItemEdit/hooks/useLayoutReorderState.ts
agalin920: resolved

finnar-bin (dismissed):

finnar-bin (approved):

PR #4154, opened 2026-06-16
[Content] - Fix 404 on Take Offline after language switch

### LINKED ISSUE #4140: Language switch updates publishing context but fails to reload localized content item state (404 on "Take Offline")
## Problem

When switching languages in Zesty Manager (eg., from AU to NZ), the UI correctly updates to the active language's publishing ZUID. However, the underlying content item state remains stuck in the previous locale (eg. AU).

When a user triggers a "Take Offline" action on the new language version, the frontend sends a DELETE request for the correct localized publishing ZUID, but the backend rejects it with a 404 Not Found. This happens because the backend detects a structural mismatch between the requested publishing record and the un-updated content item state context being processed.

## Screenshots

[User Switches Language to NZ]
       │
       ├──► ✅ Publishing Store Updates  --> (Switches to NZ ZUID)
       │
       └──► ❌ Content Item Store Fails   --> (Stays stuck on AU Locale)

Image

## Steps to reproduce

1. Go to a published page in the Zesty Manager.
2. Change the status to Take Offline (for the first language, e.g., AU).
3. Switch the language context to the second language (e.g., NZ) via the UI switcher.
4. Verify that the UI reflects the new language's publishing record.
5. Click **Take Offline** on this new version (NZ).
6. Observe the DELETE request fails with a 404 Not Found.

**Workaround**: Hard-refreshing the browser or navigating directly via the localized URL forces both the publishing record and the content item state to load the correct locale together from scratch, allowing the action to succeed.

Please also note im able to replicate the issue once after a few more tests.

agalin920:
## Problem

Closes #4140

When switching languages on a content item's publishings tab (e.g. AU → NZ), the UI correctly updates to show the new locale's publishing record, but clicking **Take Offline** (or **Cancel**) on the new version fails with a `404 Not Found`. A hard refresh works around it.

## Root cause

`PublishState` renders its DataGrid columns from a `useMemo` with an **empty dependency array**. The `renderCell` closures for the Take Offline / Cancel buttons close over `itemZUID` (from `useParams`).

The language switcher (`LanguageSelector`) preserves the current subpath when navigating, so switching language on `/content/:model/:item/publishings` pushes `/content/:model/:siblingItem/publishings` — the **same** React Router v5 route. React reuses the existing `PublishState` instance and only the route param changes; the component never remounts.

As a result:
- `useGetItemPublishingsQuery` refetches with the new `itemZUID` → grid shows the **new** locale's publishing records (correct `publishingZUID`).
- The memoized column closures still hold the **previous** locale's `itemZUID`.

So Take Offline issued `DELETE /content/models/:model/items/{previousItemZUID}/publishings/{newPublishingZUID}` — a publishing ZUID that doesn't belong to that item → backend 404. This is the "structural mismatch" described in the issue. A hard refresh remounts the component and rebuilds the memo, which is why the workaround succeeds.

## Fix

Key the `columns` memo on `[modelZUID, itemZUID]` so the delete closures always target the active locale's item. `deletePublishing` is a stable RTK Query trigger and `reloadItem` only needs to reflect the current params (which it does whenever the memo rebuilds on a param change), so the memo stays effective and isn't rebuilt on every render.

## Testing

- `tsc --noEmit` clean for the changed file.
- Manual: switch language on the publishings tab and confirm the DELETE now uses the active locale's `itemZUID` and succeeds.

### DIFF
--- a/.github/workflows/claude-auto-reviewer.yml
+++ b/.github/workflows/claude-auto-reviewer.yml
@@ -22,6 +22,7 @@ jobs:
 
       - uses: anthropics/claude-code-action@1dc994ee7a008f0ecc866d9ac23ef036b7229f84 # v1.0.127
         with:
+          use_sticky_comment: true
           github_token: ${{ secrets.GITHUB_TOKEN }}
           anthropic_api_key: ${{ secrets.ANTHROPIC_API_KEY }}
           prompt: |
@@ -42,3 +43,5 @@ jobs:
 
           claude_args: |
             --allowedTools "mcp__github_inline_comment__create_inline_comment,Bash(gh pr comment:*),Bash(gh pr diff:*),Bash(gh pr view:*),Read,Grep,Glob"
+            --model claude-sonnet-4-6
+            --max-turns 20
--- a/cypress/e2e/studio/studio-wrapper.spec.js
+++ b/cypress/e2e/studio/studio-wrapper.spec.js
@@ -745,9 +745,10 @@ describe("Studio Wrapper", () => {
   it("saves all staged content edits via the modal", () => {
     dirtyPageContent();
 
-    cy.intercept("PUT", `/v1/content/models/${modelZUID}/items/${itemZUID}`).as(
-      "saveItem"
-    );
+    // saveItem PUTs to the cross-origin instance API
+    // (https://<zuid>.api.dev.zesty.io/...), so a path-only intercept never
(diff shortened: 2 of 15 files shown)


agalin920:
Closing in favor of #4155. This branch was accidentally based on the studio/staged-multi-edit-saves branch tip and swept in unrelated Studio commits. #4155 contains only the PublishState.tsx fix, based cleanly off dev.

PR #4156, opened 2026-06-17
[Studio] - Add Layers Panel for content and layout modes

agalin920:
## Summary

Adds a **Layers Panel** to Studio — a left-hand tree of the page's element hierarchy, built from the WebEngine bridge's `LAYERS_TREE` message.

- **Content mode:** dynamic field rows (icon by datatype) select for editing; static text and structural containers shown for context.
- **Layout mode:** element rows select on the canvas and drag-reorder; static text rows trigger the existing inline static-editing flow on their parent element.
- **Attribute-bound dynamic content** (e.g. an `` `src`) is surfaced as a content row, with an image icon for img-src bindings.
- Two-way selection sync with the canvas; tree expanded by default.
- Also fixes the side-panel header showing the **page item's** title/version instead of the **selected item's** when one is selected.

## Notes for reviewers

- **Stacked on #4149** (`studio/staged-multi-edit-saves`) — base is that branch so the diff is scoped to the Layers Panel work. GitHub will retarget to `dev` once #4149 merges.
- **`Bridge.js` is intentionally excluded.** The WebEngine bridge runs inside the cross-origin preview iframe and is deployed separately to GCS (`instudio-dev`), not bundled by manager-ui. The host communicates with it only via `postMessage` commands (`LAYERS_TREE`, `moveLayoutElement`, `enterStaticEditingByLayoutId`, etc.). Those bridge changes are already deployed.

## Verification

- Verified end-to-end against a live instance: tree renders (code files, containers, static text, dynamic fields, img-src bindings); content-mode field selection opens the editor; layout-mode element selection sets the breadcrumb + canvas outline; static-row click enters inline editing; selecting another element cleanly exits static editing (no double-highlight).
- `tsc --noEmit` and Prettier clean.
- Reviewed via the repo's auto-reviewer focus areas (quality / bugs / security / performance) — no blocking findings.

### DIFF
--- a/src/apps/content-editor/src/app/views/ItemEdit/components/StudioWrapper/StudioLayersPanel.tsx
+++ b/src/apps/content-editor/src/app/views/ItemEdit/components/StudioWrapper/StudioLayersPanel.tsx
@@ -1,0 +1,190 @@
+import { DragEvent, useCallback, useEffect, useRef, useState } from "react";
+import { Box, Typography } from "@mui/material";
+import { StudioLayersTreeItem } from "./StudioLayersTreeItem";
+import { LayersFlatRow } from "../../hooks/useStudioLayersTree";
+import { LayersDropPosition, LayersTreeNode } from "../../hooks/studioTypes";
+
+// Mirrors the canvas drag behavior: the outer bands of a row mean
+// before/after, the middle means drop inside.
+const EDGE_RATIO = 0.35;
+
+type StudioLayersPanelProps = {
+  hasTree: boolean;
+  flatRows: LayersFlatRow[];
+  selectedNodeId: string | null;
+  onToggle: (nodeId: string) => void;
+  onSelect: (node: LayersTreeNode) => void;
+  canDrop: (
+    sourceId: string,
+    targetId: string,
+    position: LayersDropPosition
+  ) => boolean;
+  onDrop: (
+    sourceId: string,
+    targetId: string,
+    position: LayersDropPosition
+  ) => void;
+};
+
+export const StudioLayersPanel = ({
+  hasTree,
+  flatRows,
+  selectedNodeId,
+  onToggle,
+  onSelect,
+  canDrop,
+  onDrop,
+}: StudioLayersPanelProps) => {
+  const listRef = useRef<HTMLDivElement | null>(null);
+  const dragSourceIdRef = useRef<string | null>(null);
+  const [dropIntent, setDropIntent] = useState<{
+    targetId: string;
+    position: LayersDropPosition;
+  } | null>(null);
+
+  // Keep the canvas selection visible in the tree. Depends on flatRows too so
+  // it retries once the selected row is actually rendered.
+  useEffect(() => {
+    if (!selectedNodeId || !listRef.current) return;
+    const rowEl = listRef.current.querySelector(
+      `[data-node-id="${CSS.escape(selectedNodeId)}"]`
+    );
+    rowEl?.scrollIntoView({ block: "nearest" });
+  }, [selectedNodeId, flatRows]);
+
+  const handleRowDragStart = useCallback(
+    (nodeId: string, evt: DragEvent<HTMLElement>) => {
+      dragSourceIdRef.current = nodeId;
+      evt.stopPropagation();
+      if (evt.dataTransfer) {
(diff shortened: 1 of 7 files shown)


src/apps/content-editor/src/app/views/ItemEdit/components/StudioWrapper/StudioLayersTreeItem.tsx
```diff
@@ -0,0 +1,189 @@
...
+        onDragStart={(evt) => onRowDragStart(node.id, evt)}
+        onDragOver={(evt) => onRowDragOver(node.id, evt)}
+        onDragLeave={() => onRowDragLeave(node.id)}
+        onDrop={(evt) => onRowDrop(node.id, evt)}
+        onDragEnd={onRowDragEnd}
+        display="flex"
+        alignItems="center"
+        gap={1}
+        sx={(theme) => ({
+          p: "4px 8px 4px 4px",
```
finnar-bin: prefer to use theme values

src/apps/content-editor/src/app/views/ItemEdit/components/StudioWrapper/StudioSaveChangesModal.tsx
finnar-bin, line 78: Would this be better upstreamed to the theme?

src/apps/content-editor/src/app/views/ItemEdit/hooks/useStudioLayersTree.ts
```diff
@@ -0,0 +1,389 @@
...
+      ) {
+        return node.id;
+      }
+      if (
+        !fallbackId &&
+        node.fieldZuid === selectedElement.fieldZuid &&
+        (!selectedElement.itemZuid ||
+          node.itemZuid === selectedElement.itemZuid)
+      ) {
+        fallbackId = node.id;
```
finnar-bin, line 146: Why not just return fallbackId here? Doesn't seem like the value gets changed anymore below

finnar-bin (changes requested):

src/apps/content-editor/src/app/views/ItemEdit/hooks/useStudioLayersTree.ts
agalin920: The loop intentionally keeps scanning after setting `fallbackId`.  The exact `studioId + fieldZuid` match in the `return node.id` above must take precedence over the `fieldZuid`-only fallback. The same field can render on multiple nodes (e.g. a title used as a heading *and* as an `` binding), so an exact match found later in iteration should win over an earlier fieldZuid-only hit. `fallbackId` is the defensive last resort used only when no node matches the selected `studioId` and only returned after the full scan. returning it here would pick the wrong occurrence.

src/apps/content-editor/src/app/views/ItemEdit/components/StudioWrapper/StudioLayersTreeItem.tsx
agalin920: updated

src/apps/content-editor/src/app/views/ItemEdit/components/StudioWrapper/StudioLayersPanel.tsx
```diff
@@ -0,0 +1,190 @@
...
+          ? prev
+          : { targetId: nodeId, position }
+      );
+    },
+    [canDrop]
+  );
+
+  const handleRowDragLeave = useCallback((nodeId: string) => {
+    setDropIntent((prev) => (prev?.targetId === nodeId ? null : prev));
+  }, []);
```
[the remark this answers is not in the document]
agalin920: Good catch — fixed in dec898b7c. `handleRowDragLeave` now ignores `dragleave` events whose `relatedTarget` is still inside the row (`evt.currentTarget.contains(evt.relatedTarget)`), so crossing onto a child element no longer clears the drop indicator; it only clears on a genuine row exit.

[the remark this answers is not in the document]
agalin920: resolved

finnar-bin (approved):

PR #4162, opened 2026-06-18
Cypress: stabilize flaky e2e suite

agalin920:
## Summary

Stabilizes the intermittently-failing Cypress e2e suite. The flakiness had two sources: real test-logic races, and the **shared dev instance** being overloaded by concurrent CI runs. This fixes the former and isolates/handles the latter.

## Test-logic fixes
- **workflows**: register the label `PUT` intercepts **before** the click that fires them (they were registered after → `cy.wait` always timed out); assert publish via resulting UI state instead of a post-action intercept.
- **list › Saves bulk edits**: settle on observable UI state (Save button unmounts, toggle reflects refetched value) instead of waiting on an exact count of background refetches.
- **actions**: assert field hydration before clearing a required field; `Unschedules a Publish` now re-visits the item so it no longer depends on the previous test's open modal.
- **redirects**: poll until seeded redirects are queryable before loading the page (eventual consistency).
- **headTags**: hermetic head-tag cleanup; wait only on the head-tags fetch (the other head-page requests don't reliably fire).
- **settings/workflows**: query active labels only (the `showDeleted=true` history had bloated enough to time out the `before` hook).

## Isolation & environment
- **CI concurrency gate**: serialize all e2e CI across PRs, they share one dev instance and concurrent runs overload it (502s/timeouts). One run touches the instance at a time.

## Backend-failure attribution
When a test fails on a backend `5xx` or timeout, it's marked backend-caused; the `backend_health` CI job surfaces a verdict **on the run's summary** (+ a `::warning::` annotation). S

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -34,16 +34,16 @@ jobs:
       - name: Delete Old Screenshots
         run: gsutil rm gs://cypress_screenshots/* || true
 
-  # 5 parallel jobs: runners 0-3 split the non-global specs (SPLIT=4); runner 4
-  # runs the global-state specs (publish/labels/redirects) serially to avoid
-  # cross-runner interference. cypress-split distributes by SPLIT/SPLIT_INDEX.
+  # Single serial runner — whole suite on one runner, no intra-run parallelism.
+  # The shared dev instance 502s under concurrent runner load; serial removes that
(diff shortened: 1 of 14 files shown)


agalin920:
Closing in favor of Warning: 1 uncommitted change
https://github.com/zesty-io/manager-ui/pull/4164 — a clean, single-commit branch off `dev` with the same fixes (parallel runners + concurrency gate, comments trimmed).

PR #4164, opened 2026-06-19
fix(cypress): eliminate e2e flakiness + attribute backend-caused failures

agalin920:
Replaces #4162 with a clean, single-commit branch.

## Test-logic flakes fixed
Each was reproduced locally against a clean build and verified stable:

| Spec | Root cause | Fix |
|---|---|---|
| `content/workflows` | intercept registered after the triggering click | register before |
| `content/list` | racing `cy.wait` on background refetches | assert settled UI state |
| `content/actions` | clearing before hydration; stale modal state | assert hydration; re-visit |
| `content/redirects` | eventual-consistency indexing; content-item redirect left a duplicate → `400` | poll for indexed; clean content-item paths |
| `content/headTags` | leftovers from prior runs | hermetic clean-all before/after |
| `settings/workflows` | `showDeleted=true` cleanup can time out on label bloat | query active labels only |
| `settings/actions` (Typography) | font-family select empty with no installed fonts | use the font-weight select |
| `search/search-bar` | MUI Autocomplete option re-renders, detaching mid-action | re-query + force |

## CI / infrastructure
- **Cross-PR concurrency gate** — all PRs share one dev instance; concurrent runs overload it. Only one e2e run hits it at a time.
- **5-way parallel spec split**; runner 4 groups the global-state specs.
- **Backend 500/timeout attribution** — markers are recorded during tests; on failure the run summary flags whether it was backend-caused, so you can tell at a glance whether to just re-run.
- Generous request/response/command timeouts for the slow shared instance.

## Validation
A full CI run on these fixes went **all green** (5/5 runners, zero backend errors, zero failures).

## Known backend/infra items (not test code)
- Intermittent **`502` gateway outages** on the shared dev instance (e.g. `schema/activity-log` saw all `/v1/*` endpoints `502` at once) — surfaced by the attribution step for escalation.
- **Soft-deleted label bloat** slows `?showDeleted=true`; periodic purge recommended.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -4,6 +4,11 @@ on:
   pull_request:
     branches: [dev, beta, stable, master]
 
+# Serialize e2e across PRs — they share one dev instance; concurrent runs overload it.
+concurrency:
+  group: manager-ui-e2e-dev-instance
+  cancel-in-progress: false
+
 jobs:
   setup:
     runs-on: ubuntu-latest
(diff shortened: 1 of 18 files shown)


agalin920:
Closing in favor of https://github.com/zesty-io/manager-ui/pull/4165 — clean single-commit branch off `dev` with the same changeset (and it drops a stale ConnectToApi.tsx that would have reverted a dev refactor).

PR #4165, opened 2026-06-21
fix(cypress): stabilize flaky e2e suite + cap audits query

agalin920:
Clean single-commit replacement for #4164.

## Root cause of the recurring CI backend outages
`getAudits` fetched **`limit=100000`** — the instance's entire, ever-growing audit history (~**20 MB**) on every activity-log load. Under concurrent CI load this heavy query collapsed the shared dev instance (every runner `502`ing at once). **Capped at 10k (~5 MB, bounded).** This is also a real production perf win.

With that fixed, the full 5× parallel suite **survives and goes green** (validated: a fully-green run, zero backend markers).

## Test-logic flakes fixed (each reproduced locally + verified)
| Spec | Fix |
|---|---|
| `content/workflows` | register PUT intercept before the triggering click |
| `content/list` | assert settled UI state vs racing refetch waits |
| `content/actions` | hydration assert before clearing; re-visit; 30s timeouts |
| `content/redirects` | poll for eventual-consistency; clean content-item redirects |
| `content/headTags` | hermetic clean-all before/after |
| `settings/workflows` | query active labels only (`showDeleted` can time out) |
| `settings/actions` (Typography) | use the font-weight select (font-family is empty) |
| `search/search-bar` | re-query + force autocomplete option interactions |
| `content/content` (one-to-one, repeater) | confirm relational item attached; retries off + settle for the stateful repeater chain |
| `content/meta` (Twitter) | settle inputs + scroll the lazy card image into view |
| `schema/integration` | mock the `get-url` cloud function (live data drifted from the fixture) |

## CI / infrastructure
- Cross-PR **concurrency gate** (shared dev instance).
- 5-way parallel spec split; runner 4 groups the global-state specs.
- **Backend `500`/timeout attribution** surfaced in the run summary (tells you instantly whether a red run was BE-caused).
- Generous request/response/command timeouts.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -4,6 +4,11 @@ on:
   pull_request:
     branches: [dev, beta, stable, master]
 
+# Serialize e2e across PRs — they share one dev instance; concurrent runs overload it.
+concurrency:
+  group: manager-ui-e2e-dev-instance
+  cancel-in-progress: false
+
 jobs:
   setup:
     runs-on: ubuntu-latest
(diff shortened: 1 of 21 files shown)


agalin920:
Closing in favor of https://github.com/zesty-io/manager-ui/pull/4166 — clean single-commit branch with a concise description.

PR #4167, opened 2026-06-22
fix(cypress): stabilize flaky e2e suite + cap audits query

agalin920:
## Why
CI's e2e suite failed intermittently. Root cause of the recurring **backend outages**: `getAudits` fetched `limit=100000` — the instance's entire, ever-growing audit history (~20MB) on each activity-log load — which collapsed the shared dev instance under concurrent CI load. Capped at **10k**.

## What
- **Audits cap** (`100k → 10k`) — eliminates the outages; also a prod perf win.
- **CI structure** — non-global specs run 4-way parallel; the global-state specs (publish/labels/redirects) run alone in a dedicated job afterward, so they never run alongside other state mutators. Cross-PR concurrency gate kept.
- **Backend attribution** — on failure, the run summary flags whether it was backend-caused (5xx/timeout) and lists slow/hanging endpoints, so a red run is triaged instantly.
- **~20 spec fixes** — each reproduced and verified locally (intercept ordering, settled-UI asserts, eventual-consistency polling, deterministic mocks, state cleanup, `.click()` vs `.trigger`, etc.).

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -4,6 +4,11 @@ on:
   pull_request:
     branches: [dev, beta, stable, master]
 
+# Serialize e2e across PRs — they share one dev instance; concurrent runs overload it.
+concurrency:
+  group: manager-ui-e2e-dev-instance
+  cancel-in-progress: false
+
 jobs:
   setup:
     runs-on: ubuntu-latest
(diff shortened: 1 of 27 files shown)


agalin920:
Closing in favor of https://github.com/zesty-io/manager-ui/pull/4168 — same changeset; audits reframed as always-scoped (limit restored to 100000 now that no caller fires unscoped).

PR #4170, opened 2026-06-24
Schema - Fix app crash when adding a field

### LINKED ISSUE #4169: [Schema] App Crashes When Adding a Field to a Model with No Existing Fields
## Description

When a content model has no fields, the "Add Field" button in the header (`data-cy="AddFieldBtn"`) is correctly disabled. However, the only other available option—the "Add Field" button in the empty-state panel—triggers a crash when clicked, regardless of the field type selected. This not only prevents field creation but also breaks CI for any end-to-end test that interacts with the add-field workflow or targets the header button.

Additionally, navigating to a content model with no fields causes the item list table to flicker continuously. The expected "No Fields" empty state never renders.

**Environment:** Development (dev)

## Steps to Reproduce

**Crash:**
1. Navigate to `/schema` and select a model that has no fields.
2. Observe that the header "Add Field" button is greyed out and disabled.
3. Click the "Add Field" button displayed in the empty-state panel.
4. Choose any field type.

**Actual Result:** The application crashes immediately.
**Expected Result:** A modal for creating a new field opens normally.

**Infinite Flicker:**
1. Visit: `https://8-f48cf3a682-7fthvk.manager.dev.zesty.io/content/6-f6a3bfc1f7-pv6jv2?lang=en-US`

**Actual Result:** The item list table flickers indefinitely.
**Expected Result:** The "No Fields" empty state page is displayed.

## Affected Cypress Specs

- `cypress/e2e/schema/field.spec.js`
- Other specs that exercise the add-field flow.

## Recordings:
___
https://github.com/user-attachments/assets/beed56f3-6909-40d7-8fa4-7e90247494a9

https://github.com/user-attachments/assets/4362b1e8-ac41-4be9-ba7c-5267fe23c5e7

geodem127:
## Summary
- Add optional chaining on `fields?.map()` in `FieldFormProvider.tsx` to prevent crash when `fields` is undefined during field addition
- Add `fields` to the `useEffect` dependency array so validation re-runs when fields load
- Add optional chaining on `res.data?.sort()` in `instance.ts` to guard against undefined response data

## Test plan
- [x] Open the Schema app on an instance
- [x] Add a new field to a content model — app should not crash
- [x] Verify field validation still works correctly (duplicate name detection, etc.)
- [x] Verify fields list sorts correctly after fetching

Fixes #4169

### DIFF
--- a/src/apps/schema/src/app/components/contexts/FieldFormProvider.tsx
+++ b/src/apps/schema/src/app/components/contexts/FieldFormProvider.tsx
@@ -195,7 +195,7 @@ export const FieldFormProvider = ({
       return;
     }
 
-    const currFieldNames = fields.map((field) => field.name);
+    const currFieldNames = fields?.map((field) => field.name);
     let newErrorsObj: Errors = {};
 
     Object.keys(formData).map((inputName) => {
@@ -367,7 +367,7 @@ export const FieldFormProvider = ({
     });
 
     setErrors(newErrorsObj);
-  }, [formData, isDefaultValueEnabled]);
+  }, [formData, isDefaultValueEnabled, fields]);
 
   const handleFieldDataChange = useCallback(
     ({ inputName, value }: { inputName: string; value: FormValue }) => {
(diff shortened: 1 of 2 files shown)


agalin920 (dismissed):
@geodem127 what introduced this?

geodem127:
> @geodem127 what introduced this?

@agalin920 Traced it to #4020 (Repeater Field). That PR introduced FieldFormProvider with fields.map(...) and no optional chaining.

PR #4223, opened 2026-07-27
Studio - Add Feedback modal to Studio header

### LINKED ISSUE #4042: Studio - Feedback Modal
## Current behavior

The current studio header does not provide a way for customers to provide feedback

## Proposed change

Let's introduce a feedback modal that can have user feedback sent directly to a studio-feedback channel in Slack. Where we track the user email, instance, the active page & mode that is being edited.

## Design

[A link to the UI design related to the feature request.](https://www.figma.com/design/eksSU0hS6EMriv07WLNexZ/Content.One-Strategy?node-id=528-3405&t=yBHCLGkcwhqgirth-1)

## Additional notes

If a slack channel is a heavy lift, let's send feedback as an email to bobby@zesty.io

geodem127:
## Summary
Adds a "Feedback" entry point to the Studio header that opens a "Share Feedback" modal. Users type freeform feedback, which is submitted via the existing `sendEmail` cloud function to the team's configured Slack feedback channel, tagged with the user's email, instance, active page, and edit mode — per the issue's requirement.

## Changes
- Add `sendEmail` mutation (`{to, from, subject, body, template}`) to `cloudFunctionsApi`
- Add `StudioFeedbackModal` component with copy/layout per approved design reference
- Add "Feedback" button to `StudioHeader`, wire state/handlers in `StudioWrapper`
- Add `SLACK_FEEDBACK_EMAIL` to all four environment configs and its type declaration
- Add Cypress spec covering open/disabled/enabled/cancel/submit/error/escaping behavior

## Changes since initial review
- Addressed review comment: switched the email body from a hand-built HTML string with the message interpolated raw (HTML-injection risk) to the `sendEmail` cloud function's `template: "raw"` mode
- Restored the instance ZUID/name, active page, and interaction mode fields to the email body (the issue explicitly asks to "track the user email, instance, the active page & mode that is being edited") — each field is now on its own `<br>`-separated line with a bold label
- The `raw` template renders HTML (confirmed via live send — bold labels displayed correctly), so every interpolated value (message, email, instance name/ZUID, page, mode) is HTML-escaped before being placed in the body, closing the injection risk that reintroducing HTML formatting would otherwise reopen. A Cypress test asserts a `` payload in the message is escaped rather than sent raw.
- Note (unrelated cloud-function bug, not fixed by this PR): the `sendEmail` cloud function's session check rejects otherwise-valid dev-auth tokens, which blocked live send verification in the local dev sandbox — tracked at zesty-io/gcp-cf#242
- Escaped the email subject line (`feedbackName`/`email`) the same as the body — it was previously left unescaped while the body was fully escaped
- Fixed a Cypress spec bug: the mode-toggle test drove `StudioModeToggleOption-layout` without stubbing the staff flag first, so it always timed out for the non-staff CI account (the toggle only renders for staff, matching the pattern in `studio-mode.spec.js`). Also corrected the plain submit test's default-mode assertion from `content` to `full` — the CI account is entitled to both content and layout, so a fresh page load promotes to full mode, not content (same behavior `studio-mode.spec.js` documents for a staff account).
- Fixed the negative-QA blocker: double-clicking the "Share Feedback" submit button sent two identical emails, because `isSubmitting` (RTK Query's `isLoading`) only updates on the next render — too slow to block a second click fired in the same tick. Added a `useRef` lock set synchronously in `handleSubmit` so the second call is blocked immediately, plus a Cypress case ("sends only one email when the submit button is double-clicked") that fires two native click events back-to-back and asserts exactly one `sendEmail` request. Verified the test actually catches the regression by reverting the fix and confirming it fails (2 requests), then restoring it and confirming green.
- `bobby@content.one` in `SLACK_FEEDBACK_EMAIL` (all four env blocks) is a deliberate secondary recipient, not a leftover debugging address — flagged repeatedly by automated review; documented in [`docs/studio.md`](docs/studio.md#feedback-modal) rather than left as a bare hardcoded value.
- Merged latest `dev` into this branch (resolved one conflict in `StudioHeader.tsx` — kept both the new `Button` import and `dev`'s `useTranslation` import, both of which are used in the file).

## Testing
- Cypress: full spec passing locally (8/8), including the corrected mode-toggle/default-mode assertions and the new double-submit regression test
- `npx tsc --noEmit`: no new errors introduced by this branch
- Manual Playwright walkthrough of the acceptance steps against the local dev server: button placement/copy, disabled/enabled states, brand color, and cancel/reset behavior all verified. Live Slack delivery could not be verified end-to-end from the local dev sandbox due to zesty-io/gcp-cf#242 (dev-auth tokens rejected by the cloud function's session check even though the app's own `/verify` confirms the same token). Recommend a maintainer with stage/production access do one live click-through before merge to confirm delivery and bold/line-break rendering in the destination Slack channel.

## Related
Refs #4042

---


geodem127:
This is currently blocked, waiting for cloud function updates.

agalin920:
Failing QA Review

finnar-bin (dismissed):

cypress/e2e/studio/studio-feedback-modal.spec.js
```diff
@@ -0,0 +1,180 @@
+describe("Studio Feedback Modal", () => {
+  let studioPath = "/";
+
+  before(() => {
+    cy.task("seed:content", "fixtures/studio.json").then(({ items }) => {
+      studioPath = `/${items[0].web.pathPart}`;
+    });
+  });
+
+  const openFeedbackModal = () => {
```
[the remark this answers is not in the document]
finnar-bin: @geodem127 might be worth looking at adding a clean up like what's suggested here

src/apps/studio/components/StudioFeedbackModal.tsx
```diff
@@ -0,0 +1,156 @@
...
+        instanceName
+      )})`,
+      `<b>Page:</b> ${escapeHtml(activePage)}`,
+      `<b>Mode:</b> ${escapeHtml(mode)}`,
+      `<b>Message:</b><br>${escapeHtml(message).replace(/\n/g, "<br>")}`,
+    ].join("<br>");
+
+    sendEmail({
+      to: CONFIG.SLACK_FEEDBACK_EMAIL,
+      from: email,
```
agalin920: Im concerned with us using the customer domains from address will slack gateway not potentially drop it or spam file it? why not use our default for from like in WorkflowRequest since the address is already in the body on line 69

cypress/e2e/studio/studio-feedback-modal.spec.js
```diff
@@ -0,0 +1,180 @@
+describe("Studio Feedback Modal", () => {
+  let studioPath = "/";
+
+  before(() => {
+    cy.task("seed:content", "fixtures/studio.json").then(({ items }) => {
```
agalin920: this seeded model never gets deleted is this intended?

agalin920:
@geodem127 please make sure negative QA runs against this

src/apps/studio/components/StudioFeedbackModal.tsx
geodem127: Fixed — dropped `from`, sendEmail now uses the default sender.

cypress/e2e/studio/studio-feedback-modal.spec.js
geodem127: Fixed — added `after()` to delete the seeded model.

geodem127:
Negative QA manually triggered for the latest commit.

finnar-bin:
@geodem127 please address the negative QA blocker: https://github.com/zesty-io/manager-ui/pull/4223#issuecomment-5347673396

geodem127:
> @geodem127 please address the negative QA blocker: [#4223 (comment)](https://github.com/zesty-io/manager-ui/pull/4223#issuecomment-5347673396)

- done. updated prop type

Note: the second click in the reported double-click sequence is landing on the `MuiBackdrop`, which by default MUI behaviour closes the dialog via `onClose`. This is normal MUI `Backdrop`/`Dialog` behaviour, not a regression specific to this PR.

finnar-bin:
> > @geodem127 please address the negative QA blocker: [#4223 (comment)](https://github.com/zesty-io/manager-ui/pull/4223#issuecomment-5347673396)
> 
> * done. updated prop type
> 
> Note: the second click in the reported double-click sequence is landing on the `MuiBackdrop`, which by default MUI behaviour closes the dialog via `onClose`. This is normal MUI `Backdrop`/`Dialog` behaviour, not a regression specific to this PR.

@geodem127 the negative QA noted that it caught 2 sendEmail api calls, just wanted to make sure that you've verified that this isn't the case?

geodem127:
Re: https://github.com/zesty-io/manager-ui/pull/4223#issuecomment-5671995829

@finnar-bin good catch, you were right — my earlier reply answered the wrong bug. This one was real: double-clicking submit fired two `sendEmail` requests since `isSubmitting` didn't update fast enough to block the second click.

Fixed with a synchronous lock in `handleSubmit`, plus a Cypress test reproducing the double-click.

agalin920 (approved):

agalin920 (approved):

PR #4228, opened 2026-07-30
CI - Add a negative-QA workflow that drives the app with Playwright MCP

agalin920:
## What this adds

A fourth Claude job on PRs into `dev`. The three existing ones review a PR **from the code alone**; Cypress covers the happy paths, and its specs are written by whoever wrote the feature, so they encode the author's assumptions. Nothing in CI actually *misuses* the app.

This is the dynamic counterpart to Claude Change Verifier. It boots the dev server, seeds throwaway content, and lets a senior-QA persona attack the surfaces the diff touches with the inputs nobody thought about — empty and whitespace-only values, markup in text fields, double-submits, cancel mid-save, reload with unsaved changes, deleting a dependency out from under a selection.

Findings post as an upserted PR comment (marker `<!-- negative-qa -->`) with screenshots as evidence.

**Advisory only — it always exits 0.** Agentic exploration is non-deterministic, and a false positive blocking a merge would kill trust in it within a week. The report format carries severity levels so gating on high-severity later is a small change.

## How auth works

This is the part worth reviewing closely.

`utility/request.js` reads the session cookie with js-cookie and sends it as an `Authorization: Bearer` header with `credentials: "omit"`. So a single JS-readable `DEV_APP_SID` cookie is enough for the boot-time `verify()` to succeed — Cypress's second httpOnly `.zesty.io` cookie is incidental.

`ci/scripts/qa_seed.mjs` writes that one cookie into a Playwright **storage-state** file, and the MCP server loads it via `--storage-state`. The agent is handed an already-authenticated browser and simply navigates. **The token never enters the agent's prompt or transcript.**

`ci/scripts/qa_verify_session.mjs` then asserts `auth.valid` before the agent starts. Without it, a broken session produces an empty report that is indistinguishable from a clean PR — an expensive, silent failure. It caught a real problem the first time it ran.

## Design notes

- **Concurrency:** joins `ci.yaml`'s `manager-ui-e2e-dev-instance` group, since both mutate the same shared dev instance. Known consequence: GitHub keeps only one *pending* run per group, so queued runs can be cancelled on a busy day. Switching to a dedicated group is a one-line change if that bites.
- **Test data:** seeds its own model/item labelled `QA: … | <sha> | <uuid>`, mutates only those, and deletes them in an `if: always()` step. Several Cypress specs assert on specific live records, so the agent is told explicitly not to touch anything else.
- **Playwright is installed with `--no-save`** so it never lands in `package-lock.json` and the other CI jobs don't pay for it on every `npm install`. Installing it first means `npx playwright install` resolves the exact build `@playwright/mcp` pins.
- **`@playwright/mcp` is version-pinned**, not `@latest` — `--storage-state` is load-bearing, and a silent flag rename would look like "the whole app is behind a login screen" rather than a bug.
- **`--test-id-attribute data-cy`** so the agent's element refs line up with this repo's selector convention.
- `browser_run_code_unsafe` is deliberately **not** in the allowlist.

## Also in this PR

`Meta/index.tsx`: `REQUIRED_FIELDS` was a `useMemo` returning a constant array, keyed on a `[model]` dependency it never read, with an inner `fields` binding shadowing the outer one two memos above. Hoisted to module scope. No behaviour change.

## Verification

Everything below was run against the real dev instance:

- Seed → session check → cleanup chain, repeatedly.
- The MCP server driven over real JSON-RPC with the exact flags this workflow uses → `auth.valid: true`.
- The whitespace bug reproduced through the UI, with the stored value read back from the API.
- Workflow YAML parses; shell and Python scripts syntax-checked; the screenshot link rewriter unit-tested including the unresolved-token fallback.

**Unverified:** whether `gs://cypress_screenshots` allows public object reads. The upload script tries public URLs, falls back to `gsutil signurl -d 7d`, and degrades to an artifact reference — but which path it takes on the first real run is unknown. Check with:

```
gsutil ls -L -b gs://cypress_screenshots | grep -Ei 'Public|uniform|ACL'
```

Screenshots go under a `negative-qa/pr-<N>/<run-id>/` prefix so `ci.yaml`'s `gsutil rm gs://cypress_screenshots/*` can't reach them (gsutil's `*` doesn't cross `/`). A run artifact is uploaded regardless, since that guarantee is subtle.

The workflow should run on this PR and comment its own findings — that's the demo.


agalin920:
Superseded by #4235, which has the same content on clean history (this branch carried an unrelated refactor plus its revert). Closing rather than force-pushing.

The negative-QA comment above stays here as the record of the workflow finding both bugs on a real diff.

PR #4253, opened 2026-08-09
Studio - Relocate the implementation into src/apps/studio and document it

agalin920:
Moves the Studio implementation into `src/apps/studio/`, where the routed app already lived, and adds `docs/studio.md` documenting the result.

The implementation — 21 files, ~7,900 lines — sat under `src/apps/content-editor/src/app/views/ItemEdit/`, so the dependency arrow ran backwards: `src/apps/studio/` was a route shim importing its own implementation out of a sibling app's view directory.

**A move and nothing else** — no renames, no logic changes, no cleanups. Every file `git mv`'d, recorded as renames at 92–100% similarity, so `--follow` and blame survive.


geodem127 (approved):

finnar-bin (approved):

agalin920:
Superseded by #4257.

The commits on this branch predate commit signing on our side, and the `dev` ruleset requires signed commits — so this PR could never satisfy it, which is why auto-merge sat armed and never fired. Everything else was green: approved, mergeable, checks passing, no open threads.

#4257 is the same seven commits replayed onto `dev` and signed. Its tree is byte-identical to `a16c58be`, so nothing about the change itself has moved. No force-push and no rewrite — `docs/studio-architecture` stays as it is, since PR #4256 is stacked on it.

PR #4282, opened 2026-08-21
Build - Raise the dev and stage deploy runners to Node 22.19.0 and bump copy-webpack-plugin to 14

### LINKED ISSUE #4270: Build - Raise the Node version pinned by the dev and stage deploy workflows off 16.5.0
## Problem

Two deploy workflows pin an eight-year-old Node:

- `.github/workflows/cd-dev.yaml:32` — `node-version: "16.5.0"`, then bare `npm install` at `:36`
- `.github/workflows/cd-stage.yaml:40` — same, `npm install` at `:44`

`ci.yaml` is on `22.19.0` and is unaffected, so this only bites the dev and stage deploy path.

Node 16.5.0 bundles npm 7.19.x. That has two consequences that are not obvious from reading the workflow:

1. **`overrides` in `package.json` are silently ignored.** npm did not support `overrides` until 8.3.0. Any override the repo adds is honoured by CI and by local installs, and quietly dropped on the deploy runners, which then re-resolve those dependencies to whatever the parent ranges allow. The committed `package-lock.json` is unchanged either way, so Dependabot still reports the alerts as fixed — the deploy artifact just is not built with the pinned versions. There is no warning anywhere.
2. **It blocks the last open critical/high Dependabot alert.** Alert #165 (`serialize-javascript`, GHSA-5c6j-r48x-rmvq, CVSS 8.1) is reached through `copy-webpack-plugin`, which is `require`d at `src/shell/webpack.config.js:10` and executed at `:193` — so it runs *during the build*, on these runners. The only release that moves to the patched `serialize-javascript ^7.0.3` is `copy-webpack-plugin@14`, which declares `engines.node >= 20.9.0`. There is no `.npmrc`, so `engines` is advisory: npm would install it and then execute Node-20 code on Node 16 at build time.

Related: `npm install` on these runners is bare — no `--legacy-peer-deps`, unlike `ci.yaml` — so the deploy tree is not resolved the same way CI's is.

## Proposed change

Raise `node-version` on both workflows to match the version the repo already pins for every other context: **22.19.0** (`package.json` `volta.node`, and what `ci.yaml` uses). Consider aligning the install command with `ci.yaml`'s `npm install --legacy-peer-deps` at the same time.

This needs someone with `workflow` scope — it cannot be pushed from a token without it.

## Acceptance criteria

- [ ] `cd-dev.yaml` and `cd-stage.yaml` no longer pin a Node major below the one in `package.json`'s `volta.node`.
- [ ] A dev deploy and a stage deploy both complete successfully after the change.
- [ ] `copy-webpack-plugin` is bumped to a release that depends on `serialize-javascript ^7.0.3`, and `npm run build:stage` succeeds on the new runner Node.
- [ ] `gh api "repos/zesty-io/manager-ui/dependabot/alerts?state=open" ` no longer returns alert #165.

## Notes

- Verified that nothing currently in the tree requires more than Node 16: `webpack@5.109.2` declares `engines.node >=10.13.0`, the same as the 5.99.8 it replaced. So this is not urgent as a breakage — it is a ceiling on what can be remediated, plus a silent divergence between what CI builds and what deploys.
- The `overrides` half of this matters even though the repo has no `overrides` key today: it is the reason one was not a viable remediation route.

agalin920:
Closes #4270

`cd-dev.yaml` and `cd-stage.yaml` pinned `node-version: "16.5.0"` and ran a bare `npm install`, while `ci.yaml` and `package.json`'s `volta.node` are both on `22.19.0`. Both now use `22.19.0` and `npm install --legacy-peer-deps`, so the deploy tree resolves the way CI's does. That unblocks `copy-webpack-plugin@14` (`engines.node >= 20.9.0`), bumped here from `^9.0.0`: it is the only release depending on `serialize-javascript ^7.0.3`, and `copy-webpack-plugin` is the sole edge to `serialize-javascript` in the lockfile, so it is the only route to closing Dependabot alert #165 (GHSA-5c6j-r48x-rmvq, CVSS 8.1). To be explicit about the ordering, since the two changes ride together: `copy-webpack-plugin@9` does not fail on Node 22 — the Node pin was blocking the plugin bump, not the reverse, and the plugin bump exists to close the alert rather than to fix a build.

Facts the acceptance criteria turn on:

- **`--legacy-peer-deps` is required, not cosmetic alignment.** On Node 22 a plain `npm install` exits 1 with `ERESOLVE` (`peer react@"^16.3.0" from @aeaton/react-prosemirror@0.21.1` against React 18). Raising `node-version` without also adding the flag would have broken the Install step on both deploy workflows.
- **`npm run build:dev` and `npm run build:stage` both exit 0** with the bumped plugin, on Node 22 (22.21.1/npm 10.9.4 locally; the runners get 22.19.0/npm 10.9.3). `build:stage` matters on its own because it is the only `mode: "production"` path — terser and contenthash — and no CI job runs it.
- **The build artifact is unchanged.** All **551** files copied out of `public/` are hash-identical before and after, in both `build:dev` and `build:stage`, with none missing or added. `public/` has no dotfiles, symlinks or empty directories, so that check is exhaustive for the `globby`→`tinyglobby` engine swap in the plugin's v13. Remaining output differences are the webpack bundles and `release.json`, which differ between any two builds because `etc/release.js` stamps build metadata via `DefinePlugin`.
- `package-lock.json` stays at `lockfileVersion: 2` and the diff is confined to the `copy-webpack-plugin` subtree. `npm audit` goes 30 → 28, high 7 → 6.

### `skipLibCheck` and the CONFIG types

`copy-webpack-plugin@14`'s declarations reach `tinyglobby` → `fdir` → untyped `picomatch`, which added a `TS7016` in `node_modules` and took `tsc --noEmit` from 20 errors to 21. Rather than shim that one package, this enables `skipLibCheck: true` — every error in that baseline is a `node_modules` declaration error nobody acts on, and the count is now **0**, which retires the re-measure-the-baseline step entirely.

`skipLibCheck` stops TypeScript checking the *inside* of every `.d.ts`, including this repo's own. For ambient declarations that is safe: they fail loudly at their use sites. For a **derived** type it is not — the derivation errors only inside the declaration file, then degrades to something that behaves as `any` while every consumer keeps compiling. `src/globals.d.ts` had two such derivations (`typeof import("shell/app.config")["production"]` and `ReturnType<typeof import("utility/getRuntimeEnv").default>`), so enabling the flag alone would have silently untyped the whole `CONFIG` / `__CONFIG__` global that #4271 added.

Measured, not assumed: with `production` renamed in `src/shell/app.config.js`, the flag on and the aliases still in the `.d.ts`, `tsc` reported **0 errors** and `CONFIG.ANY_MADE_UP_KEY` type-checked clean. So the derived aliases moved to `src/shell/configTypes.ts`, a real module that stays checked — with the same edit applied there, `tsc` correctly reports `TS2339`. `src/globals.d.ts` keeps only the ambient `declare` blocks, and `CLAUDE.md`'s TypeScript section records the rule so the aliases do not drift back.


src/shell/configTypes.ts
finnar-bin, line 1: Why is this being moved?

agalin920: @finnar-bin We're turning on `skipLibCheck` in this PR, which means TS stops checking `.d.ts` files entirely. Those `CONFIG` types are derived from `app.config.js`, so if that ever broke we'd hear nothing — `CONFIG` just quietly becomes `any` everywhere and `tsc` still passes. Moving them to a real `.ts` keeps them checked.

finnar-bin (approved):

Issue #4264, opened 2026-08-10
Manager UI - Replace hardcoded colour and size literals with design system tokens (TS/JS)


agalin920:
## Problem

`src/**/*.{ts,tsx,js,jsx}` contains **176 hardcoded hex-literal occurrences across 57 files**, plus **27 raw `fontSize` px values** and **34 raw px spacing values** in `sx` props. Most are not new colours — they are literal re-typings of values the design system already exports. `#0BA5EC` is `info.main`. `#FF5D0A` is `primary.main`. `#101828` is `text.primary`. `#475467` is `text.secondary`.

The cost is not aesthetic. A literal does not follow the theme, so any palette change silently skips these 57 files, and a mis-typed literal is invisible to review. #4256 already produced a live instance of this class: a button shipped slate grey against an orange design because the colour came from somewhere other than the token.

## Depends on #4263

**Do not start this before #4263 lands.** That ticket produces `docs/design-system.md`, which settles two things this sweep cannot proceed without:

1. **Which of the two palettes in `@zesty-io/material` is canonical.** The package exports both a current-looking set (`#FF5D0A`, `#0BA5EC`, `#12B76A`) and a legacy-looking one (`#e53c05`, `#f17829`, `#497edf`, `#75bf25`, `#404759`) as `main` values. Sweeping a literal onto the wrong generation cements it.
2. **The rules for what must *not* be swept.** Roughly a third of the distinct values should stay literals — third-party brand colours (Twitter `#1DA0F0`, Instagram `#6727BB`, Google `#131CA4`) and the Studio canvas CSS, which is injected as a raw string into the customer's iframe and cannot read the theme. #4263 defines these as rules with a home to import from, so this sweep has a mechanical stopping condition rather than a list to memorise.

## Scope

**In:** `src/**/*.{ts,tsx,js,jsx}`.

**Out:** the 82 `.less` files and the 94 further hex literals inside them — tracked in #4265. Keeping them out is what makes this ticket reviewable.

Of the 73 distinct hex values in scope, **42 already exist verbatim in `@zesty-io/material`**. Those are the mechanical, low-argument core of the sweep and should land first, as their own commit. The remaining 31 are handled per the rules in `docs/design-system.md` and should be a separate commit, or deferred entirely if those rules say to leave them.

[...]

PR #4285, opened 2026-08-24
Manager UI - Collapse the second editor black onto background.editor

agalin920:
Two code-editor surfaces described the editor background as the literal `#0D1116` while `src/shell/index.js:87` declares `background.editor` as `#0F0F0F` and eight other call sites consume it.

- `BottomDrawer/FileCard.tsx:34` — carried `// Temporary: will be updated once the design is final`
- `RecentFiles/index.tsx:83` — sits inside a parent `Box` two elements up that already uses the token

Both now consume `background.editor`. `docs/design-system.md` §3 (#4283) is the rule: *being the odd one out is not drift — the question is whether the design system claims to own the value.* It owns the editor background.

**This is a rendered-value change, which is why it is not part of #4284.** `#0D1116` → `#0F0F0F` is a shift of 2/2/7 out of 255 on a near-black, on two panels already sitting on token-painted backgrounds. #4264's acceptance criteria require every substitution to be value-identical, so this one does not belong in that sweep even though it came out of the same audit. Split out on that basis.

Requested by @agalin920 after the audit surfaced the two disagreeing values.

`tsc --noEmit` at baseline: 20 errors, 0 in `src/`.

### DIFF
--- a/src/apps/code-editor/src/app/components/BottomDrawer/FileCard.tsx
+++ b/src/apps/code-editor/src/app/components/BottomDrawer/FileCard.tsx
@@ -30,8 +30,7 @@ const FileCard: React.FC<FileCardProps> = ({
     <Card
       sx={{
         boxSizing: "border-box",
-        // Temporary: will be updated once the design is final
-        bgcolor: "#0D1116",
+        bgcolor: "background.editor",
         color: "grey.400",
         width: "100%",
         px: 2,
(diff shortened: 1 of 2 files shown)


agalin920:
Folding this back into #4284 at @agalin920's direction — one PR rather than two. The change is unaltered; it is now commit 5 there.

Issue #4254, opened 2026-08-09
Upgrade monaco-editor from 0.25.2 — fixes the CSS worker null-model crash (MANAGER-UI-2EM)


agalin920:
## Summary

`monaco-editor` is pinned at **0.25.2** (released 2021). Current upstream is **0.56.0**. Upgrading closes at least one live Sentry issue and removes a workaround we would otherwise have to carry indefinitely.

## Why now

Sentry **MANAGER-UI-2EM** — `TypeError: Cannot read properties of null (reading 'languageId')`, 4 users / 25 occurrences, first seen 2025-08-01 — is a crash inside the vendored CSS web worker, not in our code.

The mechanism: opening a `.css`/`.less`/`.scss` file registers worker-backed folding and word-highlight providers. Both are debounced (200ms / 250ms). Switching files remounts the editor and disposes the model, but a request already posted to the worker still gets answered. `CSSWorker._getTextDocument` then returns `null`, and in 0.25.2 only `doValidation` guards against that — `getFoldingRanges`, `findDocumentHighlights`, `doComplete`, `doHover` and the rest pass `null` straight into `vscode-css-languageservice`, which throws on `.languageId`.

**Upstream fixed this in 0.31.0** (published 2021-12-10), in commit [`4f03f6c4`](https://github.com/microsoft/monaco-editor/commit/4f03f6c4bf64e7cb997855aab24c4db19a142923) — it adds an `if (!document) return []` early return to every method in that file, including both of the ones we hit. The guard is still present on `main` today.

Worth flagging for anyone who searches release notes: **this is not listed as a bugfix.** It rode along inside a commit titled "Enable `strict` typescript compilation", which is why the crash reads as an open upstream defect when it has actually been fixed since 2021 — roughly three and a half years before our first occurrence.

## Scope

The version bump is the easy part. The surrounding work:

- **`monaco-editor-webpack-plugin@4.2.0` declares peers `0.25.x || 0.26.x || 0.27.x || 0.28.x`** — it cannot reach 0.31.0, so the plugin has to move too. Pick the plugin version first; it constrains the achievable monaco target.
- **Known breaking changes between 0.25 and 0.31** (there will be more above 0.31): `monaco.KeyCode` enum renames, `getModeId()` → `getLanguageId()`, `IPasteEvent.mode` → `IPasteEvent.languageId`.
- `react-monaco-editor@0.47.0` needs its own compatibility check against the chosen monaco version.
[...]

PR #4301, opened 2026-08-30
Code - Upgrade monaco-editor to 0.31.1 for the null-model worker guard

### LINKED ISSUE #3895: Error: Cannot read properties of null (reading 'getText')
Sentry Issue: [MANAGER-UI-2F2](https://zestyio.sentry.io/issues/6822495241/?referrer=github_integration)

```
Error: Cannot read properties of null (reading 'getText')

TypeError: Cannot read properties of null (reading 'getText')
    at Object.Xi [as getFoldingRanges] (https://8-98d5d9c3f2-7bfzd9.manager.zesty.io/json.worker.js:1:228877)
    at e.<anonymous> (http...
  at new Promise (<anonymous>)
...
(7 additional frame(s) were not displayed)
```

### LINKED ISSUE #3909: Error: Cannot read properties of null (reading 'languageId')

agalin920:
Closes #3895

`monaco-editor` 0.25.2's `JSONWorker.getFoldingRanges` passes the result of `_getTextDocument(uri)` straight into the JSON language service without checking it. When a folding request has been posted to the worker and the model is disposed before the worker answers — closing the `JsonViewer` dialog, or switching files in the Code app, both of which dispose the model — that lookup returns `null`, the language service calls `document.getText()`, and the rejection reaches Sentry through `onUnexpectedError`. That is the reported frame. Upstream added `if (!document) { return []; }` to the method in **0.31.0**; 0.30.1 is the last unguarded release. This bumps `monaco-editor` to ^0.31.1, which carries the same guard for the CSS worker's `getFoldingRanges` and `findDocumentHighlights` (the sibling crash in the closed #3909), and it is the substance of #4254. `monaco-editor-webpack-plugin` moves 4.2.0 → 7.1.1 because 4.x peers only 0.25–0.28; `react-monaco-editor` stays at 0.47.0, and its peer on `monaco-editor@^0.31.1` — which `npm ls` currently reports as unsatisfied — becomes satisfied. No application code changes: none of the twelve monaco symbols used in `src/` is in the 0.25→0.31 breaking set, and the three feature names in the webpack allowlist that 0.31.1 no longer knows (`colorDetector`, `goToDefinitionCommands`, `goToDefinitionMouse`) are already dead against the 4.2.0 table today, so nothing changes there either. The guard is present in the emitted artifact, not just in `node_modules` — `build/json.worker.js:19456`. The whole dependency delta is these two packages; no transitive version moves across the other 1,528. The dev bundle gets **smaller**: 75.1 MB → 69.9 MB (−5.1 MB, −6.9%), mostly `ts.worker.js` and `vendors.js`.

No `reviewer` cold pass ran on this diff — this session's harness ruled out subagents, so the review gate in `docs/definition-of-done.md` §5 was not met and the diff has been read only by the context that wrote it.


agalin920:
On ⚠️ 3 — the consumer regression check was run locally, it just isn't visible from the diff. Against this branch with the bump installed:

- `npm run build:dev` → `webpack 5.109.2 compiled successfully`, all five language workers emitted.
- `cypress/e2e/code/actions.spec.js` 1/1 — this is the `MonacoDiffEditor`/`Differ` surface.
- `cypress/e2e/code/sidebar.spec.js` 1/1 and `cypress/e2e/code/all-files-page.spec.js` 8/8 — the Code app editor.
- `cypress/e2e/schema/integration.spec.js` 40/40 — this **is** the `JsonViewer.tsx` surface. `integration.spec.js:309` (`View Item's JSON data`) opens the dialog, asserts on `.monaco-editor.integrationJsonViewerEditor`, and then clicks `jsonCodeViewerCloseButton`, which is the unmount-and-dispose path this guard protects. So the "no existing spec for `JsonViewer.tsx`" note in the suggestion is not right — there is one, and it passes.
- `tsc --noEmit` exits 0 both before and after the bump.

Also worth recording since it is the other half of "does the 0.25→0.31 jump break a consumer": every monaco symbol referenced anywhere in `src/` is `monaco.Selection`, `monaco.Uri.from`, `monaco.editor.{createModel,defineTheme,getModel}`, and `monaco.languages.{register,registerCompletionItemProvider,setLanguageConfiguration,setMonarchTokensProvider,CompletionItemKind.Property,IndentAction.*}`. None of them is in the 0.25→0.31 breaking set — no `getModeId`, no `KeyCode`/`KeyMod`, no `IPasteEvent.mode`.

On the suggested race spec: declining it deliberately, same reasoning that closed #3909. The race has never reproduced on demand — it has no confirmed reproduction from the reporter, from Sentry, or from us — so a spec that rapidly switches files and asserts no error appears would pass identically with this bump reverted, which makes it a spec that cannot fail rather than coverage. The falsifiable check is on the artifact instead, and it was made: `build/json.worker.js:19456` reads `let document = this._getTextDocument(uri); if (!document) { return []; }`, and the same bytes were served over HTTP from the dev server. The unguarded call site is gone from what ships.

agalin920:
Confirmed and fixed in `f4c822408`. This was a real regression from the bump and the diagnosis above is correct down to the call site — thank you.

**Root cause, exactly as reported.** monaco 0.31 added this to `StandaloneTheme`'s `tokenTheme` getter (`standaloneThemeServiceImpl`, "Pick up default colors from `editor.foreground` and `editor.background` if available"):

```js
const editorForeground = this.themeData.colors['editor.foreground'];
```

Unguarded. `MonacoSetup.js` calls `defineTheme("parsleyDark", { base, inherit, rules })` with **no `colors` key**, so `themeData.colors` is `undefined` and the read throws. 0.25.2's `tokenTheme` getter has no `themeData.colors` access at all, which is why this only appears now.

Worth recording: **`colors` is a required member of `IStandaloneThemeData` in 0.25.2 as well** — the call has always violated the type, monaco just never read the field. It went unnoticed because `MonacoSetup.js` is `.js`, so `tsc` never type-checks it.

**Fix** is `colors: {}`, which is behaviour-preserving rather than a papering-over: both lookups stay `undefined`, no default token rule is synthesised, and tokens render exactly as on 0.25.2.

**On why the existing suite stayed green — that is the more useful half of this finding.** No spec asserted the Code editor renders. `code/sidebar.spec.js` clicks a tree item and asserts `location.pathname`, which is unchanged by a crash inside the editor pane's `ErrorBoundary`; `code/actions.spec.js` exercises `MonacoDiffEditor`, a different mount path that never runs `editorDidMount`'s `updateOptions({theme})`. So all 50 tests I ran passed through a completely broken Code app. `cypress/e2e/code/editor.spec.js` is added in the same commit and asserts the editor actually mounts — verified falsifiable: it fails with `Expected to find element: .react-monaco-editor-container, but never found it` on the previous commit and passes on this one.

Re-verified locally on `f4c822408`: `code/editor` 1/1, `code/actions` 1/1, `code/sidebar` 1/1, `code/all-files-page` 8/8, `schema/integration` 40/40, `tsc --noEmit` exit 0, `build:dev` compiled successfully.

finnar-bin (approved):

PR #4313, opened 2026-09-01
[CI] - Stop pre-checking Sentry auto-fix PR reviewers via collaborators endpoint

### LINKED ISSUE #4298: CI - Sentry auto-fix PR reviewer check false-flags team-access reviewers as having no access
## Summary

`claude-sentry-handler.yml`'s auto-fix PR step incorrectly reports that `agalin920`, `geodem127`, and `finnar-bin` "no longer have access" to the repo and skips requesting them as reviewers. This is a false positive caused by a `GITHUB_TOKEN` limitation, not an actual loss of access — all three still have push access to the repo.

## Where it happens

Workflow run: https://github.com/zesty-io/manager-ui/actions/runs/33036625238/job/98400587573 (`dev Claude: Sentry Handler`, job `sentry-rca`)

Annotations from that run:

```
! Reviewer 'finnar-bin' no longer has access to zesty-io/manager-ui — skipping.
! Reviewer 'geodem127' no longer has access to zesty-io/manager-ui — skipping.
! Reviewer 'agalin920' no longer has access to zesty-io/manager-ui — skipping.
```

Source: `.github/workflows/claude-sentry-handler.yml`, "Open auto-fix PR" step, lines ~554-568:

```bash
candidate_reviewers=(agalin920 geodem127 finnar-bin)
valid_reviewers=()
for r in "${candidate_reviewers[@]}"; do
  if gh api "repos/$REPO/collaborators/$r" >/dev/null 2>&1; then
    valid_reviewers+=("$r")
  else
    echo "::warning::Reviewer '$r' no longer has access to $REPO — skipping."
  fi
done
```

`GH_TOKEN` for this step is the job's `GITHUB_TOKEN`, and the job's `permissions:` block only grants `contents: write`, `pull-requests: write`, `issues: write` (no org-level scopes).

## Root cause

All three usernames have repo access **via org team membership**, not as direct/outside collaborators:

```
$ gh api "repos/zesty-io/manager-ui/collaborators?affiliation=direct" --jq '.[].login'
(empty)

$ gh api "repos/zesty-io/manager-ui/collaborators?affiliation=all" --jq '.[].login' | grep -E "agalin920|geodem127|finnar-bin"
agalin920
finnar-bin
geodem127
```

`GET /repos/{owner}/{repo}/collaborators/{username}` needs org/team-membership visibility to resolve access that comes through a team rather than a direct collaborator grant. The default Actions `GITHUB_TOKEN` doesn't carry that org-level visibility (it can't be granted via the workflow's `permissions:` block — there's no `organization`/`members` scope available to `GITHUB_TOKEN`), so the call 404s for all three of us even though we do have push access.

Confirmed by calling the same endpoint with an authenticated user token (which does have org visibility):

```
$ gh api repos/zesty-io/manager-ui/collaborators/agalin920/permission
{"permission":"write", ..., "role_name":"write"}
```

That returns `write` access — so the users are not stale, the check itself is blind to team-granted access under `GITHUB_TOKEN`.

## Impact

Auto-fix PRs opened by the Sentry handler workflow never get any of the three intended reviewers requested, silently. The workflow doesn't fail (by design — see the comment above the check), so this has likely been happening on every "simple" auto-fix PR since the check was introduced without being noticed.

## Suggested fix

Stop pre-checking eligibility via `repos/{owner}/{repo}/collaborators/{username}` (unreliable under `GITHUB_TOKEN` for team-based access). Instead:

1. Create the PR without `--reviewer`.
2. Loop over the candidate reviewers afterward and attempt `gh pr edit "$pr_number" --add-reviewer "$r" || echo "::warning::Could not add reviewer $r"` individually, so one bad/stale username degrades gracefully instead of relying on a pre-check that has false negatives.

This preserves the original intent (don't let a stale reviewer block PR creation) without depending on an API call that can't see team-granted access.

## Related

- Introduced in #4269 (CI - Add automated Sentry issue root-cause analysis workflow)

finnar-bin:
Fixes #4298

## Summary
- The "Open auto-fix PR" step in `.github/workflows/claude-sentry-handler.yml` used to pre-check each candidate reviewer (`agalin920`, `geodem127`, `finnar-bin`) via `gh api repos/{owner}/{repo}/collaborators/{username}` before calling `gh pr create --reviewer`.
- That endpoint can't see access granted via org team membership under the default `GITHUB_TOKEN`, so it 404'd for all three despite them having real write access through team membership — the pre-check treated them as having no access and silently skipped every reviewer request on every auto-fix PR.
- Fix: create the PR first without `--reviewer`, then request each candidate individually by POSTing directly to `repos/{owner}/{repo}/pulls/{n}/requested_reviewers`.
  - **Not** `gh pr edit --add-reviewer`: confirmed live against this PR that its GraphQL query pulls the deprecated Projects (classic) `projectCards` field, which errors out on *every* invocation in this repo regardless of the reviewer's validity — using it would have warned on every single reviewer, every time, masking valid ones just like the original bug.
  - The REST endpoint has its own quirk: it returns HTTP 200 and silently drops an unknown/nonexistent login instead of erroring, so exit code alone can't detect a bad login.
  - The warning now surfaces the actual reason instead of a generic message: GitHub's `.message` for a real access error (e.g. "not a collaborator"), or an explicit note when the login was silently dropped from a 200 response with no reviewer added.
- Verified against three real response shapes by testing live against this PR (cleaning up each test review request immediately after):
  - valid login, has access → added, appears in `requested_reviewers`, no warning
  - nonexistent login → HTTP 200, silently absent from `requested_reviewers` → warning: "the login may not exist"
  - real user, not a collaborator → HTTP 422 with a message → warning includes that message verbatim
  
  All three are handled correctly, and `set -euo pipefail` does not abort the step on either failure case (the failing command sits inside `resp=$(...) || true`, so it isn't part of any `&&`/pipeline chain that `-e` would catch).
- Single file changed: `.github/workflows/claude-sentry-handler.yml`.

## Test plan
- [x] Validated YAML syntax locally: `python3 -c "import yaml; yaml.safe_load(open('.github/workflows/claude-sentry-handler.yml'))"`
- [x] Live-tested the REST reviewer-request call against this PR itself (#4313) for all three outcome shapes (real access, nonexistent login, real user without access), then removed the test review requests immediately after
- [x] Confirmed `gh pr edit --add-reviewer` (and `gh pr edit` in general) fails unconditionally in this repo due to an unrelated Projects (classic) deprecation, which is why the fix avoids it entirely
- [ ] Full end-to-end verification still requires a real `sentry-rca` workflow run that produces a `simple` verdict (per the issue's acceptance criteria), since org team-membership resolution under the Actions `GITHUB_TOKEN` can only be exercised from within Actions

### DIFF
--- a/.github/workflows/claude-sentry-handler.yml
+++ b/.github/workflows/claude-sentry-handler.yml
@@ -579,9 +579,18 @@ jobs:
           for r in "${candidate_reviewers[@]}"; do
             resp=$(gh api --method POST "repos/$REPO/pulls/$pr_number/requested_reviewers" \
               -f "reviewers[]=$r" 2>/dev/null) || true
-            if ! jq -e --arg r "$r" '(.requested_reviewers // []) | any(.login == $r)' <<<"$resp" >/dev/null 2>&1; then
-              echo "::warning::Reviewer '$r' no longer has access to $REPO — skipping."
+            if jq -e --arg r "$r" '(.requested_reviewers // []) | any(.login == $r)' <<<"$resp" >/dev/null 2>&1; then
+              continue
             fi
+            # A real access problem (e.g. not a collaborator) comes back as a JSON error
+            # with a `.message`; an unknown/misspelled login instead gets silently
+            # dropped from a 200 response with no error at all, so fall back to a
+            # explanatory reason in that case rather than leaving the warning bare.
+            reason=$(jq -r '.message // empty' <<<"$resp" 2>/dev/null)
+            if [ -z "$reason" ]; then
+              reason="GitHub accepted the request but did not add '$r' as a reviewer — the login may not exist."
+            fi
+            echo "::warning::Reviewer '$r' was not added to $REPO: $reason"
           done
 
       # Checkpoint 2: confirm the branch that ran actually produced its outcome on
--- a/.github/workflows/claude-sentry-handler.yml
+++ b/.github/workflows/claude-sentry-handler.yml
@@ -568,9 +568,18 @@ jobs:
           # via org team membership under GITHUB_TOKEN (it 404s even for users who do
           # have push access), so it can't be used as a pre-check. Request each reviewer
           # directly instead and let a genuinely stale login warn without blocking the rest.
+          #
+          # This calls the REST endpoint directly rather than `gh pr edit --add-reviewer`:
+          # that command's GraphQL query pulls the deprecated Projects (classic)
+          # `projectCards` field and fails outright on every invocation in this repo,
+          # regardless of the reviewer. The REST endpoint also returns 200 and silently
+          # drops an unknown login instead of erroring, so success is confirmed by
+          # checking the response body for the login rather than trusting the exit code.
           candidate_reviewers=(agalin920 geodem127 finnar-bin)
           for r in "${candidate_reviewers[@]}"; do
-            if ! gh pr edit "$pr_number" --repo "$REPO" --add-reviewer "$r" >/dev/null 2>&1; then
+            resp=$(gh api --method POST "repos/$REPO/pulls/$pr_number/requested_reviewers" \
+              -f "reviewers[]=$r" 2>/dev/null) || true
+            if ! jq -e --arg r "$r" '(.requested_reviewers // []) | any(.login == $r)' <<<"$resp" >/dev/null 2>&1; then
               echo "::warning::Reviewer '$r' no longer has access to $REPO — skipping."
             fi
           done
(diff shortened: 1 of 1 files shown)


finnar-bin:
Thanks for confirming — matches expectations, since this PR only touches `.github/workflows/claude-sentry-handler.yml` (the Sentry auto-fix PR's reviewer-request step) and has no `src/` or UI surface for the negative-QA browser pass to exercise.

For anyone following along: this change was also verified live against this PR itself while developing it — the reviewer-request REST call was exercised against three real response shapes (valid reviewer added, unknown login silently dropped by GitHub, and a real non-collaborator returning 422), with each test review request removed immediately after. Full end-to-end coverage of the actual reviewer-assignment behavior still needs a real `sentry-rca` run with a `simple` verdict, since it depends on org team-membership resolution under the Actions `GITHUB_TOKEN` that can only be exercised from within Actions.

geodem127 (dismissed):

agalin920 (dismissed):

agalin920 (approved):

PR #4342, opened 2026-09-19
Shell - Recover TinyMCE editor from upstream list-merge HierarchyRequestError

### LINKED ISSUE #4325: HierarchyRequestError: The operation would yield an incorrect node tree.
Sentry Issue: [MANAGER-UI-3E6](https://zestyio.sentry.io/issues/7716471443/?referrer=github_integration)

```
HierarchyRequestError: The operation would yield an incorrect node tree.
  at appendChild ([native code])
  at moveChildren (manager-ui/./node_modules/tinymce/plugins/lists/plugin.js:1675:1)
  at callback (manager-ui/./node_modules/tinymce/plugins/lists/plugin.js:1765:1)
  at ignore (manager-ui/./node_modules/tinymce/tinymce.js:18133:1)
  at ignore (manager-ui/./node_modules/tinymce/tinymce.js:18348:1)
...
(11 additional frame(s) were not displayed)
```

geodem127:
## Summary
Recovers TinyMCE fields from an upstream `HierarchyRequestError` thrown inside
TinyMCE's own `lists` plugin when a Safari user merges two adjacent lists with
Backspace, so the affected editor reverts to its last consistent content
instead of being left with a partially-merged, corrupted list.

## Root cause
Every frame in the Sentry stack trace is inside TinyMCE itself
(`tinymce.js`, `tinymce/plugins/lists/plugin.js`) — there is no first-party
bug. `backspaceDeleteFromListToListCaret` → `moveChildren` → `appendChild`
can throw a native `HierarchyRequestError` on a Safari-specific DOM-timing
race during list-to-list merge. No `tinymce` release from the current pin
(`^6.7.2`) through the latest 6.x/7.x/8.x changelog documents a fix for this
merge path, so a version bump isn't the fix here.

## Changes
- `src/shell/components/FieldTypeTinyMCE/index.tsx` — add a module-scoped
  singleton `window` error listener that detects this specific
  `HierarchyRequestError` and calls `undo()` on the focused editor only.
  This fixes the one concrete gap in a mitigation Sentry's own RCA had
  already prototyped and discarded: a listener registered per editor
  instance would incorrectly `undo()` every mounted editor on the page
  (e.g. repeater rows), not just the one that failed.

## Testing
- `npx tsc --noEmit` — clean.
- Code review (this plugin's automated pass) — confirmed singleton/API
  usage correctness, no destructive false-positive path.
- Manual/Playwright regression pass on the dev instance: seeded two
  adjacent lists in a TinyMCE field and Backspace-merged them — editor
  stayed alive and content was preserved, no new console errors. The
  upstream Safari-only throw itself isn't reliably reproducible outside
  Safari, so this validates the fix doesn't regress normal editor behavior.
- No Cypress spec added — there's no deterministic assertion to write
  against an intermittent third-party race condition.

## Related
Refs #4325

---


agalin920:
@geodem127 This is extremely hacky I don't want us to maintain this. TinyMCE should patch this and we should be using the patched version

PR #4343, opened 2026-09-20
Tooling - Stand up ESLint with a single design-system rule

### LINKED ISSUE #4288: Manager UI - Stand up ESLint with a single design-system rule
## Problem

**There is no ESLint in this repo** — no config, no dependency, 873 source files. So `docs/design-system.md` (#4263) is a convention with nothing behind it, and #4264's sweep of 84 hardcoded colour literals is a one-time cleanup that decays. #4264 says so itself: *"Expect to re-run the count in six months."*

## Division of labour — this is the whole scoping rule

- **Prettier owns formatting.** No formatting rules. `pretty-quick --staged` already runs on pre-commit and stays the only formatter.
- **TypeScript owns types.** No type-aware rules and no `@typescript-eslint` plugin. We need `@typescript-eslint/parser` only so ESLint can *read* TSX syntax — the parser, not the rule set.
- **ESLint owns project semantics** — the one thing neither of the above can express: *this value should come from the design system.*

Anything that isn't the third bullet is out of scope.

## Scope

**In:**
- `eslint` + `@typescript-eslint/parser` as devDependencies, one flat config file.
- **Exactly one rule**, at `error`: string literals matching `^#[0-9a-fA-F]{3,8}$` under `src/`.
- Allowlist the six paths `docs/design-system.md` §3 already exempts: `utility/brandColors.ts`, `schema/.../StarterBlocks/configs.ts`, `settings/.../Workflows/constants.tsx`, `.../VersionSelector/VersionItem.tsx`, `shell/index.js`, `reports/.../Metrics/index.js`.
- An `npm run lint` script.

**Out:** `eslint:recommended` and every plugin (React, hooks, a11y, import). Turning those on across 873 previously-unlinted files is a different ticket with a different risk profile, and bundling them is how this one fails to land — see #4262/#4279, closed for being larger than the problem.

**Out by default: the CI wiring.** `.github/workflows/` changes need explicit go-ahead — prepare the diff and hand it over. No workflow runs a lint or typecheck today, so there is no static-checks job to attach to.

## Why a rule beats the grep

Anchored on *string literals* rather than raw text, the pattern automatically excludes the two false-positive classes that cost real time on #4263 and #4264: issue references in comments (`#4256`) and ZUID URL fragments (`#a630bb24-0760-…`).

## The residue, measured

91 hex occurrences remain after #4264. **43 sit inside the allowlisted paths** and pass silently. **48 need an inline disable with a reason** — and that is the point: each is a value with no token, which §4 defines as a design escalation. Today they are invisible; a disable comment turns `grep` into the design-ask backlog.

## Acceptance criteria

- [ ] `npm run lint` exits 0 on a clean tree.
- [ ] The rule **fires** on a hex literal added to a non-exempt file — demonstrated, not asserted.
- [ ] The rule **does not fire** on the six allowlisted paths, on a comment containing `#4256`, or on the ZUID fragment in `schema/.../FieldList.tsx`.
- [ ] Every inline disable carries a one-line reason.
- [ ] No rule other than the hex rule is enabled, and no formatting or type-aware rule is present.
- [ ] CI/hook wiring prepared as a diff, not merged without go-ahead.

## Notes

- **Land after #4264**, or it is ~175 violations instead of 48.
- A second rule for raw `fontSize` px (#4287) is the obvious follow-on. Not here.
- Related: #4263, #4264, #4265, #4287.

geodem127:
### **CHANGE:**
- Adds ESLint to the repo (previously absent) with exactly one rule: hex-color string literals are an error anywhere under `src/`, except six files `docs/design-system.md` §3 already exempts.
- `npm run lint` now exists and exits 0 on a clean tree.

### **IMPLEMENTATION:**
- `eslint.config.js` (ESLint 9 flat config, CommonJS): `no-restricted-syntax` with selector `Literal[value=/^#[0-9a-fA-F]{3,8}$/]`, scoped to `src/**/*.{ts,tsx,js,jsx}` via `@typescript-eslint/parser` (JSX-enabled, used only to read TSX syntax — no `@typescript-eslint` plugin/rules).
- A second config block turns the rule off for the six allowlisted files.
- Every other hex literal under `src/` either got a one-line-reasoned `eslint-disable-next-line` citing a design-system.md §3 category, or was replaced with an existing/newly-minted `@zesty-io/material` theme token (5 new tokens added to `createTheme` in `src/shell/index.js` per §4).
- Removed 4 dead `eslint-disable-next-line react-hooks/exhaustive-deps` comments — `eslint-plugin-react-hooks` was never a dependency, and ESLint 9 errors on directive comments referencing an unknown rule ID.
- New devDependencies: `eslint@9.39.5`, `@typescript-eslint/parser@8.70.0`.

### **TESTING PLAN:**
- `npm run lint` exits 0 on the full tree; `npx tsc --noEmit` exits 0.
- Verified live: the rule fires on a hex literal added to a non-exempt file, and stays silent on the six allowlisted files, on a `#4256`-style comment, and on the ZUID URL fragment in `FieldList.tsx`.
- Playwright spot-check of every touched app surface (Media, Content, Schema, Reports) on the dev instance — no visual regressions.

### **NOTE — proposed CI wiring, not included in this PR:**
```yaml
name: Lint
on: [pull_request]
jobs:
  lint:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v4
      - uses: actions/setup-node@v4
        with:
          node-version: "22.19.0"
      - run: npm install --legacy-peer-deps
      - run: npm run lint
```

Closes #4288


src/shell/index.js
```diff
@@ -77,6 +77,16 @@ const themeOverrides = {
     },
     warning: {
       contrastText: "#fff",
+      // No ramp step matches this dark, accessible-on-tinted-background
+      // warning text color (see docs/design-system.md §4) — named here so
+      // it's greppable in one place; consumed via color: "warning.alertText"
+      // in src/apps/media/src/app/components/FileModal/RenameFileModal.tsx.
+      alertText: "#643a03",
```
agalin920: We definitely shouldnt be adding arbitrary things to the theme. Find what semantic value this should be and use that instead. Work with Bobby if needed

src/mui.d.ts
```diff
@@ -31,3 +31,20 @@ declare module "@mui/material/IconButton" {
...
 }
+
+declare module "@mui/material/styles" {
+  // One-off tokens minted per docs/design-system.md §4 for values with no
+  // existing ramp/semantic match, consumed outside sx (so a "path.string"
+  // can't resolve them) — see the themeOverrides.custom comment in
+  // src/shell/index.js for each entry's rationale.
+  interface Theme {
+    custom: {
+      navTreeDragOverBackground: string;
```
agalin920: remove once correct key is landed on

agalin920 (changes requested):
Make sure eslint errors fail on CI so a reviewer can catch it

geodem127:
### Addressed review feedback:
___
- **CI lint gate** (agalin920, blocking): added `.github/workflows/lint.yaml`, matching the YAML this PR's description had proposed but left uncommitted — `npm run lint` now runs and blocks on `pull_request`.
- **`warning.alertText` / `custom.navTreeDragOverBackground`** (agalin920, blocking, "we definitely shouldn't be adding arbitrary things to the theme"): reverted both back to disabled hex literals at their call sites (`RenameFileModal.tsx`, `RichTreeItem.tsx`) and removed the now-unused `custom` type augmentation in `mui.d.ts`. Neither value matches an existing ramp/semantic token (checked `grey`, `yellow`), so this needs a real design-system decision — **flagging for follow-up with Bobby/design**, not resolved by this push.
- **Bonus fix**: while reverting, found the `theme.palette.error.main` swap in `Field.tsx`/`SubField.tsx` (for the "field missing options" warning) silently changed the rendered color — `error.main` is `#F04438`, not the original `#e53c05` — same root cause as above (no token matches `#e53c05` either, and `docs/design-system.md` explicitly flags that value as a `legacyTheme` color that's "never the right answer"). Restored the original literal there too; also flagging for the same design follow-up.

`npm run lint` and `npx tsc --noEmit` are both clean on the branch.

### Still open, needs design input** (not resolved here — value pending):
___
- Text/icon color for the rename-file warning Alert (currently `#643a03` literal)
- Nav-tree drag-over highlight background (currently `#f6f6f7` literal)
- "Field missing options" warning heading color (currently `#e53c05` literal)

**For @bobakzahedi **: built an interactive comparison tool to make this easier to call — shows each color in its real context, live WCAG contrast against the actual background, nearby ramp-step candidates, and a spot to type in a custom hex to try live: [token-triage-pr-4343.html](https://github.com/user-attachments/files/32497206/token-triage-pr-4343.html)

src/shell/components/NavTree/components/RichTreeItem.tsx
```diff
@@ -17,6 +17,7 @@ export const RichTreeItem = memo(
   forwardRef((props: CustomTreeItem2Props, ref: React.Ref<HTMLLIElement>) => {
     const { onItemDrop, dragAndDrop, ...otherProps } = props;
     const { id, itemId, label, disabled, children } = otherProps;
+    const theme = useTheme();
```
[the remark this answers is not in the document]
agalin920: @geodem127 address

src/apps/content-editor/src/app/components/FieldTypeMedia.tsx
```diff
@@ -849,6 +849,7 @@ export const MediaItem = ({
             px={0.25}
             boxSizing="border-box"
             sx={{
+              // eslint-disable-next-line no-restricted-syntax -- third-party brand color for the Bynder integration badge (design-system.md §3)
               backgroundColor: "#0af",
```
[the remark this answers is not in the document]
agalin920: @geodem127 address

agalin920:
@geodem127 claude AC QA seems to still be failing

geodem127:
### Addressed remaining review feedback:
- **`RichTreeItem.tsx:20`** (unused `useTheme()` call, github-actions review + agalin920 "address"): removed the dead call and its now-unused import.
- **`FieldTypeMedia.tsx:853`** (Bynder brand color left inline instead of `brandColors.ts`, github-actions review + agalin920 "address"): moved `#0af` into `BRAND_COLORS.bynder` in `src/utility/brandColors.ts`, imported at the call site.
- **cv-verifier FAIL** ("claude AC QA seems to still be failing"): the PR description's "NOTE — proposed CI wiring, not included in this PR" was stale — `.github/workflows/lint.yaml` was already committed in the prior push. Updated the description to reflect that it's now an actual CI gate, and to reflect the reverted theme-token changes instead of the original "5 new tokens" claim.

`npm run lint` and `npx tsc --noEmit` are both clean.

The three "still open, needs design input" items from the prior push (rename-file alert color, nav-tree drag-over background, field-missing-options heading color) remain unresolved pending a design-system decision — unchanged by this push.

.github/workflows/lint.yaml
finnar-bin: I'd rather this be in included in the ci.yaml as this is a CI step. Also suggest making this a requirement for all_tests_passed to succeed.

finnar-bin (changes requested):

src/shell/components/NavTree/components/RichTreeItem.tsx
```diff
@@ -91,6 +91,7 @@ export const RichTreeItem = memo(
               onDragOver: (event: any) => {
                 if (dragAndDrop) {
                   event.preventDefault();
+                  // eslint-disable-next-line no-restricted-syntax -- pending design-system value (PR #4343 review); no ramp step matches this drag-over highlight, reverted from a minted theme.custom.navTreeDragOverBackground token
                   event.currentTarget.style.backgroundColor = "#f6f6f7";
```
finnar-bin: This should use the design system values.

finnar-bin (changes requested):
There seems to be a lot of usage of colors not using the design system values. These should be converted to what we already have on the design system as this is the point of setting up eslint. Probably be worth checking in with Bobby in order to get these converted to our design system color scheme.

src/shell/components/NavTree/components/RichTreeItem.tsx
geodem127: Agreed, it should. No existing ramp step matches `#f6f6f7`, and minting a one-off theme token was rejected earlier in this review, so for now it's a disabled literal with a "pending design-system value" note. I'll update it to the design-system color as soon as @bobakzahedi provides it. The comparison tool is linked in [this comment](https://github.com/zesty-io/manager-ui/pull/4343#issuecomment-5765047060).

geodem127:
@finnar-bin re: [your review](https://github.com/zesty-io/manager-ui/pull/4343#pullrequestreview-5299500829): thanks. Three values here need a design-system color: the rename-file alert text and icon (`#643a03`), the nav-tree drag-over background (`#f6f6f7`), and the field-missing-options heading (`#e53c05`). None of them matches an existing token, so I'll update them once @bobakzahedi provides the colors. The `LinkedSchema.tsx` syntax-highlight colors (`#95c65c`, `#91ace8`) are going to him too. I've converted `TopUsers.js`'s `#fff` to `theme.palette.common.white` (f06f34f27). The remaining disabled literals are stored color-field values and comparisons (design-system.md §3), where a token would break data.

Issue #3537, opened 2025-05-22
Workflows - External publish actions being disrupted by allow publish workflow status


finnar-bin:
- [ ] If a content item is included when bulk publishing from the multipage content item table, the error message is confusing and does not immediately tell the user that there's an issue related to a missing allow publish workflow status.
  - To be resolved by https://github.com/zesty-io/instances-api/issues/674

- [ ] Publishing related relational fields (one-to-one & one-to-many) on a content item
  - FE ticket: https://github.com/zesty-io/manager-ui/issues/3539
  - BE ticket: https://github.com/zesty-io/instances-api/issues/676
 
- [ ] Instance module installer
- [ ] Freestyle blocks publish

agalin920:
@geodem127 whats the status of this ticket?

geodem127:
> [@geodem127](https://github.com/geodem127) whats the status of this ticket?

@agalin920 
Status on the checklist:

- **Bulk publish confusing error (item 1)** — PR #4344 (open) fixes this. The backend fix (instances-api#674, closed via instances-api#678) only improved the error message for *single*-item publish — the batch endpoint (`items/publishings/batch`) still fails the whole batch atomically if one item is blocked, with no per-item detail. #4344 pre-checks each item's workflow status client-side, publishes the valid subset, and names the excluded item(s) in a toast.
- **Related one-to-one/one-to-many publish (item 2)** — already resolved. Split into FE #3539 (closed, merged via #3844) and BE instances-api#676 (closed via the same instances-api#678). #3844 shows an error toast when a related item's publish fails; #4344 additionally makes this proactive — blocked related items show disabled with a reason before the user even clicks Publish, instead of attempting and failing.
- **Instance module installer** — not started. No related code found in manager-ui; unclear if this even lives in this repo. Needs its own investigation before it can be scoped.
- **Freestyle blocks publish** — not started, and can't be fixed from manager-ui. The Freestyle editor is a separate iframed app (its own repo); `FreestyleWrapper.tsx` here only mounts the iframe and doesn't own any publish call.

Once #4344 merges I'd suggest splitting this issue — closing it for items 1–2 and filing separate tracked tickets for the module installer and Freestyle gaps, since neither has an owner in manager-ui as-is.

PR #4344, opened 2026-09-20
[Content] - Block publish for items missing allow-publish status

geodem127:
### **SCOPE:**
Addresses checklist items 1 & 2 of #3537 only (the other two — instance module installer, Freestyle blocks publish — are untouched by this PR and unowned in this repo; see issue comment for detail). Not a full close of #3537.

Item 1's backend ticket (instances-api#674) and item 2's backend ticket (instances-api#676) are both closed via the same backend PR (instances-api#678), which only improved the error message for **single**-item publish. The batch endpoint (`items/publishings/batch`) used by bulk publish from the multipage list was not touched by that backend fix — it still fails the entire batch atomically with no per-item detail if any one item is blocked. This PR's client-side pre-check is what actually fixes that for item 1.

Item 2 (related-item publish) already got a reactive fix in #3844 (toast on failure, using the backend's clearer per-item message). This PR upgrades that to a proactive check — blocked related items show disabled with a reason before Publish is clicked, instead of being attempted and failing.

### **RCA:**
- Bulk publish from the multipage content list called `createItemsPublishing` for every selected item with no workflow-status check — a batch containing one item whose status doesn't allow publishing failed the *entire* batch with a generic, unhelpful error, giving no indication which item or why.
- Publishing a content item together with its one-to-one/one-to-many related items co-published every "unpublished related item" found, with no check of each related item's own workflow status — a blocked related item was either silently attempted and failed, or offered as checkable when it shouldn't have been.

### **FIX:**
- `src/utility/workflowStatus.ts` (new) — extracted the existing single-item allow-publish label-intersection check into a reusable, framework-agnostic helper (preserves "no allow-publish labels configured → publish always allowed" default).
- `src/shell/hooks/useCheckPublishAllowed.ts` (new) — RTK-Query-aware hook that batch-checks a list of items' workflow status against the instance's allow-publish labels.
- `src/apps/content-editor/src/app/views/ItemList/UpdateListActions.tsx` — pre-checks selected items before `createItemsPublishing`; publishes only the allowed subset and shows a named message for any excluded items.
- `src/apps/content-editor/src/app/views/ItemEdit/components/ItemEditHeader/ItemEditHeaderActions.tsx` and `UnpublishedRelatedItem.tsx` — related items are checked against their own workflow status; a blocked one renders disabled with an inline reason and is excluded from co-publish.
- New i18n keys `content.itemListCannotPublishStatus` (full CLDR plural set per locale) and `content.itemEditRelatedItemCannotPublishStatus` across all 6 locales (English placeholder text in the 5 non-English locales per this repo's convention — translation is a separate manual/QA step).
- Extended `cypress/e2e/content/workflows.spec.js` with a bulk-publish/blocked-item scenario.

### **TESTING PLAN:**
- Test data: two workflow-status labels (one `allowPublish: true`, one `false`), two content items in a multipage model, one of each label; a second parent/child model pair with a one-to-one field, the child given a newer draft version carrying the blocking label.
- Navigation: `/content/<modelZUID>` → select both items → Publish → Confirm; and `/content/<modelZUID>/<itemZUID>` → Publish → confirm dialog's related-items checklist.
- Expected/verified: blocked item excluded from the batch with a named toast, allowed item's publish completes, dialog closes; blocked related item shown disabled with a reason and excluded from the parent's co-publish. Verified via 3 rounds of Playwright UI verification (the first two rounds caught and fixed a data-loading race and an RTK Query subscribe/unsubscribe bug) plus direct API checks of publishing records.

### **RECORDING:**
https://github.com/user-attachments/assets/ce62f440-2929-405b-b3bb-01fb232f8320

Refs #3537 (partial — see Scope)


geodem127:
Addressed review feedback:

- **Modal stuck open on blocked publish** (negative-qa): `handlePublish`'s `else` branch (main item itself blocked) now calls `setIsConfirmPublishModalOpen(false)`, matching the allowed-publish path.
- **Effect re-fires on unrelated content updates** (code review): the related-items workflow-status check effect now keys off a stable ZUID+version string instead of the `unpublishedRelatedItems` array reference, so it no longer re-runs on every unrelated content fetch elsewhere in the app.
- **Unbounded parallel requests for large bulk-publish batches** (code review): `useCheckPublishAllowed` now chunks `getItemWorkflowStatus` requests into batches of 15 instead of firing them all in one `Promise.all`.

Not changed: the 10 untranslated-placeholder findings from the localization reviewer. Per this repo's documented i18n convention, new keys intentionally ship with English placeholders in the 5 non-English locales, with translation as a separate manual/QA step (also called out explicitly in this PR's description) — so these aren't defects.

`npx tsc --noEmit` passes. No unit test suite in this repo; existing Cypress coverage in `cypress/e2e/content/workflows.spec.js` already exercises this flow, and these are behavior-preserving fixes so no spec changes were needed.

agalin920:
@geodem127 

Item 1's backend ticket (instances-api#674) and item 2's backend ticket (instances-api#676) are both closed via the same backend PR (instances-api#678), which only improved the error message for single-item publish. The batch endpoint (items/publishings/batch) used by bulk publish from the multipage list was not touched by that backend fix — it still fails the entire batch atomically with no per-item detail if any one item is blocked. This PR's client-side pre-check is what actually fixes that for item 1.

Its not clear what this PR is/ How is the client side getting around this issue?  Is this really a bug?

geodem127:
@agalin920

The batch publish endpoint (`items/publishings/batch`, used when you select multiple items in the content list and click Publish) has always been all-or-nothing: if *any* one item in the batch has a workflow status that doesn't allow publishing, the whole request fails, and the error returned doesn't say which item or why.

Backend PR instances-api#678 fixed the *error message* for that scenario, but only on the **single-item** publish endpoint. It didn't touch the batch endpoint — so a bulk publish still fails entirely with no per-item detail if one item is blocked, exactly as before.

This PR doesn't change server behavior. It adds a client-side pre-check: before calling the batch endpoint, `useCheckPublishAllowed` looks up each selected item's workflow status and its labels, filters out any that aren't allowed to publish, and only sends the allowed subset to the batch endpoint. The excluded items are named in a toast so the user knows what got skipped and why.

I re-verified this live just now rather than relying on the original testing notes: seeded a model with one allow-publish item and one blocked item on the dev instance, selected both from the list, and confirmed via network inspection that the actual `items/publishings/batch` request body contained only the allowed item's ZUID — the blocked one never went out — while the toast named it and explained why.

So the gap is real, and this fix addresses it for the bulk case specifically. Given that, I'd like guidance: is a client-side pre-check the right place for this, or should this be routed to the backend team to fix the batch endpoint itself (e.g. having it return per-item results/errors like a partial-success response, instead of failing atomically)? A backend fix would be the more durable answer, but I don't know if that's already planned/scoped anywhere.

finnar-bin (changes requested):
@geodem127 I'm not sure what this is trying to address. The listed linked tasks on https://github.com/zesty-io/manager-ui/issues/3537 have already been resolved

